PR #1, opened 2020-01-13
Notary v2 initial scenarios

SteveLasker:


scenarios.md
```diff
@@ -0,0 +1,190 @@
...
+**Implications of this requirement:**
+
+- The aspect of being a signed artifact is not germane to the CD solution. The CD solution simply requests a deployment. The enforcement of signature validation is a policy of the execution environment.
+
+### Scenario #8 Automated Deploy - Failure
+
+Same as #7, with the exception that the signature is deemed invalid.
+1. A CD solution is triggered to initiate a deployment.
+2. The CD solution uses a deployment template to request an orchestrator to run a specific image. In this case, the CD solution mistakenly attempted to deploy a dev signed artifact to a production environment.
+3. The production orchestrator receives the request, validates the artifact is signed and recognizes the signature is **NOT** one of the production accepted signatures. As the signature is deemed invalid, the deployment fails.
```
samuelkarp: I'd replace "one of the production accepted signatures" with "produced by a trusted key".

scenarios.md
```diff
@@ -0,0 +1,190 @@
...
+- Signatures represent validations through different dev, staging, production environments.
+
+1. A CI/CD solution builds, signs, pushes and deploys a collection of artifacts to a staging environment.
+1. Once integrations tests are completed, the artifacts are signed with a production signature, copying them to a production registry or production set of repositories.
+1. The Staging and Production container hosts validate the artifacts contain the signatures required for their environment.
+
+**Implications of this requirement:**
+
+- Multiple signatures, including signatures from multiple sources can be associated with a specific artifact.
+
```
samuelkarp: One of the scenarios that we talked about in the face-to-face meeting but isn't covered here is how signatures are expected to work with base images/layer inheritance.  For example, consider the following:

An image A consists of layers 1, 2, and 3 and is signed by vendor FOO.  This image is then used as the basis of image B which adds layers 4, 5, and 6 and is signed by another party BAR.

Can we validate that layers 1, 2, and 3 (the base image) still have a valid signature by FOO?  Or are we only able to validate the signature of all the layers generated by BAR?

scenarios.md
```diff
@@ -0,0 +1,190 @@
+# Notary Signing - Scenarios
+
+As containers and cloud native artifacts become the common unit of deployment, users want to know the artifacts in their private registries and the artifacts deployed are the same artifacts that were initially published.
+
+The [OCI TOB][oci-tob] has adopted [OCI Artifacts][artifacts-repo], generalizing container images as one of many types of artifacts that may be stored in a registry. Other artifact types currently include:
```
lumjjb: There is some work being done with signing of yaml files or other type of policy based files, which are not necessarily OCI artifacts. Would something like that be in scope? 

i.e. if i want to verify the signature of an object, will the object need to be resident in an OCI registry and/or need to be an OCI object?

samuelkarp: > if i want to verify the signature of an object, will the object need to be resident in an OCI registry and/or need to be an OCI object?

This is proposed in Scenario 1:

> - The signing and validation of artifacts does not require a registry.
> - The signature used for validation may be hosted in a registry, or other accessible location.

scenarios.md
```diff
@@ -0,0 +1,190 @@
...
+1. The Staging and Production container hosts validate the artifacts contain the signatures required for their environment.
+
+**Implications of this requirement:**
+
+- Multiple signatures, including signatures from multiple sources can be associated with a specific artifact.
+
+## Open Discussions
+
+- What is the relationship between a signature, an artifact and a registry?
+- Can signature validation be dependent on an external entity?
```
lumjjb: Orthogonally, can signing be done by an external entity, what formats would the signatures be?

Motivation for this is that if there is an existing generic signing service (with HSM integrations, etc.) that has been used to sign packages, can they be used to generate signatures as well?

scenarios.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Scenarios
+
+### Scenario #1: Local Build, Sign, Validate
+
+Prior to doing any deployment, a developer can test the: build, sign, validate scenario.
+
+1. Locally build a container image using a non-registry specific `name:tag`, such as:  
+  `$docker build hello-world:1`
+1. Locally sign `hello-world:1`
```
lumjjb: On the similar vein as another comment below on external validation. Is it possible to sign something with an external service which adheres to a particular signature format? Is the interface to signing with a cryptographic hardware module be a direct interface from notary library via PKCS11, or maybe a gRPC API to have the specific cryptographic signing of a hash be done on another server? 

Or would the signing format be exposed, and signing services can adhere to the format to produce its own signatures with the help of notary libraries?

scenarios.md
lumjjb: Gotcha! Thanks I missed that.

Is there a restriction of signing requiring an OCI object or are generic blobs fine? Or do you see the signing of a blob or yaml just associated with creating an "OCI Blob" type and encompassing everything else under that?

scenarios.md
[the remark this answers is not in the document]
SteveLasker: I believe we ended with:
We'll focus on the scenarios here, and work with each specific implementation to support the scenario. For eg: Justin mentioned Derek has PRs to merge efforts into the ContainerD work, where this would be enabled.

scenarios.md
```diff
@@ -0,0 +1,190 @@
...
+
+1. Locally build a container image using a non-registry specific `name:tag`, such as:  
+  `$docker build hello-world:1`
+1. Locally sign `hello-world:1` 
+2. Run the image in a local environment which was configured to only support signed artifacts. *(**note** - details on the usability is deferred from this requirement. This could be a local docker host, a local kind or minikube instance configured to only run signed images)*  
+  `$ docker run hello-world:1`
+
+**Implications of this requirement:**
+
+- The local developer has access to the signing keys
```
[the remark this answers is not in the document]
SteveLasker: Good catch. Please re-activate if I didn't clarify correctly.

scenarios.md
```diff
@@ -0,0 +1,190 @@
...
+  * Singularity
+  * Operator Bundles
+
+## Scenarios
+
+### Scenario #1: Local Build, Sign, Validate
+
+Prior to doing any deployment, a developer can test the: build, sign, validate scenario.
+
+1. Locally build a container image using a non-registry specific `name:tag`, such as:
```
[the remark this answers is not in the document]
SteveLasker: ugh, the dreaded default registry. Ok. I've added the following:
- The lack of a registry name does not infer docker.io as the default registry. This does imply changes to either the docker client, the notary client or both.

@dmcgowan do you have any work in flight with containerd that addresses the default registry issues?

scenarios.md
```diff
@@ -0,0 +1,190 @@
...
+### Scenario #2: Sign, Rename, Push, Validate
+
+Once the developer has locally validated the build, sign, validate scenario, they will likely want to push the artifact to a registry used for deployment.
+
+1. Locally build and sign an artifact, such as a container image
+1. Rename the artifact to reflect the registry it will be pushed to:  
+  `$ docker tag hello-world:1 myregistry.domain.com/marketing/web:1`
+  `$docker push myregistry.domain.com/marketing/web:1`
+1. Deploy the artifact to a cluster that requires signatures:  
+  `$kubectl apply -f kubedeploy.yaml`
```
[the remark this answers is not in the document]
SteveLasker: I agree with the generalized statement here. What I'm trying to suggest is we do want/need integration from container, or any artifact host that retrieves artifacts and supports signing verification.
If we generalize this to pull, are we overly focusing on the docker client, and missing the scenarios where artifact hosts interact with registries directly? Or, using CRI. 
Suggestions?

scenarios.md
```diff
@@ -0,0 +1,190 @@
...
+  > **Note:** the above example presumes `kubedeploy.yaml` references the myregistry.domain.com/marketing/web:1 image
+
+**Implications of this requirement:**
+
+- Signatures can be validated based on the tag referenced, however the signature is not tied to a specific name. The tag is a pointer to the signed content.
+- This does not preclude tag locking scenarios. If a tag is updated with another signed artifact, the tag is considered signed and follows the signature validation rules of the new signature. For example, if a base image `fx:1.0` is signed, and rebuilt with a patched version, the updated `fx:1.0` is new content, and a valid scenario.
```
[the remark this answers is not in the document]
SteveLasker: There are a bunch of scenarios where customers avoid using tags, as human readable references because tags can be changed. I believe other registries have done this as well, but in ACR we support locking a tag to a digest. You can't assign a tag to another manifest, nor delete it. http://aka.ms/acr/tag-locking
I've added some more verbiage to hopefully clarify. Please reactivate if it needs more.

scenarios.md
```diff
@@ -0,0 +1,190 @@
...
+- If the tag for `fx:1.0` is redirected to another unsigned manifest, the signature validation of `fx:1.0` will fail as it's no longer signed. This is no different than signed binaries being updated or replaced on a users local computer.
```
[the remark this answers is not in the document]
SteveLasker: Hmm. I'm trying to understand how you're reading this. I wrote it with the assumption that a customer requires artifacts to be signed. In this case, one of two things could happen.
1. the new version of `fx:1.0` isn't signed, so it fails validation is all artifacts must be signed
1. the new version is signed with a different key, which isn't really covered here. 

I'll leave this open for now, and perhaps we can discuss more on the next call.

scenarios.md
```diff
@@ -0,0 +1,190 @@
...
+
+A CI/CD system promotes validated artifacts from a dev repository to production repositories.
+
+1. A CI/CD solution responds to a git commit notification, cloning, building, signing and pushing the artifacts to a development repo within their registry.
+1. As the CI/CD solution runs functional, determining the artifacts are ready for production, the artifacts are moved from one repo to another.  
+  `$ docker tag myregistry.domain.com/dev/alpha-team/web:1abc myregistry.domain.com/prod/web:1abc`
+
+**Implications of this requirement:**
+
+- Artifact movement or copy means the original signature is still available. In this scenario, the signature is saved to the same registry.
```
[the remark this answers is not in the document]
SteveLasker: Updated with your more concise text. For the implementation for how manifests, and repositories store signatures, I've left that to the spec, to specify the how.

scenarios.md
```diff
@@ -0,0 +1,190 @@
...
+1. ACME Rockets acquires networking monitoring software they run in their environment.
+1. ACME Rockets copies the artifacts, renaming them to reflect their private registry, and pushes them to their registry.  
+  `$ docker pull registry.wabbit-networks.com/netmonitor:1.0`
+  `$ docker tag registry.wabbit-networks.com/netmonitor:1.0 registry.acme-rockets.com/infra/networking/wabbit-net-monitor:1.0`
+  `$ docker push registry.acme-rockets.com/infra/networking/wabbit-net-monitor:1.0`
+1. As ACME Rockets deploys the networking monitoring software, the signatures are validated to be signed by ACME Rockets.
+
+**Implications of this requirement:**
+
+- In addition to the above scenario where an artifact is promoted within the same registry, this scenario requires the signature to be avaialble in the new registry. Whether the signature is moved, or referenced from it's original location is a subject for design discussion. In this scenario, ACME Rockets could validate the signature against a wabbit-network signature server.
```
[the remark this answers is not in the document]
SteveLasker: I was a bit confused if you're commenting on scenario 5, where ACME Rockets can reach registry.wabbit-networks.com to validate the signature, or scenario 6, where ACME Rockets must copy the signature to a trusted source. Is this a question for the how? I've added a comment that defers the how to the spec and UX designs. 
Please re-enable if I didn't address.

scenarios.md
SteveLasker: I was hoping to account for this. In the Hackmd.io history of notes you can find the following:
> ### Inherited Layer Signatures
> Should signature verification support artifacts that source from another? The most obvious are docker images, where one image is `FROM`-d another. Should a user be able to see the inheritance of an artifact, and know the signature of the base image, the layered runtime are represented in the artifact in question? 

There was a comment that this wasn't feasible. I don't know whether we should pull it in as a scenario, discuss and account for, defer, or promote this to be an SBOM scenario where we can see the history of an image. 

I do believe there's a common request for understanding what base layers are used. Suggesting we defer and add later. 
Feedback?

scenarios.md
SteveLasker: Thanks @lumjjb, this is a good example for where we need to be crisper on artifacts, registries and signing. These are each separate things that we want to enable uniting, but not coupling. 

With Notary v2, you should be able to:
- Sign a thing, (yaml, singularity image, helm chart, YetUnknownThing) 
  - Notary v2 will provide reference examples for how you sign things, for which the tooling around the thing can implement
- [ORAS](http://github.com/deislabs/oras) will most likely adopt the generic Notary v2 signing of things. So, you could put your `.yaml` file into ORAS, sign it, and optionally push it to a registry. 
- Push it to an OCI Compliant registry, which would know how to store Notary v2 signatures. 

The OCI Blob is simply a means to storing "things". What those things are may be determined with the [OCI Artifacts](https://github.com/opencontainers/artifacts) use of `manifest.config.mediaType`

As we get further into the Notary v2 specs, and we update OCI Artifacts, we'll get more clarity for how these will work together.

scenarios.md
SteveLasker: I believe our goal for Notary v2 is to support signing of things. This does imply a specific pattern. However, this seems like more of a spec question. If this could be re-phrased as a scenario, I can pull it back in.

scenarios.md
SteveLasker: Also deferring to a spec question, unless this can be rephrased as a scenario.

scenarios.md
```diff
@@ -0,0 +1,190 @@
+# Notary Signing - Scenarios
+
+As containers and cloud native artifacts become the common unit of deployment, users want to know the artifacts in their private registries and the artifacts deployed are the same artifacts that were initially published.
```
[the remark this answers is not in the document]
SteveLasker: I see the clarity and distinction you're asking. What I'm trying to articulate is the consumer of the artifact, in their current possession, can be validated as authentic. They may not know the original registry source to do a digest comparison. But, that's not the way the intro was written. 

I was thinking of adding this to the initial paragraph, but felt it went too deep, too quick:
> Customers can then further decide the set of entities they trust. 

Let me know what you think about the latest edits. This is hardest, but most important part to setup the following content.

scenarios.md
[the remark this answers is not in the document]
dmcgowan: Not sure what you mean by "the default registry issue". If you try to push an image called `hello-world:1` with the containerd client (or `ctr`) it will spit it back out at you for not defining a registry. If you put `hello-world:1` into Docker or Kubernetes (CRI), it will interpret it as `docker.io/library/hello-world:1`. This won't ever change in Docker and likely never will in Kubernetes (CRI) either since it would just break everyone for no real usability gain. From a notary perspective, it should only ever see `docker.io/library/hello-world:1` and leave it to the clients to take care of interpreting short names according to their own UX experience. I wouldn't really categorize it as the dreaded default registry though, each client is free to make their own opinionated UX. Notary should have no opinions at its core, in its tool, or in its documentation. I would suggest just using examples which include a full registry name except where you are explicitly giving an example of an official image in the Docker UX.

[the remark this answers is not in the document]
SteveLasker: > Default registry issue:

The docker client assumption that all un-scoped images belong to docker hub (docker.io)

> If you try to push an image called hello-world:1 with the containerd client (or ctr) it will spit it back out at you for not defining a registry. 

This is the desired behavior as it's presumptive for an open source project to default to a specific vendor. This has caused lots of issues and frustrations. 

> From a notary perspective, it should only ever see docker.io/library/hello-world:1 and leave it to the clients to take care of interpreting short names 

Agreed. As we venture into Notary v2, and new tooling, I'm asking we don't assume a default registry, nor assume one must be assigned. If a user names an entity `foo:1`, it's really a loose artifact, and has no affinity to a registry until given a domain. Which could be `docker.io/someone/foo:1` or  `mydomain.com/foo:1`
The disconnect is Artifacts shouldn't be tied to a registry, rather they happen to be stored in them at various states. But, there's no home, or even assumption it must be assigned a registry. 
So, we just need to agree that an Artifact that doesn't have a domain, isn't associated with any registry. Which isn't an error, it's just a state.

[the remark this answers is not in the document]
dmcgowan: > So, we just need to agree that an Artifact that doesn't have a domain, isn't associated with any registry. Which isn't an error, it's just a state.

I think we should be careful here though. I have come to the conclusion that partial names are actually worse than no name at all. For example, in the OCI index specification it tried to define the reference name as only a tag component of a name (such as say "v1.0.0"), however, this made these images unable to be imported by clients without specifying the rest of the name and making a decision (which allows for opinionated interpretation) of how the full name would get reconstructed. Without a name, it is clear that a name should be provided and it can be anything. With a full name, also clear with assigning a new name a common and understand mechanism. 

I would suggest either recommending full names which include the registry or using a label which represents the "partial name". Just avoid having an ambiguous field that requires the client to figure out how to interpret the string, even just identifying domain/no domain isn't always straight forward to implement.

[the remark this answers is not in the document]
samuelkarp: > So, we just need to agree that an Artifact that doesn't have a domain, isn't associated with any registry. Which isn't an error, it's just a state.

I think a better interpretation here is really just that the signature is independent of the image name.  We already have that implication in scenario 4.

For this specific scenario, I'd suggest changing the wording to 

> Locally build a container image with a name and tag

and avoiding mention of the registry.  As long as the example doesn't include pushing the image, it's still clear that the image must be able to be signed and validated independently of its presence in a registry.

scenarios.md
```diff
@@ -0,0 +1,199 @@
...
+2. Run the image in a local environment which was configured to only support signed artifacts. *(**note** - details on the usability is deferred from this requirement. This could be a local docker host, a local kind or minikube instance configured to only run signed images)*  
+  `$ docker run hello-world:1`
+
+**Implications of this requirement:**
+
+- The local developer has access to signing keys
+- The local environment has a policy by which it states the set of keys it accepts.
+- The signing and validation of artifacts does not require a registry.
+- The signature used for validation may be hosted in a registry, or other accessible location.
+- The lack of a registry name does not infer docker.io as the default registry. This does imply changes to either the docker client, the notary client or both.
```
samuelkarp: > The lack of a registry name does not infer docker.io as the default registry. This does imply changes to either the docker client, the notary client or both.

I don't think this is the right takeaway.  We don't expect to change Docker's handling of bare names here, but rather that signing and verification can both happen locally without a registry.

scenarios.md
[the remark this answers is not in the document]
samuelkarp: > are we overly focusing on the docker client, and missing the scenarios where artifact hosts interact with registries directly?

I think we want to design the spec in such a way that any interested client is able to integrate, including container runtimes that pull images (like Docker), orchestrators attempting to apply something like an admission or trust policy, scanning systems, etc.

The examples right now are pretty focused on Docker.  That's somewhat reasonable, as we expect most of the interested parties to be at least familiar with how to use Docker today.  But I don't think adding Kubernetes or CRI here makes this particular scenario better (and as someone who works on a non-Kubernetes orchestrator that doesn't use CRI), I'd prefer to not see those enshrined as the default way to do something.

It might be worth adding a separate scenario covering things other than container runtimes, and using a scanning service or an orchestrator as an example.

scenarios.md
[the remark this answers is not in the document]
samuelkarp: My confusion might stem from "validation" here: which party is doing that validation?

My initial read was that a client is performing validation on the signature.  For the client to know that the new `fx:1.0` is invalid because it's unsigned, the client must have some way to know that a signature is expected.  That could be one of several situations:

* the client has been configured to expect all images to have signatures
* the client has been configured to expect `fx:1.0` specifically to have a signature
* the client retrieved a previous manifest named `fx:1.0` that was signed, and now expects future manifests by that name to have a signature

> the new version is signed with a different key, which isn't really covered here.

This might be worth covering in another scenario, though I would expect the answer to be client-dependent and related to a particular trust policy.

scenarios.md
```diff
@@ -0,0 +1,199 @@
...
+1. The CI system signs the artifact with locally available keys *
+  (**Note:** key management deferred from this requirement)*
+1. The artifact is pushed to a registry:  
+  `$ docker push myregistry.domain.com/sample/hello-world:aa31`
+  `$ helm chart push myregistry.domain.com/sample/charts/hello-world:aa31`
+1. The artifacts are deployed, referencing the registry they were pushed to.
+
+**Implications of this requirement:**
+
+- nothing additional from previous scenarios.
```
samuelkarp: A new implication here: "Signing must work for other OCI artifacts like Helm charts"

scenarios.md
samuelkarp: > There was a comment that this wasn't feasible.

I think I missed that part of the discussion.  If I read into it a bit, someone might have thought this to be infeasible as the manifest is the actual part that is signed and a derived image will have a different manifest (and typically a different config).  But we might be able to save the original manifest (and config) as well and use that just for signature validation if we really want to cover this.

I think it'd be worth including this as a scenario and see if we can come up with an approach for it that doesn't have too many downsides.  Anyone else have an opinion here?

trishankatdatadog:
Thanks for starting this, @SteveLasker, I think it's a great start!

I wonder whether we could also add the full power of delegations from the [TUF 1.0.0 specification](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md)? The reason why I ask is because I think we could radically simplify the signing model for most developers this way. Using the full power of [delegations](https://www.usenix.org/node/194973), we can require developers to manage one signing key instead of four (`root`, `timestamp`, `snapshot` _and_ `targets`). This would vastly increase usability for most developers. What do you think?

scenarios.md
```diff
@@ -0,0 +1,199 @@
...
+A CI system is triggered by a git commit. The system can build the artifacts, sign them, and push to a registry. The production system can pull the artifacts, verify the signatures and run them.
+
+1. A CI solution responds to a git commit notification
+1. The CI system clones the git repo and builds the artifacts, with fully qualified names: **image**: `myregistry.domain.com/sample/hello-world:aa31` and a **helm chart**: `myregistry.domain.com/sample/charts/hello-world:aa31`
+1. The CI system signs the artifact with locally available keys *
+  (**Note:** key management deferred from this requirement)*
+1. The artifact is pushed to a registry:  
+  `$ docker push myregistry.domain.com/sample/hello-world:aa31`
+  `$ helm chart push myregistry.domain.com/sample/charts/hello-world:aa31`
+1. The artifacts are deployed, referencing the registry they were pushed to.
```
jonjohnsonjr: I'm not sure what this means -- ideally the artifacts (at least the images) wouldn't reference any registry, since that would break content-addressability. Are you saying the helm chart references the image? Or the deployment system references the registry?

scenarios.md
```diff
@@ -0,0 +1,199 @@
...
+**Implications of this requirement:**
+
+- nothing additional from previous scenarios.
+
+### Scenario #4: Promote Artifacts Within a Registry, Using a Different repo
+
+A CI/CD system promotes validated artifacts from a dev repository to production repositories.
+
+1. A CI/CD solution responds to a git commit notification, cloning, building, signing and pushing the artifacts to a development repo within their registry.
+1. As the CI/CD solution runs functional, determining the artifacts are ready for production, the artifacts are moved from one repo to another.
```
jonjohnsonjr: > As the CI/CD solution runs functional,

missing "tests"?

scenarios.md
```diff
@@ -0,0 +1,199 @@
...
+1. ACME Rockets acquires networking monitoring software they run in their environment.
+1. ACME Rockets copies the artifacts, renaming them to reflect their private registry, and pushes them to their registry.  
+  `$ docker pull registry.wabbit-networks.com/netmonitor:1.0`
+  `$ docker tag registry.wabbit-networks.com/netmonitor:1.0 registry.acme-rockets.com/infra/networking/wabbit-net-monitor:1.0`
+  `$ docker push registry.acme-rockets.com/infra/networking/wabbit-net-monitor:1.0`
+1. As ACME Rockets deploys the networking monitoring software, the signatures are validated to be signed by ACME Rockets.
+
+**Implications of this requirement:**
+
+- In addition to the above scenario where an artifact is promoted within the same registry, this scenario requires the signature to be available in the new registry. Whether the signature is moved, or referenced from it's original location is a subject for design discussion. In this scenario, ACME Rockets could validate the signature against a wabbit-network signature server.
```
jonjohnsonjr: > from it's original

from _its_ original

scenarios.md
```diff
@@ -0,0 +1,199 @@
...
+### Scenario #6: Validate Artifact Signatures Within Restricted Networks
+
+ACME Rockets runs secure production environments, limiting all external network traffic. To assure the wabbit-networks network monitor software has valid signatures, they will need to trust a resource within their network to proxy the signature validations.
+
+1. ACME Rockets acquires network monitoring software, copying it to their firewall protected production environment.
+1. As part of the artifact copy, they will copy/proxy the signature validation to trusted resources within their network protected environment.
+
+**Implications of this requirement:**
+
+- In this scenario, the wabbit-netowrks signature must be validated within the ACME Rockets network. How this is done is open for design. However, the requirement states the signature must be validated without external access. When the artifact is copied to the private/network restricted registry, the signature may need to be copied, and is assumed to be trusted if available in the trusted server within the private network. How ACME Rockets would copy/proxy the signatures is part of the design and UX for a secure, but usable pattern.
```
jonjohnsonjr: > wabbit-netowrks

wabbit-_networks_

scenarios.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Scenario #7 Automated Deploy - Success
+
+A CD solution initiates a deployment to a development environment. The development environment has been previously configured to only accept signed artifacts.
+
+1. A CD solution is triggered to initiate a deployment
```
jonjohnsonjr: Missing a period at the end of this line.

scenarios.md
```diff
@@ -0,0 +1,199 @@
...
+**Implications of this requirement:**
+- The coordination between a CD and operational environment for how a request and response is managed is not in-scope for artifact signing.
+- An operational environment enforces a policy for which signatures it accepts. Although the artifact signature may be valid, it's NOT the signature required by this operational environment.
+
+### Scenario #9: Multiple Signatures
+
+Customers may require multiple signatures for the following scenarios:
+
+- Validate the artifact is the same as what the vendor provided.
+- Secondarily sign the artifact by the consuming company, attesting to it's validity within their production environment.
```
jonjohnsonjr: > attesting to it's validity

attesting to _its_ validity

scenarios.md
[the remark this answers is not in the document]
SteveLasker: I changed validated to verified.
You're teasing out some good questions. 
I imagined a scenario where the user can specify a list of signatures that are considered valid, or * for any signature. But, I didn't call out the scenario where the signature was required. I don't consider * to imply it must be signed, rather any signature would be valid. 
In the `fx:1.0` scenario, the customer could say they only accept signatures from the org that owns fx.
I was going to open an issue for the signature required scenario, but I think that's not something we need to make an open question. It seems like a completely valid scenario.

scenarios.md
```diff
@@ -0,0 +1,199 @@
...
+1. Deploy the artifact to a cluster that requires signatures:  
+  `$kubectl apply -f kubedeploy.yaml`
+  > **Note:** the above example presumes `kubedeploy.yaml` references the myregistry.domain.com/marketing/web:1 image
+
+**Implications of this requirement:**
+
+- Signatures can be validated based on the tag referenced, however the signature is not tied to a specific name. The artifact can be renamed without invalidating the signature.
+- This does not preclude tag locking scenarios, where a registry can provide users with the ability to lock a tag to a specific. If a tag is updated with another signed artifact, the tag is considered signed and follows the signature validation rules of the new signature. For example, if a base image `fx:1.0` is signed, and rebuilt with a patched version, the updated `fx:1.0` is new content, and a valid scenario.
+- If the tag for `fx:1.0` is redirected to another unsigned manifest, the signature validation of `fx:1.0` will fail as it's no longer signed. This is no different than signed binaries being updated or replaced on a users local computer.
+
```
samuelkarp: In addition to Scenario 2 ("Sign, Rename, Push, Validate"), I think it might be worth adding a scenario involving exporting and importing rather than pushing and pulling.  Scenario 1 ("Local Build, Sign, Validate) implies that signatures are not expected to be tied to a registry and don't need to be pushed to a registry in order for validation to occur, so it might be worth going further and specifying that the signature can be exported and imported independently of a registry as well.

scenarios.md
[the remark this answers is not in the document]
samuelkarp: > I imagined a scenario where the user can specify a list of signatures that are considered valid, or * for any signature. But, I didn't call out the scenario where the signature was required. I don't consider * to imply it must be signed, rather any signature would be valid.

We might be getting too far into the trust policy space here.  I think that we should be specifying the mechanisms for creating, storing, and distributing signatures as well as how the signatures can be validated, but not the specific logic of "should this signature be trusted given a particular policy".

It might be worth thinking about other ecosystems here.  For example, web browsers: different browsers implement different default policies (and many allow those policies to be further configured).  Some browsers may decide to trust a given CA, or decide to trust certificates generated by that CA only before a certain date, or decide only to trust certificates that have been published to a transparency log.  But these policies are not part of a spec; they're implementation decisions made by the browser implementers.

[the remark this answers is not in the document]
SteveLasker: Finding the balance between being able to enable policies of various products and specifying a policy is the goal. I've been trying to avoid how k8s, docker, or cloud/vendor specific implementations might work. Rather, here's the basic scenarios we're trying to enable.

scenarios.md
SteveLasker: I do like the scenario, but I don't yet know how to think about it as it pertains to the SBOM discussions that *might be* a better tool. 
I'd suggest the following: What if we add it after the base set are in. This way we can discuss that specific addition, while the common/agreed upon can be in for the spec to begin.

scenarios.md
SteveLasker: How do we tease out the question of an image being signed, without a registry name associated, but without the assumption it defaults to `docker.io`?

scenarios.md
```diff
@@ -0,0 +1,199 @@
...
+## Scenarios
+
+### Scenario #1: Local Build, Sign, Validate
+
+Prior to doing any deployment, a developer can test the: build, sign, validate scenario.
+
+1. Locally build a container image using a non-registry specific `name:tag`, such as:  
+  `$docker build hello-world:1`
+1. Locally sign `hello-world:1`
```
lahirurd: Isn't the statement "locally sign" (and the implication below of the developer having access to signing keys) overly restrictive and unnecessary? What's required here is that a signature is available at point. Where it's generated doesn't have to be defined in this document. This is also related to what @lumjjb suggested under "Open discussions".

scenarios.md
SteveLasker: I've added: 
> - Notary v2 supports a pattern for signing any type of artifact, from OCI Images, Helm Charts, Singularity to yet unknown types.

> This might be worth covering in another scenario, though I would expect the answer to be client-dependent and related to a particular trust policy.
This is where I'm also seeing the policy overlap. There's a balance between concrete scenarios and mentioning specific products. We use generalized company names. I'm wondering if we should start using generalized orchestrator names. Or, should we work on a concrete implementation along the way to assure we land this as we'd like?

scenarios.md
SteveLasker: grr, I'm trying to hard to catch these...

scenarios.md
```diff
@@ -0,0 +1,199 @@
...
+
+## OCI Images & Artifacts
+
+The [OCI TOB][oci-tob] has adopted [OCI Artifacts][artifacts-repo], generalizing container images as one of many types of artifacts that may be stored in a registry. Other artifact types currently include:
+
+* [Helm Charts][helm-registry]
+* [Singularity][singularity]
+* Car firmware updates, deployed from OCI Artifact registries
+
+## The Need for a Generalized Signing Solution
```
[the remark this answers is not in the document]
SteveLasker: Please take a look at the latest...

scenarios.md
SteveLasker: I struggle with this one. On one hand, I think we'd like to enable pull-->tag-->push to "just work", including any signatures that move with it.
On the other hand, we do need a general export/import API. If we solve the above, the export/import is more for bulk scenarios. 
Or, are you referring to something else, such as exporting/importing keys?

scenarios.md
SteveLasker: Hi @lahirurd,
What we're trying to capture here is the scenario is enabled locally, without any dependency on other services. For instance, signing shouldn't be limited to a registry, or a Notary signing server. "local" is contextual. For instance, local could also be on a remote VM or inside a container in a build scenario. The point is it can be done in isolation, with access to keys. 
For instance, a developer may only have access to dev keys they can sign locally on their laptop. The "production" keys are likely more secured, and restricted to roles that have more secure access, such as the identity of the build system.

scenarios.md
SteveLasker: I opened #4 to capture this, allowing us to resolve this part of the discussion for this pr.

scenarios.md
SteveLasker: Resolving this in lieu of the Definition & Terms @omieomye will be capturing.

scenarios.md
SteveLasker: I re-read with your question in mind and realized "referencing the registry they were pushed to" was vague. While helm, as a deployment artifact, does need to reference the artifacts being deployed, that details wasn't the intent.
Based on today's call, I've started changing references to "the orchestrator".

scenarios.md
samuelkarp: Should we change "local" to "offline" here to try and convey that this scenario is more about being able to sign images outside a registry than about necessarily having keys on laptops?

scenarios.md
samuelkarp: > I think we'd like to enable pull-->tag-->push to "just work", including any signatures that move with it.

Agreed

> On the other hand, we do need a general export/import API. If we solve the above, the export/import is more for bulk scenarios.

I don't know that we need to define an API.  The image spec already specifies the layout of a valid image, and container runtimes like containerd are already capable of writing that to disk as a single, exported file.

Import/export has use-cases beyond bulk scenarios.  It's useful for things like backup, including to offline storage like optical media.  It's also useful for providing a mechanism to transport images into an air-gapped system.  I think we probably want signatures to go along for the ride when the images are exported.

>  Or, are you referring to something else, such as exporting/importing keys?

That's another good point.  Key management and distribution isn't really covered in this scenario doc so far but we'll need to either declare it out-of-scope or cover it as well.  (I don't think we need to figure this out before merging this doc, however.)

scenarios.md
lahirurd: Okay, so I take it that the word "local" was used here to mean "outside a registry". 

My point was that this scenario would be just as valid if the dev used some external signing service to sign the image, without having access to the private key. Something like "2. Generate a signature for `hello-world:1`" might make it clearer that we aren't interested in how/where the signature is computed. My concern is that the two places where we describe signing in detail (this, and [scenario #3 step 3](https://github.com/notaryproject/requirements/pull/1/files#diff-685d3bc1e5cc6093032b54e6b54a81e1R95)) both calling out that the signature is generated locally with local keys, might be interpreted as that being the _only_ way to sign something in Notary v2.

If people feel that it's obviously not the case, please feel free to ignore me :)

lumjjb: I agree on the wording "local" seems to be restrictive. Would this be considered a separate scenario? - since it seems like they are targeting separate use-cases - developer vs enterprise.

scenarios.md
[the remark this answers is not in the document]
SteveLasker: please review the updated PR.

scenarios.md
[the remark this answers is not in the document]
SteveLasker: I removed the details on stable tags, and made a simpler reference in the new 6.2 mydb scenario.

scenarios.md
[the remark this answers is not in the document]
SteveLasker: I removed references to specific orchestrators or other projects.

SteveLasker:
I've incorporated feedback, and created a cleaner PR #7

PR #6, opened 2020-02-13
initial design options

samuelkarp:
This document is meant to capture a few different ideas around how to relate signatures to the objects we wish to sign.  This is meant as a discussion point, and not as a concrete proposal to use a particular approach (or even a recommendation).  This document is *not* exhaustive, and is meant as a companion to both @SteveLasker's [scenarios pull request][stevelasker-scenarios] and @justincormack's [signing overview][jc-signing-overview].

[jc-signing-overview]:   https://docs.google.com/document/d/1lffOYCDXBoxRumQre32yv1jUN2NfOkWugoDEJ_2DdEk/edit#
[stevelasker-scenarios]: https://github.com/notaryproject/requirements/pull/1

### DIFF
--- a/design-options.md
+++ b/design-options.md
@@ -1,0 +1,472 @@
+# Design proposals
+
+This document is meant to capture a few different ideas around how to relate
+signatures to the objects we wish to sign.  This is meant as a discussion
+point, and not as a concrete proposal to use a particular approach (or even a
+recommendation).  This document is *not* exhaustive, and is meant as a companion
+to both @SteveLasker's [scenarios pull request][stevelasker-scenarios] and
+@justincormack's [signing overview][jc-signing-overview].
+
+## Terms
+
+Options here make use of the following terms:
+
+* [OCI Image Manifest][oci-manifest]: A description of a container image,
+  including the constituent layers and configuration as content-addressable
+  references ([descriptors][oci-descriptor]).
+* [OCI Descriptor][oci-descriptor]: A structure describing content, including
+  the media type, a content-addressable digest, the size, and other properties.
+  Descriptors are used to describe layers and configuration in a manifest.
+* [OCI Image Index][oci-image-index]: A higher-level manifest which points to
+  specific [image manifests][oci-manifest], typically used to describe
+  platform-specific (e.g., architecture- and operating-system specific)
+  images that can be identified collectively and referred to together.  The
+  specific image manifiests are identified by modified
+  [descriptors][oci-descriptor] with additional properties and restrictions.
+* [OCI Annotations][oci-annotations]: A key-value map that can be associated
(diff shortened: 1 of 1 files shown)


reasonerjt:
Thanks for the write up!

It seems all the options in the PR are trying to add signature by adding the artifacts and signature into an index.

I may have misunderstanding, but with such approach, if a user knows the `repo:tag` or `repo:digest` of an artifact, how does he know whether this artifact is signed?  Or if he wants to move the artifact to another registry, how does he know the digest of the index that contains the signature so the signature can also be moved?

samuelkarp:
> It seems all the options in the PR are trying to add signature by referencing the artifacts and signature in an index.

That's correct; this is the set of initial ideas I've put together.  However, please do not think that this is exhaustive or that these are the only acceptable approaches; if you have other suggestions please feel free to add them!

> if a user knows the `repo:tag` or `repo:digest` of an artifact, how does he know whether this artifact is signed?

If a user knows the name of the artifact (`repo:tag`), that is generally expected to resolve to the index, which will then contain the signature.

For a user trying to pull by the content-address (`repo:digest`), the answer depends on what the digest is.  If it's the digest of the index, they'll see the signature.  If it's the digest of the manifest, they won't see the signature.  However, when you are pulling by digest, you already know what content you are expecting to receive and the signature may not tell you any additional information.

> Or if he wants to move the artifact to another registry, how does he know the digest of the index that contains the signature so the signature can also be moved?

To move an image/artifact, I would expect that a user would reference the index by name (`repo:tag`), retrieve all the references in each descriptor, and push all of those to the new registry.

reasonerjt:
@samuelkarp thanks for the reply, so that means it will introduce a change in the flow. 

For the images that are not signed, we expect users use the URI of the image to access it.
For those that are signed we expect users to use the URI of the index to access it but never use the URI of the image?

IMHO, this design may cause some difficulty for deployer and registry implementers.  We'd better have a way to find the index digest, which has the signature of imageA using the digest of imageA, I don't have a quick idea how to implement that within a native oci-registry though.

jonjohnsonjr:
> For the images that are not signed, we expect users use the URI of the image to access it.
For those that are signed we expect users to use the URI of the index to access it but never use the URI of the image?

I would expect registry clients to understand how to take an index and resolve it to an image. They already need to do this based on platform information for multi-platform images, so I strongly suspect we can roll this out in a way that is backwards compatible with existing clients.

SteveLasker:
These are more notes, great notes, but not something we were going to merge as a doc. Should we move this to an issue or hackmd doc?

samuelkarp:
Reopened as https://github.com/notaryproject/requirements/pull/31 after the branch deletion issue.

PR #8, opened 2020-02-13
End to end scenarios, accounting for PR #1 feedback

SteveLasker:


scenarios.md
```diff
@@ -0,0 +1,225 @@
...
+1. The index, and its contents are pushed to a registry:  
+  `$ docker push wabbitnetworks.example.com/networking/net-monitor:1.0-alpine`  
+  `$ deploy-tool push wabbitnetworks.example.com/networking/net-monitor:1.0-deploy`  
+  `$ oci-tool push wabbitnetworks.example.com/networking/net-monitor:1.0`
+1. The artifacts are deployed to a production orchestrator.
+1. The orchestrator verifies the artifacts are signed by a set of specifically trusted keys. Unsigned artifacts, or artifacts signed by non-trusted keys are rejected.
+
+**Implications of this requirement:**
+
+- Keys for signing are securely retrieved by build systems that create & destroy the environment each time.
```
lumjjb: Hmm... is the ability of access to keys here mean actual access to key bytes? Or ability to sign with a particular key?

I think the ability for key bytes to be explicitly be accessible is restrictive. My thought process is around key based protocols like KMIP are moving towards ability to sign something remotely. And the same would be for PKCS11 for HSMs (and services built around it). 

The implication for this would be signatures stored require support for a variety of standard formats, like PKCS7. Or have some library that has the ability to move signatures in/out of the format of notaryv2 and pkcs7.

scenarios.md
```diff
@@ -0,0 +1,225 @@
...
+By developing a generalized solution, artifact authors may develop their unique artifact types, allowing them to leverage Notary for signing and OCI Compliant registries for distribution.
+
+## OCI Images & Artifacts
+
+The [OCI TOB][oci-tob] has adopted [OCI Artifacts][artifacts-repo], generalizing container images as one of many types of artifacts that may be stored in a registry. Other artifact types currently include:
+
+* [Helm Charts][helm-registry]
+* [Singularity][singularity]
+* Car firmware updates, deployed from OCI Artifact registries
+
```
JustinCappos: It would be good to have an explicit threat model.  I've plopped in a draft to help get things started.

```suggestion
## Threat model

It is assumed that an attacker may perform one or more the following actions:

1. intercept and alter network traffic
2. compromise some set of weak crypto algorithms which are supported in some legacy cases
3. compromise a repository, including gaining access to use any keys stored on the repository
4. compromise a signing key, for example due to malicious action or accidental disclosure by the key owner
5. compromise a step in the software supply chain.  This can happen in many different ways, such as by gaining access to the server, compromising the software used in the step of the supply chain, passing different software to a subsequent step than what was intended, or causing an operator to make an error in a step. 

While it is not always possible to protect against all scenarios, the system should to the extent possible mitigate and/or reduce the damage caused by a successful attack, detect the occurrence of an attack and notify appropriate parties, yet remain usable for parties operating the system.  Furthermore, the system should recover from successful attacks in a way that presents low operational overhead and risk to users.

```

scenarios.md
```diff
@@ -0,0 +1,225 @@
...
+1. Locally sign `net-monitor:dev` 
+1. Run the image on the developers local machine which is configured to only accept signed images. 
+  `$ docker run net-monitor:dev`
+
+**Implications of this requirement:**
+
+- The developer has access to signing keys. How they get the keys is part of the usability spec.
+- The local environment has a policy by which it states the set of keys it accepts.
+- The signing and validation of artifacts does not require a registry. The local host can validate the signature using the public keys it accepts.
+- The key used for validation may be hosted in a registry, or other accessible location.
```
JustinCappos: This feels more like you're proposing a solution than talking about scenarios / requirements.  Is this appropriate for the document?  

Implications in general seems to be a mix of actual implications / requirements and details from a proposed implementation intertwined.   Can we separate this out so that it is just requirements?  It's weird to read about details of a proposed implementation with no description.

scenarios.md
```diff
@@ -0,0 +1,225 @@
...
+1. The `mydb:1.0` is tagged with a unique id `mydb:1.0-202002131000` and signed with an ACME Rockets production key.
+1. The retagged image, with both the mydb and ACME Rockets signatures are copied to a prod registry/repository.
+1. The release management system deploys the new `mydb:1.0-202002131000` image.
+1. The production orchestrator validates it's signed with the Acme Rockets production key.
+
+**Implications of this requirement:**
+
+- Multiple signatures, including signatures from multiple sources can be associated with a specific artifact.
+- Original signatures are maintained, even if the artifact is re-tagged.
+
```
JustinCappos:
```suggestion

#### Scenario 7: A repository compromise occurs

An attacker manages to compromise a repository and gain access to all keys on the repository.  This includes access to sign artifacts using keys sorted on the repository's HSM.  

**Implications of this requirement:**

1. The potential damage/risk to users in this case must be limited
2. There must be a secure way for users to recover to a known, secure state and verify this has occurred even in the face of an attacker that can act as a man-in-the-middle on the network.

```

I'll add more of these later, but wanted feedback here first.

samuelkarp: > This includes access to sign artifacts using keys sorted on the repository's HSM.

I think this is the first scenario talking about the registry having an HSM or stored keys.

JustinCappos: That's a great point.  My text should indicate that it provides access to keys stored in hardware if they exist.  I certainly don't mean to imply there must (or even should) be a HSM.  I more meant that it should not be assumed to adequately protect a key in this case.

scenarios.md
SteveLasker: The main point I was trying to make here was private signing keys are accessible by external entities, and not directly part of the notary workflow. Nor, should we assume a developers workstation, or build system must be heavily configured. Details on the types of keys and such, I've left out of the scenario.
This also hints at the vendor agnostic components as well. We had a cloud/vendor requirement that allows the workflows to take advantage of the vendor/cloud providers keyvault solution. 
Is there a suggestion for how I can reword this to not imply a constraint on how things are signed?

scenarios.md
SteveLasker: These are great! Do we think we should add to scenarios, with a threat model section, or a separate `threat-model-scenarios.md` 
After trying to get a single doc merged, I'm looking for smaller docs we can agree upon individually. If we think we can agree easily, I'm all for merging.

Please vote: 
:+1: `:+1:` - merge in this scenarios doc
:-1: `:-1:` - create a `threat-model-scenarios.md`

scenarios.md
SteveLasker: I really like the scenario, and it builds on the threat requirement section. If you can reword it to avoid referencing a specific implementation, such as HSM which doesn't work well in cloud environments, we can merge this in as well. It's not to say we wouldn't support HSMs if the customers environment can, rather it's a specific implementation that I'm trying to avoid.

I have been assuming the private signing keys are NOT part of Notary v2, rather externally stored. 
1. so a registry compromise doesn't compromise all signed objects
1. as cloud providers and vendors each have their own, integrated secure key management system. 

There's a separate requirement, from a requirements pr, that talks to Notary being able to integrate with, through some sort of extension model, a cloud providers solution. For a reference implementation, we could use https://github.com/hashicorp/vault.

I'm happy to copy/paste the above text, and modify. Or, commit a slightly modified suggestion.

scenarios.md
justincormack: I think separate files are better, editing PRs is kind of painful anyway.

scenarios.md
lumjjb: Hmm... how about  "The CI system performs signing of the artifact via locally available keys or through a remote signing services"?

PR #9, opened 2020-02-17
Definitions & terms

SteveLasker:
A set of definitions & terms to be used across repos. References can be made as `[fingerprint](./definitions-terms.md/#fingerprint)`


SteveLasker:
@samuelkarp, an extraction of terms from #6

definitions-terms.md
```diff
@@ -0,0 +1,108 @@
...
+Reference: [OCI Artifacts][oci-artifact]  
+![](./media/artifact-ref.png)  
+
+
+## Fingerprint
+
+A short identifier of a given public key.  
+Reference: [Fingerprint][fingerprint]
+
+## Multi-Tenant Registry
```
samuelkarp: I'd vote for excluding this definition for now.  This is getting a little slippery with codifying the practice of registry operators instead of defining a term that we intend to use going forward.

definitions-terms.md
```diff
@@ -0,0 +1,108 @@
...
+* [OCI Index](#oci-index)
+* [OCI Manifest](#oci-manifest)
+* [Registry](#registry)
+* [SBoM](#sbom)
+* [Subject](#subject)
+* [Tag](#tag)
+
+## Artifact / Image
+
+Artifacts are a generalization of how images are stored in an instance of the [OCI Distribution Spec][oci-distribution-spec]. An [OCI Image][oci-image] is a type of [OCI Artifact][oci-artifact]. When referencing artifacts, within a [fully qualified reference](#registry-fully-qualified-reference), the artifact is right most element of the namespace, combined with the `:tag`.
```
samuelkarp: > the artifact is right most element of the namespace

This isn't really accurate.  The right-most element of the name is still just part of the name, and can resolve to anything the registry stores.  This could be an index, an image manifest, or an artifact.

We should probably include a definition for a name, referencing the [distribution spec](https://github.com/opencontainers/distribution-spec/blob/master/spec.md#overview):

> The rules for a repository name are as follows:
>
> 1. A repository name is broken up into *path components*. A component of a repository name MUST begin with one or more lowercase alpha-numeric characters. Subsequent lowercase alpha-numeric characters are OPTIONAL and MAY be separated by periods, dashes or underscores. More strictly, it MUST match the regular expression `[a-z0-9]+(?:[._-][a-z0-9]+)*.`
> 2. If a repository name has two or more path components, they MUST be separated by a forward slash ("/").
> 3. The total length of a repository name, including slashes, MUST be less than 256 characters.

definitions-terms.md
```diff
@@ -0,0 +1,108 @@
...
+OCI spec reference: [OCI Annotations][oci-annotations]
+
+## OCI Descriptor
+
+A structure describing content, including the media type, a content-addressable digest, the size, and other properties. Descriptors are used to describe layers and configuration in a manifest.  
+OCI spec reference: [OCI Descriptor][oci-descriptor]
+
+## OCI Index
+
+A higher-level collection of [image manifests][oci-manifest], or other [oci indexes][oci-index], typically used to describe platform-specific (architecture and operating-system) images that can be identified collectively and referred to together. The specific image manifests are identified by modified [descriptors][oci-descriptor] with additional properties and restrictions. [OCI Artifacts][oci-artifact] takes this a step further and provides for an OCI Index to reference a collection of artifacts, such as an image, a deployment chart, its SBoM and deployment configuration references.
```
samuelkarp: > [OCI Artifacts][oci-artifact] takes this a step further and provides for an OCI Index to reference a collection of artifacts, such as an image, a deployment chart, its SBoM and deployment configuration references.

Suggestion: [OCI Artifacts][oci-artifact] are another use-case for indexes where an OCI Index can reference a collection of artifacts, such as an image, a deployment chart, its SBoM and deployment configuration references.

definitions-terms.md
```diff
@@ -0,0 +1,108 @@
...
+OCI spec reference: [OCI Descriptor][oci-descriptor]
+
+## OCI Index
+
+A higher-level collection of [image manifests][oci-manifest], or other [oci indexes][oci-index], typically used to describe platform-specific (architecture and operating-system) images that can be identified collectively and referred to together. The specific image manifests are identified by modified [descriptors][oci-descriptor] with additional properties and restrictions. [OCI Artifacts][oci-artifact] takes this a step further and provides for an OCI Index to reference a collection of artifacts, such as an image, a deployment chart, its SBoM and deployment configuration references.  
+OCI spec reference: [OCI Image Index][oci-index]
+
+## OCI Manifest
+
+A description of an [artifact][oci-artifact] within an instance of an [OCI Distribution][oci-distribution-spec] complaint registry. The manifest references optional configuration and blobs (layers) as content-addressable references ([descriptors][oci-descriptor]).  Artifacts may be container images, helm charts or other artifact types that may be signed and stored in a registry.
```
samuelkarp: The manifest is defined by the [image spec](https://github.com/opencontainers/image-spec/blob/master/manifest.md), not by the distribution spec.  We should remove the language about "within an isntance of an OCI Distribution compliant registry", since a manifest is valid both inside and outside a registry.

definitions-terms.md
```diff
@@ -0,0 +1,108 @@
...
+
+## OCI Manifest
+
+A description of an [artifact][oci-artifact] within an instance of an [OCI Distribution][oci-distribution-spec] complaint registry. The manifest references optional configuration and blobs (layers) as content-addressable references ([descriptors][oci-descriptor]).  Artifacts may be container images, helm charts or other artifact types that may be signed and stored in a registry.    
+OCI spec reference: [OCI Image Manifest][oci-manifest]
+
+## Registry
+
+A registry is a collection of [OCI Indexes][oci-index] and [OCI Manifestes][oci-manifest] for a specific org or entity.  
+A unique registry may be uniquely identified with a domain
```
samuelkarp: Same comment as above about naming; should be broken out into a separate definition that comes from the distribution spec.

definitions-terms.md
```diff
@@ -0,0 +1,108 @@
...
+## Registry
+
+A registry is a collection of [OCI Indexes][oci-index] and [OCI Manifestes][oci-manifest] for a specific org or entity.  
+A unique registry may be uniquely identified with a domain  
+![](./media/unique-registry-domain-ref.png)  
+ Or a root namespace:  
+![](./media/unique-registry-namespace-ref.png)
+
+
+## Registry Namespace
```
samuelkarp: I'd remove this as namespaces are not a specific part of the definition of a name.  You could include a definition of a "path component", since that's the term used in the distribution spec.

definitions-terms.md
SteveLasker: This looks like the definition I wrote for repo, below.

definitions-terms.md
SteveLasker: How do we articulate the difference between a unique registry, and the registry as a whole? What is "your registry" on docker hub, or within AWS? Just think we need to articulate the difference.

definitions-terms.md
SteveLasker: The image spec and distribution spec have things intertwined between them. We have a work-item to detangle these, likely refactoring the schemas as a top level object, that artifacts use to represent objects that get persisted in a distribution. The image-spec is one of many artifact types that use these schemas. 
- [Need a Working Group for Updating the OCI distribution-spec](https://github.com/opencontainers/artifacts/issues/9)
- [Need a working group for updating the image spec](https://github.com/opencontainers/artifacts/issues/10)

definitions-terms.md
SteveLasker: Are you suggesting we use the term path, instead of namespace? That makes sense. Although, I'd still want to separate the element that doesn't include the artifact name. 
Here's the larger post that goes into more detail:  
[Registry Names, Namespaces, Images, Artifacts & Tags](https://stevelasker.blog/2020/02/17/registry-namespace-repo-names/)

definitions-terms.md
samuelkarp: I'd vote on holding off on articulating the difference at all until we need it.  I'm not convinced that it's really necessary for the discussion.

definitions-terms.md
samuelkarp: I still think we should remove the language about "within an instance of an OCI Distribution compliant registry".  The manifest is valid both inside and outside the registry; the location of the manifest does not play into its validity.

definitions-terms.md
samuelkarp: "path component" is the terminology used in the distribution spec, so I would bias toward that.  I don't really see that we need to describe all of the components of a name, though.  It's not clear to me that we need a distinction between "namespaces" and "images", for example.  One of the goals of this project is for the signatures to be independent of the location of the images/artifacts, and right now the name is synonymous with the location.

definitions-terms.md
SteveLasker: After reviewing the registry definition, it already covered unique orgs by namespace or domain. This basically covers the thing I was trying to cover for multi-tenant registries.

definitions-terms.md
SteveLasker: Lets have this conversation on Monday.

definitions-terms.md
SteveLasker: A good conversation for Monday as well.
The issue is what I describe here: https://stevelasker.blog/2020/02/17/registry-namespace-repo-names/
> Potential life cycle of the web:v1.0 artifact. In the following example, the same web:v1.0 artifact is copied across several repos, using different namespaces, as it makes it way through production, to an archived state.

1. org.example.com/**dev/blue-team**/web:v1.0
1. org.example.com/**staging/marketing**/web:v1.0
1. org.example.com/**prod/marketing**/web:v1.0
1. org.example.com/**archive/marketing**/web:v1.0

We need a way for the `web:1.0` artifact to move between repos, while maintaining the same signature.
We've already discussed moving across registries:
org-archive.example.com/**marketing**/web:v1.0

What I'd like to accomplish is defining the path as a separate element of the artifact. An artifact can move between paths without changing.

PR #12, opened 2020-02-21
Adding a few more compromise-related scenarios...

JustinCappos:
Signed-off-by: Justin Cappos <justincappos@gmail.com>

### DIFF
--- a/scenarios.md
+++ b/scenarios.md
@@ -221,7 +221,7 @@ It is discovered that a server used to construct software has potentially been c
 1. There should be a means to warn about and/or stop future container deployments that were built using this server.
 
 #### Scenario 7.4: A Crypto Algorithm Is Deprecated
-A weakness is discovered in a widely used cryptographic algorithm and a decisions is made to deprecate its use in favor of one or more newer algorithms.  It should be possible to migrate smoothly to the new algorithm(s), despite the fact that adoption by developers will adopt the new signing algorithm(s) piecemeal.
+A weakness is discovered in a widely used cryptographic algorithm and a decisions is made to deprecate its use in favor of one or more newer algorithms.  It should be possible to migrate smoothly to the new algorithm(s), despite the fact that developers will adopt the new signing algorithm(s) piecemeal.
 
 **Implications of this requirement:**
 1. There must be a means to support using multiple different cryptographic algorithms of varying types.  The client and server must be able to use them in coordination, without requiring a "flag day" where everyone swaps over.
--- a/scenarios.md
+++ b/scenarios.md
@@ -218,7 +218,7 @@ It is discovered that a server used to construct software has potentially been c
 
 **Implications of this requirement:**
 1. There should be auditable metadata to understand when a build was created on this server.  Furthermore, there should be adequate information to understand when and/or what version of the build software was used, as this may be relevant for auditing.
-1. There should be a means to warn about and/or stop future container deployments what were built using this server.
+1. There should be a means to warn about and/or stop future container deployments that were built using this server.
 
 #### Scenario 7.4: A Crypto Algoritm Is Deprecated
 A weakness is discovered in a widely used cryptographic algorithm and a decisions is made to deprecate its use in favor of one or more newer algorithms.  It should be possible to migrate smoothly to the new algorithm(s), despite the fact that adoption by developers will adopt the new signing algorithm(s) piecemeal.
--- a/scenarios.md
+++ b/scenarios.md
@@ -194,6 +194,42 @@ A deployment requires a mydb image. The mydb image is routinely updated for secu
 1. The release management system deploys the new `mydb:1.0-202002131000` image.
 1. The production orchestrator validates it's signed with the Acme Rockets production key.
 
+
+
+### Scenario #7: Malicious and Erroneous Actions
+
+#### Scenario 7.1: A Repository Compromise Occurs
+An attacker manages to compromise a repository and gain access to the repository as an administrator.   
+**Implications of this requirement:**
+1. The potential damage/risk to users in this case must be limited
+1. There must be a secure way for users to recover to a known, secure state and verify this has occurred even in the face of an attacker that can act as a man-in-the-middle on the network.
+
+#### Scenario 7.2: A Developer Discloses Their Key and/or Credentials
+A developer accidentally discloses the key they use to certify their software is authentic.  For example, the developer may have unintentionally checked their key into a github repository.  
+
+Furthermore, the developer may have scripted the means by which software is added to a repository and this script (and related credentials / secrets) have all been disclosed.
+
+**Implications of this requirement:**
+1. There should be a means to get back to a secure state.  It should not be the case that the first person to notice this action now can irrevocably control the software software distribution mechanism used by the developer.
+1. There should be some means to know what actions have been performed using the developer's key and/or credentials and when those actions occurred.
+
+#### Scenario 7.3: A Step in the Supply Chain Is Compromised
+It is discovered that a server used to construct software has potentially been compromised.  This means that the software coming from this server could be at risk.  For example, a build server may have been compromised and now the container images built there may have been backdoored.
+
+**Implications of this requirement:**
+1. There should be auditable metadata to understand when a build was created on this server.  Furthermore, there should be adequate information to understand when and/or what version of the build software was used, as this may be relevant for auditing.
+1. There should be a means to warn about and/or stop future container deployments what were built using this server.
+
+#### Scenario 7.4: A Crypto Algoritm Is Deprecated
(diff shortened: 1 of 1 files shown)


scenarios.md
```diff
@@ -194,6 +194,42 @@ A deployment requires a mydb image. The mydb image is routinely updated for secu
...
+**Implications of this requirement:**
+1. There should be a means to get back to a secure state.  It should not be the case that the first person to notice this action now can irrevocably control the software software distribution mechanism used by the developer.
+1. There should be some means to know what actions have been performed using the developer's key and/or credentials and when those actions occurred.
+
+#### Scenario 7.3: A Step in the Supply Chain Is Compromised
+It is discovered that a server used to construct software has potentially been compromised.  This means that the software coming from this server could be at risk.  For example, a build server may have been compromised and now the container images built there may have been backdoored.
+
+**Implications of this requirement:**
+1. There should be auditable metadata to understand when a build was created on this server.  Furthermore, there should be adequate information to understand when and/or what version of the build software was used, as this may be relevant for auditing.
+1. There should be a means to warn about and/or stop future container deployments what were built using this server.
```
jonjohnsonjr: nit: `deployments what were built` -> `deployments that were built`

scenarios.md
```diff
@@ -194,6 +194,42 @@ A deployment requires a mydb image. The mydb image is routinely updated for secu
...
+
+#### Scenario 7.4: A Crypto Algoritm Is Deprecated
```
jonjohnsonjr: nit: Algoritm -> Algorithm

scenarios.md
```diff
@@ -194,6 +194,42 @@ A deployment requires a mydb image. The mydb image is routinely updated for secu
...
+A weakness is discovered in a widely used cryptographic algorithm and a decisions is made to deprecate its use in favor of one or more newer algorithms.  It should be possible to migrate smoothly to the new algorithm(s), despite the fact that adoption by developers will adopt the new signing algorithm(s) piecemeal.
```
jonjohnsonjr: nit: a decisions -> a decision

`the fact that adoption by developers will adopt` -> `the fact that developers will adopt` ?

scenarios.md
JustinCappos:
```suggestion
1. There should be a means to warn about and/or stop future container deployments that were built using this server.
```

scenarios.md
JustinCappos:
```suggestion
#### Scenario 7.4: A Crypto Algorithm Is Deprecated
```

scenarios.md
JustinCappos: > nit: a decisions -> a decision
> 
> `the fact that adoption by developers will adopt` -> `the fact that developers will adopt` ?

```suggestion
A weakness is discovered in a widely used cryptographic algorithm and a decisions is made to deprecate its use in favor of one or more newer algorithms.  It should be possible to migrate smoothly to the new algorithm(s), despite the fact that developers will adopt the new signing algorithm(s) piecemeal.
```

scenarios.md
```diff
@@ -194,6 +194,42 @@ A deployment requires a mydb image. The mydb image is routinely updated for secu
 1. The release management system deploys the new `mydb:1.0-202002131000` image.
 1. The production orchestrator validates it's signed with the Acme Rockets production key.
 
+
+
+### Scenario #7: Malicious and Erroneous Actions
+
+#### Scenario 7.1: A Repository Compromise Occurs
+An attacker manages to compromise a repository and gain access to the repository as an administrator.
```
SteveLasker: This is a great scenario, but seems incomplete as it doesn't seem actionable as how would the state be known? It might just be the wording. Are you saying, the owner learns one or more repos have been compromised and needs to take action to re-secure it?

scenarios.md
```diff
@@ -194,6 +194,42 @@ A deployment requires a mydb image. The mydb image is routinely updated for secu
...
+### Scenario #7: Malicious and Erroneous Actions
+
+#### Scenario 7.1: A Repository Compromise Occurs
+An attacker manages to compromise a repository and gain access to the repository as an administrator.   
+**Implications of this requirement:**
+1. The potential damage/risk to users in this case must be limited
+1. There must be a secure way for users to recover to a known, secure state and verify this has occurred even in the face of an attacker that can act as a man-in-the-middle on the network.
+
+#### Scenario 7.2: A Developer Discloses Their Key and/or Credentials
+A developer accidentally discloses the key they use to certify their software is authentic.  For example, the developer may have unintentionally checked their key into a github repository.
```
SteveLasker: great scenario. It's the uh-o moment. Now, what do they do?

scenarios.md
```diff
@@ -194,6 +194,42 @@ A deployment requires a mydb image. The mydb image is routinely updated for secu
...
+#### Scenario 7.2: A Developer Discloses Their Key and/or Credentials
+A developer accidentally discloses the key they use to certify their software is authentic.  For example, the developer may have unintentionally checked their key into a github repository.  
+
+Furthermore, the developer may have scripted the means by which software is added to a repository and this script (and related credentials / secrets) have all been disclosed.
+
+**Implications of this requirement:**
+1. There should be a means to get back to a secure state.  It should not be the case that the first person to notice this action now can irrevocably control the software software distribution mechanism used by the developer.
+1. There should be some means to know what actions have been performed using the developer's key and/or credentials and when those actions occurred.
+
+#### Scenario 7.3: A Step in the Supply Chain Is Compromised
```
SteveLasker: Can we bump this out to it's own scenario? Scenario 7 feels focused on the credentials have been compromised; now what? 
This is another great set of scenarios where dependent content is compromised. It could be included packages, or the compiler used to compile safe packages does something unsafe.

scenarios.md
```diff
@@ -194,6 +194,42 @@ A deployment requires a mydb image. The mydb image is routinely updated for secu
...
+It is discovered that a server used to construct software has potentially been compromised.  This means that the software coming from this server could be at risk.  For example, a build server may have been compromised and now the container images built there may have been backdoored.
+
+**Implications of this requirement:**
+1. There should be auditable metadata to understand when a build was created on this server.  Furthermore, there should be adequate information to understand when and/or what version of the build software was used, as this may be relevant for auditing.
```
SteveLasker: While I'd like to keep the scenario, can we refer to this as something out of scope for Notary v2? I'm suggesting we focus Notary v2 on signing content. The signed content may be additional content that includes an SBoM. But, the details of the SBoM, or which SBoM project the user might include is outside our Notary v2 scope.
What is in-scope here is: 
A compromise to the supply chain has been identified. The user must be able to account for all content signed by entities that are no longer considered trusted.

scenarios.md
SteveLasker: Another great scenario, that deserves it's own major number. I believe we're at 9?

scenarios.md
samuelkarp: I'm not really sure we should bring in repository management and authorization scenarios to this, especially as there is not a common permissions model among the existing registries that are available today.

scenarios.md
samuelkarp: I think that designing an audit mechanism for image creation seems related to the overall security of the repository, but distinct from signing.

scenarios.md
```diff
@@ -194,6 +194,42 @@ A deployment requires a mydb image. The mydb image is routinely updated for secu
...
+1. There should be a means to warn about and/or stop future container deployments that were built using this server.
+
+#### Scenario 7.4: A Crypto Algorithm Is Deprecated
```
samuelkarp: Can this be broken out into its own top-level scenario?  This is important for us to get right.

scenarios.md
JustinCappos: Yes, this is my intent here.  I'm not sure I understand the distinction you are trying to make.  Can you propose an edit that clarifies?

scenarios.md
JustinCappos: > I think that designing an audit mechanism for image creation seems related to the overall security of the repository, but distinct from signing.

This may be distinct from the mechanism of signing, but my sense is that we should capture the situations and properties we want in scenarios and then figure out how to map them to mechanisms later.  We likely will declare a lot of things out of scope, but doing this exercise first makes it clear how we're protecting things and where we rely on others...

scenarios.md
JustinCappos: Sure, I'm fine with this.  Feel free to edit / tweak however you like.  I'm happy to have others come in and edit / add on too.

scenarios.md
JustinCappos: I'd be interested to chat more and see if there is a way to capture these models and see how they may change things in the case of a compromise.  Ideally our design would work well under all current and envisioned permissions models...

scenarios.md
JustinCappos: Sounds fine.  Feel free to change how you like.  I'm happy to accept the edit or have others do so.

SteveLasker:
Closing this in lieu of PR #15

PR #15, opened 2020-03-05
Scenario updates, incorporating scenarios, addressing PR feedback

SteveLasker:
Incorporates feedback from #12


scenarios.md
```diff
@@ -199,6 +205,63 @@ A deployment requires a mydb image. The mydb image is routinely updated for secu
...
+
+**Implications of this requirement:**
+
+1. All Notary v2 implementations support key revocation as part of their implementation to assure signed content is still valid content.
+1. Registry operators routinely check for revoked keys and remediate the exploited content. 
+1. Notary v2 clients routinely check for revoked keys and block the content.
+
+### Scenario #9: A Crypto Algorithm Is Deprecated
+
+A weakness is discovered in a widely used cryptographic algorithm and a decisions is made to deprecate its use in favor of one or more newer algorithms. It should be possible to migrate smoothly to the new algorithm(s), despite the fact that developers will adopt the new signing algorithm(s) piecemeal.
```
jonjohnsonjr:
```suggestion
A weakness is discovered in a widely used cryptographic algorithm and a decision is made to deprecate its use in favor of one or more newer algorithms. It should be possible to migrate smoothly to the new algorithm(s), despite the fact that developers will adopt the new signing algorithm(s) piecemeal.
```

SteveLasker:
Can we get some :eyes: on this so we can close these out?

SteveLasker:
@JustinCappos, can you ack that we've captured the scenarios portion

SteveLasker:
> The document is really design heavy now. This covers a few cases but things like 7 only make sense in the context of a specific design, whereas 8 and 9 are much more general.

Are you referring to the scenario or implications? Do you have a proposed edit so we can merge?

JustinCappos:
> > The document is really design heavy now. This covers a few cases but things like 7 only make sense in the context of a specific design, whereas 8 and 9 are much more general.
> 
> Are you referring to the scenario or implications? Do you have a proposed edit so we can merge?

It's more of a conceptual change to parts of the document.  Some of the content like this feels like it is quite specific to the technologies and not the goals / scenarios.  I'm just going to cut and paste a few example, but the conceptual difference is something I'm seeing throughout.

"The ACME Rockets environment enforces various company policies prior to any deployment, evaluating the content in the `SBoM`. The policy manager trusts the content within the SBoM is accurate, because they trust artifacts signed by wabbit-networks. The `src` content isn't evaluated at deployment time and can be left within the registry.
1. Once the policy manager completes its validation, the deployment to the hosting environment is initiated. The `SBoM` is no longer needed, allowing the `image` to be deployed separately. A `deploy` artifact, referencing a specific configuration definition, may also be signed and saved, providing a historical record of what was deployed. The hosting environment also validates content is signed by trusted entities."

"The local environment has a policy by which it states the set of keys it accepts.
- The signing and validation of artifacts does not require a registry. The local host can validate the signature using the public keys it accepts.
- The key used for validation may be hosted in a registry, or other accessible location.
- The lack of a registry name does not infer docker.io as a default registry."

"Users may reference the `sha256` digest directly, or the `artifact:tag`. While tag locking is not part of the [OCI Distribution Spec][oci-distribution], various registries support this capability, allowing users to reference human readable tags, as opposed to long digests. Either reference is supported with Notary v2, however it's the unique manifest that is signed."

I don't really know if this should be moved to a separate document, if general text should replace it, or if I'm just keying in on things that I shouldn't.  

Perhaps @justincormack should weigh in also and say whether what I'm objecting to makes any sense.

SteveLasker:
> I don't really know if this should be moved to a separate document, if general text should replace it, or if I'm just keying in on things that I shouldn't.

@JustinCappos I think we're possibly seeing a view from different perspectives. There's a perspective that we need a signature solution to artifacts in a registry. A solution that can move with the artifact as it moves within and across registries. That does define a sandbox by which we're working within. In the call last week we discussed how we may need to tweak the sandbox, but we are scoping to storing signatures in a registry.
The work you've done in Tuf is about taking validation to a another confidence level. We've been talking about scope creep quite a bit, and how we can minimize that scope creep, working from a base that covers the scenarios and building upon it. For instance the generic SBoM references may refer to a Tuf implementation, a RedHat implementation or others. 
Now I know @justincormack was thinking it might be more broad, and he was writing up something. This seems like the conversation we should be having. What is the scope of what we want to carve out here.

scenarios.md
```diff
@@ -199,6 +205,63 @@ A deployment requires a mydb image. The mydb image is routinely updated for secu
...
+
+A developer accidentally discloses the private key they use to certify their software is authentic. 
+
+1. A developer accidentally checks their private key into a github repository.
+1. A developer references a compromised package that searches for private keys within the build system and `curl`s it a location of an attacker.
+1. A developer may have scripted the means by which secured content is added to a registry and the script (and related credentials & secrets) have all been disclosed.
+1. Whether 5 minutes or 5 days, the developer must assume the pubic key was copied and new content may be pushed to this, or any other registry claiming it represents the identify of the private key.
+1. The attacker may not have credentials to push content to a the registry in focus, but they may push similarly named content to other public registries, misrepresenting the identity.
+1. The developer revokes their public key, indicating it's no longer valid.
+1. The developer issues a new public key and re-signs their content.
```
directionless, line 239: Forgive me if this is covered elsewhere, I'm quite new here -- what are the implications, and verifications, needed for this new key authenticity? I don't see it in the implications below.

scenarios.md
```diff
@@ -18,9 +18,31 @@ The [OCI TOB][oci-tob] has adopted [OCI Artifacts][artifacts-repo], generalizing
...
 This document serves as the requirements and constraints of a generalized signing solution. It focuses on the scenarios and needs, and very specifically avoids any reference to other projects or implementations. As our working group forms a consensus on the requirements, the group will then transition to a spec.
 
+The following represent the core goals of Notary v2:
+
+1. Registry Native
+   * Signatures and artifacts co-located for easier and secure management
+1. Secure
+   * Attesting to its authenticity and/or certification
+   * No trust on first use, no implicit permissions on rotated keys, secure private keys and PKI
+1. Multiple & Indepdent Signatures
```
trishankatdatadog, line 28:
```suggestion
1. Multiple & Independent Signatures
```

scenarios.md
```diff
@@ -18,9 +18,31 @@ The [OCI TOB][oci-tob] has adopted [OCI Artifacts][artifacts-repo], generalizing
...
+1. Usable
+   * Simple commands to integrate with toolchains, supporting key hierarchies
+   * Signing an artifact must not change the tag or digest enabling existing dev through production workflows
+   * Key acquisition must support users from hobbyists, open source projects to large software vendors
+
 ## Non-Goals
 
-- Notary v2 does not account for what the content represents or its lineage. Other efforts may attach additional content, and re-sign the super set of content to account for other scenarios. 
+1. Notary v2 does not account for what the content represents or its lineage. Other efforts may attach additional content, and re-sign the super set of content to account for other scenarios.
+2. Backwards compatibiliyt with Notay v1
```
trishankatdatadog, line 45:
```suggestion
2. Backwards compatibility with Notary v1
```

scenarios.md
```diff
@@ -62,15 +62,15 @@ In a world of consuming public software, we must account for content that's acqu
```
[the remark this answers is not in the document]
SteveLasker: I've updated the e2e scenario to be #0, to avoid references to the others by number. 
Please reactivate if I didn't account for the great feedback.

scenarios.md
```diff
@@ -101,17 +101,17 @@ Once the developer has locally validated the build, sign, validate scenario, the
 
 1. Locally build and sign an artifact, such as a `net-monitor:abc123` container image
 1. Rename the artifact to reflect the registry it will be pushed to:  
-  `$ docker tag net-monitor:abc123 wabbitnetworks.example.com/networking/net-monitor:1.0`  
-  `$ docker push wabbitnetworks.example.com/networking/net-monitor:1.0`
+  `$ tool tag net-monitor:abc123 wabbitnetworks.example.com/networking/net-monitor:1.0`  
+  `$ tool push wabbitnetworks.example.com/networking/net-monitor:1.0`
 1. Deploy the artifact to a cluster that requires signatures:  
   `$ orchestrator apply -f deploy.yaml`
 1. The orchestrator in the dev environment accepts any signed content, enabling it to trace where deployed artifacts originated from.
```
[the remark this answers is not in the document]
SteveLasker: In Scenario #1, I'm using a local docker/moby instance. It could also be **kind**, or other local hosts. Where the validation is out of scope. We could imagine the `docker run` command could support a notary v2 policy. The main aspect of # is no registry is required to create and validate a signature. All the content (image and signature) were created locally.
In this scenario, I'm adding the registry to the scenario. 

But, yes, implementing Notary v2 in all these tools is aspirational. If we get it right, we hope they would. If they don't, we must ask _why_?

In the top picture, you'll see the blue symbols are things that are considered in-scope.

scenarios.md
```diff
@@ -166,7 +166,10 @@ ACME Rockets runs secure production environments, limiting all external network
...
 
-- In this scenario, the wabbit-networks signature must be validated within the ACME Rockets network. How this is done is open for design. However, the requirement states the signature must be validated without external access. When the artifact is copied to the private/network restricted registry, the signature may need to be copied, and is assumed to be trusted if available in the trusted server within the private network. How ACME Rockets would copy/proxy the signatures is part of the design and UX for a secure, but usable pattern.
+- In this scenario, the wabbit-networks signature must be validated within the ACME Rockets network. How this is done is open for design. However, the requirement states the signature must be validated without external access.
+- When the artifact is copied to the private/network restricted registry, the signature may need to be copied, and is assumed to be trusted if available in the trusted server within the private network. 
+- How ACME Rockets would copy/proxy the signatures is part of the design and UX for a secure, but usable pattern.
+- How ACME Rockets handles revoked keys is also part of the design phase.
```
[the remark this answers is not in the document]
SteveLasker: I'd like to punt the first bullet to the key managmenet working group. 
We will support key revocation, as part of the key management working group. 
Rollback protection is not yet a key requirement and something we're still discussing as part of phase 2. It's not that we don't think it provides more security. We're just not sure it meets the security/usability/practical use case for registries - yet.

scenarios.md
SteveLasker: I'm leaving this open to see if we cover this in the key management working group.

scenarios.md
SteveLasker: Thanks. I found another spell checking error. I would have used the [commit suggestion] button, but I was afraid it wouldn't `-s` the commit and I'd be in DCO hell :(

PR #16, opened 2020-03-05
Mirror compromise and mitm scenarios

mnm678:
I added a couple of scenarios to reflect the attacker capabilities from the threat model.

### DIFF
--- a/scenarios.md
+++ b/scenarios.md
@@ -215,7 +215,7 @@ An attacker is able to sniff network traffic between the repository and client a
 
 **Implications of this requirement:**
 
-- All downloaded files and metadata files must be validated, even if they are downloaded from a known repository.
+- All downloaded files and metadata files should be checked for integrity, even if they are downloaded from a known repository.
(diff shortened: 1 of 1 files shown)


scenarios.md
```diff
@@ -199,6 +199,26 @@ A deployment requires a mydb image. The mydb image is routinely updated for secu
...
+- Clients should not trust files that are signed only by the mirror, and not by the original repository.
+- Mirrors should not have permission to edit packages and signatures.
+
+### Scenario 8: An attacker performs a MitM attack
+
+An attacker is able to sniff network traffic between the repository and client and alter packets as they are passed through the network.
+
+**Implications of this requirement:**
+
+- All downloaded files and metadata files must be validated, even if they are downloaded from a known repository.
```
SteveLasker: What does "validated" imply here? SSL transmission with current TLS support. Or, are you suggesting some sort of reproducible build like scope?

mnm678: This could mean TLS, signature verification, or any other technique that does not automatically trust files. I can work on a more clear way to word that

PR #17, opened 2020-03-06
Update readme to fix def/terms URL

omieomye:


omieomye:
Can't do in UI coz signatures are needed

PR #22, opened 2020-06-01
add signature verification lookup design

SteveLasker:
Copied from google doc: [Notary v2 - Verification Persistence by Reference](https://docs.google.com/document/d/1VvT6IsnRFuWlhewbL4iY2h6PhLeA57aOoz_mejwRzzE/)


verification-by-reference.md
```diff
@@ -0,0 +1,331 @@
...
+      "digest": "sha256:3c3a4604a545cdc127456d94e421cd355bca5b528f4a9c1"
+    }
+  ]
+}
+```
+
+### web:a2b2 staging verification w/ index
+
+An index is pushed that supports `index.config`, which tells the registry this index is a type of Notary Verification. This follows the same pattern we use with [OCI Artifacts to identify oci manifest as specific type](https://github.com/opencontainers/artifacts/blob/master/artifact-authors.md#defining-oci-artifact-types).
+This index would NOT be pushed with a tag. It would be a digest only reference.
```
josephschorr, line 205: Under Quay's current GC model, this would cause the new index to (eventually) be GCed, since it is unreferenced. We'd either have to keep around "reverse" references or whitelist these artifact types... I wonder if there is any upside to having these be tagged as well. If not, we'll just make the necessary changes on our end.

verification-by-reference.md
```diff
@@ -0,0 +1,331 @@
...
+1. Minimize requirements to change persistence and object stores used by registry operators. Similar to garbage collection, we should work within the constraints of manifest, index and layers to represent verification objects.
+
+## Non-goals
+
+1. Work within the constraints of the existing OCI-distribution spec api set. 
+Supporting Notary v1, or Docker Content Trust, requires new APIs. APIs are comparatively cheap to implement atop the persistence stores of registries. 
+New APIs should be part of the Notary v2 spec, which may represent changes to the OCI-distribution spec or implementations of a new extension model over the Distribution spec.
+1. Compatibility with Notary v1. Registries that have implemented Notary v1 are looking for a better solution. We cumulatively have little existing usage, and if successful, we expect all customers would rapidly move to Notary v2 eliminating the need to maintain two sets of APIs for an extended period of time.
```
directionless, line 28: I love that this is explicitly called out, but as in implementor (and user) of notary-v1, I don't see v2 as solving any of my issues. To wit, v1 has very rough tooling, and is complex. v2, adds significant complexity.

mtrmac:
BTW the traditional TUF, where the signature metadata is stored _completely separately_ using a separate storage API, is also not inconsistent with the requirement that tags/digests should not change on signing.

verification-by-reference.md
SteveLasker: thanks @directionless, I don't think we've gotten far enough along with v2 to say we've added complexity. Our primary goal is to balance complexity with usability. I'd actually say I'm confident we'll tame the complexity beast, as if we don't we won't ship. As we've seen, if it's too complex to use, customer wont. We need a usable solution.

SteveLasker:
> BTW the traditional TUF, where the signature metadata is stored _completely separately_ using a separate storage API, is also not inconsistent with the requirement that tags/digests should not change on signing.

While true, this is one of the issues with registries and docker content trust today. If I pull `demo42.azurecr.io/foo:a1b2c3` with docker content trust enabled, it's completely different content from pulling that same artifact reference with docker content trust disabled. 
This is one of the unifications we'd like to fix. Although, I'm realizing it's not called out.

I'm also not sure how we support multiple signatures with the previous model.

mtrmac:
> While true, this is one of the issues with registries and docker content trust today. If I pull `demo42.azurecr.io/foo:a1b2c3` with docker content trust enabled, it's completely different content from pulling that same artifact reference with docker content trust disabled.

This would be almost equally an issue with this proposal: A user could download foo@digest from one registry and get the signatures (because they have been uploaded there), or foo@digest from another registry and not get the signatures (because haven’t been uploaded there).

(“Almost” because in Notary V1, TUF completely replaces tag lookup, but that’s not inherent to the design; TUF could easily have been used for “tag verification” only, with pulls failing in the intermediate time when only one of (tag, TUF metadata) is updated, or maybe by combining the registry and Notary to update both atomically but still separately.)

SteveLasker:
> This would be almost equally an issue with this proposal: A user could download foo@digest from one registry and get the signatures (because they have been uploaded there), or foo@digest from another registry and not get the signatures (because haven’t been uploaded there).

Artifact movement between registries is covered in [scenario # 5](https://github.com/notaryproject/requirements/blob/master/scenarios.md#scenario-5-validate-artifact-signatures-within-restricted-networks)
It's also covered in the [end to end scenario](https://github.com/notaryproject/requirements/blob/master/scenarios.md#end-to-end-orchestrator-scenario)

If you pull content from one registry, and push it to another, the verification objects should go with it. We also need to account for air-gapped (network constrained) environments as well. 

This will take some time to integrate into the clients. If a user pulls an artifact from one registry, that doesn't have the notary v2 client code, when it's pushed to another registry, only the artifact would be pushed. We'll need to make these artifact clients notary aware so the verification objects go with it.  If we're successful and it's usable, I suggest this would be easy to adopt. If not, then we've failed.

By having all the registry operators engaged, we can build something that we all implement, consistently.

dlorenc:
> Overall, I do think “signing things should not change the digest of the thing being signed” is an important requirement.

As an interested lurker on most of this discussion, I agree with this requirement as well. It makes most of the flows I've thought through easier to understand and use.

verification-by-reference.md
```diff
@@ -0,0 +1,331 @@
...
+
+The pros of this approach are:
+
+- Multiple signatures may be added without changing the original artifact digest as no content changes are made to the digest, or even annotations
+- All deployment references remain: `web:a2b2` and it’s associated digest do not change.
+
+The cons of this approach:
+
+- A new API must be added. We’ve said this was within scope and ok as registries must add Notary v1, Content Trust to support existing scenarios. Since most registries need to do some work, adding a new API is not a stretch
+- Reverse lookup, and its impact on garbage collection. Registries tend to traverse down. If multiple manifests reference common base layers, the registry knows to reference count shared layers and only deletes the layers once all manifests are removed. Orphaned manifests may be deleted if they’re no longer referenced by tags.
```
jonjohnsonjr, line 307: > Orphaned manifests may be deleted if they’re no longer referenced by tags. 

I don't really agree with this, in general. Images are addressable by digest, so it's not always safe to just GC orphaned manifests. It's a reasonable policy to enable, but this would be a dangerous default.

verification-by-reference.md
```diff
@@ -0,0 +1,331 @@
...
+- All deployment references remain: `web:a2b2` and it’s associated digest do not change.
+
+The cons of this approach:
+
+- A new API must be added. We’ve said this was within scope and ok as registries must add Notary v1, Content Trust to support existing scenarios. Since most registries need to do some work, adding a new API is not a stretch
+- Reverse lookup, and its impact on garbage collection. Registries tend to traverse down. If multiple manifests reference common base layers, the registry knows to reference count shared layers and only deletes the layers once all manifests are removed. Orphaned manifests may be deleted if they’re no longer referenced by tags.  
+Most importantly, a manifest is likely blocked from deletion if it’s referenced by an index. In this reverse lookup model, the user likely wants all verifications to be deleted when the artifact is deleted. This isn’t necessarily an immediate blocker, but it does go against the garbage collection constraint outlined above.
+This is work registry operators must implement so users don’t have to change their artifact references? Is this the right trade-off?
```
jonjohnsonjr, line 309: I really struggle with this.

There are tons of registry implementations out there, many of which will never be updated. Any API that we add will cause some friction, and I'd be somewhat worried about incompatible implementations of that API.

If we can do everything client side without updating registries, we can more easily get widespread adoption without being blocked by registries. It would be nice if there was some fallback pattern for registries that don't implement a reverse lookup.

verification-by-reference.md
SteveLasker: Deleting untagged manifests is a choice of a registry operator and a customer. We'll have more info on this as we've had a few discussion if cross references should be tracked by manifests and indexes? 
Or, is there an additional way to track references?
- Reading a config object?
- A separate API to link a signature to a manifest? Having a separate API would enable a different role for SIGNER. But, we could also say the role should be applied to a given Artifact type: `vnd.cncf.notary.config.v2` for instance.

I think we agree on the constraints of the pros/cons and need to figure out a tenable solution.

verification-by-reference.md
SteveLasker: Not having an impact on a registry to support Artifacts was the goal, as we knew it didn't hit the "high bar" for requiring a major change. We actually just require a registry operator to relax the mediaType constraints, which is consistent with the spec.
For Signing, we think this is much needed feature. Registries that implemented Notary v1/Docker Content Trust had a LOT of work to do. For notary v2, we're hoping this is limited to one or two APIs.

The point being, the desired results are worth the added investment. If a registry operator doesn't want to support signing, that's their choice. I suspect all registries will want/need to support signing, so we just need to identify the benefits for security and usability.

PR #24, opened 2020-06-22
Create keymanagementscenarios.md

NiazFK:


NiazFK:
Creating a doc to track key management scenarios. Starting off with questions that need to get answered through use cases.

SteveLasker:
Hi Niaz, 
These are a great set of questions for us to consider. They look more like requirements and/or questions than scenarios. 
Can you possibly merge with a statement that supports what we should do, from a scenario perspective? Here's some ideas, with your expertise for key management added to the __'s

- Keys must be stored in ___ "your expertise here".  Some reference to being an extensible model to support different key management solutions
- Keys must be at least ___ (size), supporting migration to future/more secure keys through ___
- Key revocation must be supported through ___ industry standards (whatever the key managmenet experts like Niaz and others say it should be). Including how air-gapped environments should be supported
- timestamping ___ for __ must be supported through ___

For minimum number of keys, isn't this s a policy type thing? The registry would support multiple keys, and I imagine multiple keys are added as the artifact makes it's way through a workflow.

PR #27, opened 2020-07-10
Draft of use cases for key management scenarios

NiazFK:
This draft lists out scenarios to be covered for key management.

Future revisions will detail out steps for each scenario.

### DIFF
--- a/keymanagementscenarios.md
+++ b/keymanagementscenarios.md
@@ -1,3 +1,35 @@
+Key management for container signing can be broadly categorized into three general use cases:
+- Signing (How are keys made available when Notary v2 signatures are generated?)
+- Trust store configuration (How do runtime environments determine which keys to trust?)
+- Key setup/revocation (How do key owners create/revoke keys? How do runtime environments get this information?)
+
+Signing Use Cases:
+- Developer runs docker build and notary v2 sign on local machine. Keys are stored in key store or plain text on local machine (Need to identify keystores we want to support. Should we preclude signing with plain text keys to improve security posture?).
+- Developer runs docker build and notary v2 sign on local machine. Keys are stored in a USB Token.
+- Developer runs docker build and notary v2 sign on local machine. Keys are stored in a network HSM.
+- Developer runs docker build and notary v2 sign on local machine. Keys are stored in a Cloud Key Management Service.
+- Automated build system runs docker build and notary v2 sign on build machine.
+- Automated build system runs docker build on a build machine. Signing is done on a second machine that verifies the container meets developer's release criteria.
+
+Trust Store Configuration:
+- Developer adds/removes their root public keys to the trust store used on a runtime host. The trust store will validate images pulled from any registry.
+- Developer adds/removes root public keys for third parties they trust to their trust stores. The trust store will validate images pulled from any registry.
+- Enterprises can restrict users to add/remove their root public keys to the trust store used across a fleet of runtime hosts. The trust store will validate images pulled from any registry.
+- Enterprises can restrict users to add/remove root public keys for trusted third parties to their trust stores. The trust store will validate images pulled from any registry.
+- [TBD] Do we envision a scenario where the trust store will need to define subordinate keys in addition to the root?
+
+Key setup/revocation use cases:
+- Developer sets up a single root key and delegates additional keys from it.
+- Developer revokes/rotates chained keys. The revocation/rotation information is automatically available to run time environments.
+- Developer revokes root keys. The run time environment administrator needs to update their trust store.
+- Run time environments can automatically pull revocation lists for keys in their trust store.
+- Air gapped run time environments can periodically pull and cache revocation lists for keys in their trust store.
+
+
+In addition we also need to consider the following for cryptographic security:
+- Supported key types
+- Supported signing algorithms
+
 Identify use cases to clarify:
 1. Where can keys be stored? What interfaces need to be supported?
 2. Any limitations ok key types/sizes supported?
--- a/keymanagementscenarios.md
+++ b/keymanagementscenarios.md
@@ -6,7 +6,7 @@ Key management for container signing can be broadly categorized into three gener
 Personas:
 - Publisher: User who builds and signs containers.
     - Publisher Admin: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Admin users (i.e. security administrator) will be responsible for configuring roots, provide access to use or generate delegate keys, and make decisions for key revocation.
-    - Publisher Builder [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Builders (i.e. developers, build hosts) will be responsible for building contatiner images. 
+    - Publisher Builder [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Builders (i.e. developers, build hosts) will be responsible for building container images. 
     - Publisher Singer [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Signers (i.e. developers, signing hosts) will be responsible for signing container images.
 - Deployer: User who deploys containers.
 
--- a/keymanagementscenarios.md
+++ b/keymanagementscenarios.md
@@ -1,0 +1,8 @@
+Identify use cases to clarify:
+1. Where can keys be stored? What interfaces need to be supported?
+2. Any limitations ok key types/sizes supported?
+3. How will public keys (root of trust) be distributed?
+4. How will key revocation information be distributed?
+5a. What is the minimum number of keys needed to succesfully sign a container?
+5b. What is the recommended number of keys to sign a container?
+6. Any additional requirements for timestamping?
--- a/keymanagementscenarios.md
+++ b/keymanagementscenarios.md
@@ -1,30 +1,87 @@
 Key management for container signing can be broadly categorized into three general use cases:
-- Signing (How are keys made available when Notary v2 signatures are generated?)
+- Key Setup/Signing (How are keys set up? How are keys accessed by Notary v2 client when signatures are generated?)
 - Trust store configuration (How do runtime environments determine which keys to trust?)
-- Key setup/revocation (How do key owners create/revoke keys? How do runtime environments get this information?)
+- Revocation (How do key owners create/revoke keys? How do runtime environments get this information?)
+
+Personas:
+- Publisher: User who builds containers.
+- Deployer: User who deploys containers.
 
 Signing Use Cases:
-- Developer runs docker build and notary v2 sign on local machine. Keys are stored in key store or plain text on local machine (Need to identify keystores we want to support. Should we preclude signing with plain text keys to improve security posture?).
-- Developer runs docker build and notary v2 sign on local machine. Keys are stored in a USB Token.
-- Developer runs docker build and notary v2 sign on local machine. Keys are stored in a network HSM.
-- Developer runs docker build and notary v2 sign on local machine. Keys are stored in a Cloud Key Management Service.
-- Automated build system runs docker build and notary v2 sign on build machine.
-- Automated build system runs docker build on a build machine. Signing is done on a second machine that verifies the container meets developer's release criteria.
+- Local Key Store
+    - Publisher creates new root key on local machine. Keys are stored in key store or plain text on local machine (Need to identify keystores we want to support. Should we preclude signing with plain text keys to improve security posture?).
+    - Publisher uploads root public key to registry (registry can verify user for container uploads). 
+    - Publisher shares root public key (other users can verify containers before running).
+    - Publisher creates new delegate keys on local machine that chain to root key.
+    - Publisher runs docker build and notary v2 sign on local machine. Notary v2 client generates signatures with key in memory.
+- USB Token Key Store
+    - Publisher creates new root key on USB token. 
+    - Publisher uploads root public key to registry (registry can verify user for container uploads). 
+    - Publisher shares root public key (other users can verify containers before running).
+    - Publisher creates new delegate keys on USB token that chain to root key.
+    - Publisher runs docker build and notary v2 sign on local machine. Notary v2 client generates signatures with key in USB Token.
+- Network HSM
+    - Publisher creates new root key on Network HSM. 
+    - Publisher uploads root public key to registry (registry can verify user for container uploads). 
+    - Publisher shares root public key (other users can verify containers before running).
+    - Publisher creates new delegate keys on Network HSM that chain to root key.
+    - Publisher runs docker build and notary v2 sign on local machine. Notary v2 client generates signatures with key in Network HSM.
+- Cloud Key Management Service
+    - Publisher creates new root key on Cloud Key Management Service 
+    - Publisher uploads root public key to registry (registry can verify user for container uploads). 
+    - Publisher shares root public key (other users can verify containers before running).
+    - Publisher creates new delegate keys on Cloud Key Management Service that chain to root key.
+    - Publisher runs docker build and notary v2 sign on local machine. Notary v2 client generates signatures with key in Cloud Key Management Service.
+- Hybrid
+    - Publisher creates new root key on USB Token/Network HSM/Cloud Key Management Service 
+    - Publisher uploads root public key to registry (registry can verify user for container uploads). 
+    - Publisher shares root public key (other users can verify containers before running).
+    - Publisher creates new delegate keys on local machine that chain to root key.
+    - Publisher runs docker build and notary v2 sign on local machine. Notary v2 client generates signatures with keys on local machine.
+- Signing by Build Machines
+    - Publisher creates new root, uploads root public key to registry, and shares root public key.
+    - Publisher creates new delegate keys on build machine or on USB Token/Network HSM/Cloud Key Management Service
+    - Publisher configures build scripts with location of delegate keys.
+    - Build machine runs docker build and then notary v2 sign with keys in configured location.
+- Signing by Dedicated Machines
+    - Publisher creates new root, uploads root public key to registry, and shares root public key.
+    - Publisher creates new delegate keys on build machine or on USB Token/Network HSM/Cloud Key Management Service
+    - Publisher configures signing machine with location of delegate keys.
+    - Build machine runs docker build and passes objects to be signed to signing host.
+    - Signing host generates signatures with keys in configured location.
 
 Trust Store Configuration:
-- Developer adds/removes their root public keys to the trust store used on a runtime host. The trust store will validate images pulled from any registry.
-- Developer adds/removes root public keys for third parties they trust to their trust stores. The trust store will validate images pulled from any registry.
+- Specify trusted public root keys
+    - Deployer gets root public key from publisher.
+    - Deployer adds root public key to runtime configuration.
+    - Container pulled from any registry is be validated with listed root public keys before execution.
+- Specify location of trusted public root keys
+    - [TODO]
 - Enterprises can restrict users to add/remove their root public keys to the trust store used across a fleet of runtime hosts. The trust store will validate images pulled from any registry.
 - Enterprises can restrict users to add/remove root public keys for trusted third parties to their trust stores. The trust store will validate images pulled from any registry.
 - [TBD] Do we envision a scenario where the trust store will need to define subordinate keys in addition to the root?
+- Air gapped environments
 
-Key setup/revocation use cases:
-- Developer sets up a single root key and delegates additional keys from it.
-- Developer revokes/rotates chained keys. The revocation/rotation information is automatically available to run time environments.
-- Developer revokes root keys. The run time environment administrator needs to update their trust store.
-- Run time environments can automatically pull revocation lists for keys in their trust store.
-- Air gapped run time environments can periodically pull and cache revocation lists for keys in their trust store.
-
+Key rotation/revocation use cases:
+- Root Revocation
+    - Publisher deletes revoked root public key on registry (registry can stop sharing containers with revoked key).
+    - Registry stops vending containers signed with old root key.
+    - Publisher removes revoked root public key from shared list (other users stop trusting artifacts with revoked key).
+    - Publisher creates new root key.
+    - Publisher uploads new root public key to registry (registry can verify user for container uploads). 
+    - Publisher updates shared root public key (other users can verify containers with new key before running).
+    - Publisher creates new delegate keys.
+- Delegate Key Revocation
+    - Publisher lists revoked delegate keys as untrusted. Revocation list is signed by root key.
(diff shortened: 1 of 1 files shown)


keymanagementscenarios.md
```diff
@@ -0,0 +1,40 @@
...
+- [TBD] Do we envision a scenario where the trust store will need to define subordinate keys in addition to the root?
+
+Key setup/revocation use cases:
+- Developer sets up a single root key and delegates additional keys from it.
+- Developer revokes/rotates chained keys. The revocation/rotation information is automatically available to run time environments.
+- Developer revokes root keys. The run time environment administrator needs to update their trust store.
+- Run time environments can automatically pull revocation lists for keys in their trust store.
+- Air gapped run time environments can periodically pull and cache revocation lists for keys in their trust store.
+
+
```
SteveLasker, line 28: Suggest adding a scenario for moving keys between registries 
- public to private
- source to air-gap, discussing the one-way model for air-gap environments

keymanagementscenarios.md
```diff
@@ -0,0 +1,40 @@
...
+In addition we also need to consider the following for cryptographic security:
+- Supported key types
+- Supported signing algorithms
+
+Identify use cases to clarify:
+1. Where can keys be stored? What interfaces need to be supported?
+2. Any limitations ok key types/sizes supported?
```
SteveLasker, line 35: Should this be a customer configuration - perhaps the registry states the types of keys supported/required

keymanagementscenarios.md
```diff
@@ -0,0 +1,40 @@
...
+- Developer runs docker build and notary v2 sign on local machine. Keys are stored in a Cloud Key Management Service.
+- Automated build system runs docker build and notary v2 sign on build machine.
+- Automated build system runs docker build on a build machine. Signing is done on a second machine that verifies the container meets developer's release criteria.
+
+Trust Store Configuration:
+- Developer adds/removes their root public keys to the trust store used on a runtime host. The trust store will validate images pulled from any registry.
+- Developer adds/removes root public keys for third parties they trust to their trust stores. The trust store will validate images pulled from any registry.
+- Enterprises can restrict users to add/remove their root public keys to the trust store used across a fleet of runtime hosts. The trust store will validate images pulled from any registry.
+- Enterprises can restrict users to add/remove root public keys for trusted third parties to their trust stores. The trust store will validate images pulled from any registry.
+- [TBD] Do we envision a scenario where the trust store will need to define subordinate keys in addition to the root?
```
SteveLasker, line 19: A requirement of Notary v2 is to provide a pluggable model where the customer can plug-in their preferred key managmenet solution. For Notary v2, we'll use a CNCF project. Each cloud vendor can implement their keyvault, providing feedback to the Notary v2 working group.

keymanagementscenarios.md
```diff
@@ -0,0 +1,40 @@
...
+Signing Use Cases:
+- Developer runs docker build and notary v2 sign on local machine. Keys are stored in key store or plain text on local machine (Need to identify keystores we want to support. Should we preclude signing with plain text keys to improve security posture?).
+- Developer runs docker build and notary v2 sign on local machine. Keys are stored in a USB Token.
+- Developer runs docker build and notary v2 sign on local machine. Keys are stored in a network HSM.
+- Developer runs docker build and notary v2 sign on local machine. Keys are stored in a Cloud Key Management Service.
+- Automated build system runs docker build and notary v2 sign on build machine.
+- Automated build system runs docker build on a build machine. Signing is done on a second machine that verifies the container meets developer's release criteria.
+
+Trust Store Configuration:
+- Developer adds/removes their root public keys to the trust store used on a runtime host. The trust store will validate images pulled from any registry.
```
mtrmac, line 15: What does “from any registry” mean here?
- Just that every image must be signed?
- Or that any “root” key can sign any image?

Neither is really attractive:
- It should be possible — it might even be the default! — for users to opt out and accept unsigned images from e.g. Docker Hub, while enforcing signatures from known vendors and their registries/repositories
- It should never be possible for one vendor to sign images that were supposed to come from some other vendor.

Or does it mean something else entirely?

NiazFK: This applies to the runtime environment. It means that if the trusted root keys are defined, images from any registry can be verified as being signed with delegates belonging to the roots. I'll clarify the language in the next revision.

keymanagementscenarios.md
NiazFK: Commit 54891ae addresses this for delegating keys for signing.

keymanagementscenarios.md
NiazFK: The client should limit choices to key types/sizes that are agreed upon as secure. Users can configure from the options for what best meets their needs. Any reason registries can't support all the of the limited choices supported by Notary v2?

keymanagementscenarios.md
NiazFK: Commit 54891ae addresses this.

keymanagementscenarios.md
```diff
@@ -0,0 +1,105 @@
...
+- Key Setup/Signing (How are keys set up? How are keys accessed by Notary v2 client when signatures are generated?)
+- Trust store configuration (How do runtime environments determine which keys to trust?)
+- Revocation (How do key owners create/revoke keys? How do runtime environments get this information?)
+
+Personas:
+- Publisher: User who builds and signs containers.
+    - Publisher Admin: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Admin users (i.e. security administrator) will be responsible for configuring roots, provide access to use or generate delegate keys, and make decisions for key revocation.
+    - Publisher Builder [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Builders (i.e. developers, build hosts) will be responsible for building contatiner images. 
+    - Publisher Singer [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Signers (i.e. developers, signing hosts) will be responsible for signing container images.
```
trishankatdatadog, line 10:
```suggestion
    - Publisher Signer [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Signers (i.e. developers, signing hosts) will be responsible for signing container images.
```

keymanagementscenarios.md
```diff
@@ -0,0 +1,105 @@
+Key management for container signing can be broadly categorized into three general use cases:
+- Key Setup/Signing (How are keys set up? How are keys accessed by Notary v2 client when signatures are generated?)
+- Trust store configuration (How do runtime environments determine which keys to trust?)
+- Revocation (How do key owners create/revoke keys? How do runtime environments get this information?)
+
+Personas:
+- Publisher: User who builds and signs containers.
+    - Publisher Admin: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Admin users (i.e. security administrator) will be responsible for configuring roots, provide access to use or generate delegate keys, and make decisions for key revocation.
+    - Publisher Builder [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Builders (i.e. developers, build hosts) will be responsible for building contatiner images.
```
trishankatdatadog, line 9:
```suggestion
    - Publisher Builder [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Builders (i.e. developers, build hosts) will be responsible for building container images. 
```

keymanagementscenarios.md
```diff
@@ -0,0 +1,105 @@
...
+- Revocation (How do key owners create/revoke keys? How do runtime environments get this information?)
+
+Personas:
+- Publisher: User who builds and signs containers.
+    - Publisher Admin: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Admin users (i.e. security administrator) will be responsible for configuring roots, provide access to use or generate delegate keys, and make decisions for key revocation.
+    - Publisher Builder [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Builders (i.e. developers, build hosts) will be responsible for building contatiner images. 
+    - Publisher Singer [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Signers (i.e. developers, signing hosts) will be responsible for signing container images.
+- Deployer: User who deploys containers.
+
+Signing use cases:
```
trishankatdatadog, line 13: Great coverage of possible use cases here

keymanagementscenarios.md
```diff
@@ -0,0 +1,105 @@
...
+- Supported signing algorithms
+
+Identify use cases to clarify:
+1. Where can keys be stored? What interfaces need to be supported?
+2. Any limitations ok key types/sizes supported?
+3. How will public keys (root of trust) be distributed?
+4. How will key revocation information be distributed?
+5a. What is the minimum number of keys needed to succesfully sign a container?
+5b. What is the recommended number of keys to sign a container?
+6. Any additional requirements for timestamping?
```
trishankatdatadog, line 105: By "timestamping," do you mean [this](https://knowledge.digicert.com/generalinformation/INFO1119#:~:text=Timestamping%20ensures%20that%20code%20will,a%20timestamp%20for%20your%20code.)?

keymanagementscenarios.md
```diff
@@ -0,0 +1,105 @@
...
+    - Publisher updates shared root public key (other users can verify containers with new key before running).
+    - Publisher creates new delegate keys from new root following any scenario listed earlier in Signing use cases.
+- Delegate Key Rotation
+    - Publisher creates new delegate keys from existing root following any scenario listed earlier in Signing use cases.
+
+In addition we also need to consider the following for cryptographic security:
+- Supported key types
+- Supported signing algorithms
+
+Identify use cases to clarify:
```
trishankatdatadog, line 98: What about cryptographic algorithms? Do we wanted versioned protocols, or cryptographic agility?

keymanagementscenarios.md
SteveLasker: I can see this as a policy that registry owners (not the cloud operator, but the security team that is responsible for the content in the registry) would want to impose a policy for the types of keys they consider secure. The client would also enforce this.

keymanagementscenarios.md
SteveLasker: When an artifact is signed, can we get a song to play? Ding for good, dong for error.

SteveLasker:
Some nits on images vs. artifacts, but overall looks good for base iteration.

keymanagementscenarios.md
```diff
@@ -0,0 +1,105 @@
...
+- Key Setup/Signing (How are keys set up? How are keys accessed by Notary v2 client when signatures are generated?)
+- Trust store configuration (How do runtime environments determine which keys to trust?)
+- Revocation (How do key owners create/revoke keys? How do runtime environments get this information?)
+
+Personas:
+- Publisher: User who builds and signs containers.
+    - Publisher Admin: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Admin users (i.e. security administrator) will be responsible for configuring roots, provide access to use or generate delegate keys, and make decisions for key revocation.
+    - Publisher Builder [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Builders (i.e. developers, build hosts) will be responsible for building container images. 
+    - Publisher Signer [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Signers (i.e. developers, signing hosts) will be responsible for signing container images.
```
marcofranssen, line 10: Here I would also include things like (CI jobs).

keymanagementscenarios.md
```diff
@@ -0,0 +1,105 @@
...
+    - Container pulled from any registry is be validated with listed root public keys before execution.
+- Specify location of trusted public root keys
+    - [TODO]
+- Enterprises can restrict users to add/remove their root public keys to the trust store used across a fleet of runtime hosts. The trust store will validate images pulled from any registry.
+- Enterprises can restrict users to add/remove root public keys for trusted third parties to their trust stores. The trust store will validate images pulled from any registry.
+- Trust store also includes subordinate keys
+- Air gapped environments
+
+Key rotation/revocation use cases:
+- Root Revocation (compromised root should not be needed in process to designate itself as revoked, otherwise attacker can use compromised root for a key rotation locking out publisher)
```
mnm678, line 74: This is an important case to consider. If an attacker has compromised a root signing key, the system needs a way to recover. However, I think it would be hard to rely on an outside channel to communicate a new key to all clients, because then you need a way for the client to trust this outside channel, and revoke that trust if needed. If instead the root role itself had multiple required keys, each stored offline, the system could use other root signing keys to revoke the compromised key using the regular channels. This would ensure that no clients miss the revocation announcement.

mtrmac: Very few users are going to maintain three or more offline keys, stored in _separate_ ways that don’t all share a single point of failure. Even a single off-line key is a stretch for many. N keys means N times bigger chance to lose the ability to sign.

In practice, there _always_ is an “outside channel”: If a vendor publishes “we’ve been compromised, replace your keys (our your entire signing implementation) with $this” on a web site, maybe e-mails or calls important clients, or tells the media, the clients are eventually (OTOH, after a considerable delay) going to trust the people over the cryptography.

(Also, at least anecdotally, high-value keys are comparatively rarely compromised as in “private key material is disclosed” (though it _does_ happen); hardware security modules decrease such risk. More common seems to be for the HSM/key holder to be instructed to sign something it shouldn’t, e.g. by compromising a build system or a user account.)

---

(Personally, I’m very willing to give up on automated key revocation / compromise recovery, if the result is a much less complex system. A signing system that automatically recovers itself is perfect/necessary for self-updating closed systems (single self-updating applications, maybe cars) where users can’t be involved, but for signing generic container images, where there is a dedicated administrator and the container images intentionally don’t have control over their software environment, it’s, in some sense, almost an anti-feature, because the signing system client has no idea what the container images do and how to recover anything but the signing system.

The administrator _must be told_ about the compromise, and must inventory the signed images / possibly compromised data, investigate how the compromise has spread … and maybe recover from backups or consult network connection audit logs, and things like that (which are intentionally not accessible to the ordinary containers, so a typical application can’t self-update itself to a “fix everything” version). Replacing a public key on all nodes is the simplest and most easily automated part of all this, using an existing configuration management system, so the cost and complexity of off-line keys and delegations and rollback protection for the key delegation and delivering state to ephemeral cloud cluster members and … is not overall that attractive in my judgement (though opinions on this certainly differ).)

trishankatdatadog: > Very few users are going to maintain three or more offline keys, stored in _separate_ ways that don’t all share a single point of failure. Even a single off-line key is a stretch for many. N keys means N times bigger chance to lose the ability to sign.

With things like Hashicorp Vault and Azure Key Vault/AWS KMS/Google Cloud KMS, managing multiple keys shouldn't be too much of a challenge these days.

mtrmac: It’s not the _number_ of keys, it’s the attack scenarios. If, by assumption, a root key has already been compromised, for the recovery using the other keys to work, the other keys must have been stored in a way that is unlikely to have been compromised the same way. Storing all the keys in a single service may not meet that bar.

trishankatdatadog: > It’s not the _number_ of keys, it’s the attack scenarios. If, by assumption, a root key has already been compromised, for the recovery using the other keys to work, the other keys must have been stored in a way that is unlikely to have been compromised the same way. Storing all the keys in a single service may not meet that bar.

This is a fair point. Yes, ideally, each root key should be distributed over different services/hardware/locations/etc. I don't think this is the point so much as being able to rotate the root keys in the first place, even if a threshold has been compromised, because once you regain control of the server, you would like to be able to transparently rotate them for the users who have not yet been compromised. How many root keys and how you manage them should be up to you, even if you only use one.

mnm678: > How many root keys and how you manage them should be up to you, even if you only use one.

I think this is the main point. In some cases it might be true that you'll want to rely on a single key, then have direct discussions with clients when that key is compromised (although I see some potential for a social engineering attack and some of the TOFU issues). But if the system will ever be used in a more security critical application, or one in which you have too many clients to individually handle a key compromise, you want to make a compromise as unlikely as possible and be able to support automated recovery in most cases, and that means having support for multiple root keys.

> Personally, I’m very willing to give up on automated key revocation / compromise recovery, if the result is a much less complex system. 

The only complexity that is added is in one-time setup of root keys and the use of these keys in the event of a compromise. The verification can all be done transparently to users by adding it to cli tools. This is probably less complex than going through an outside channel for any key rotation (even if say there was 1 compromised key out of 4 root keys, and clients require a threshold of 3 signatures to trust the root metadata).

I'd be curious to hear from other potential adopters about their ability to manage offline root keys (or at least multiple root keys stored in different locations) to see if there's any tooling we could add to make this more manageable.

mtrmac:
> The only complexity that is added is in one-time setup of root keys and the use of these keys in the event of a compromise. The verification can all be done transparently to users by adding it to cli tools.

There’s also the implementation complexity. E.g. Notary v1 had a bug where a signed “targets” sub-delegation was actually allowed to override a delegation/trusted key for any other role including root. (Sure, that one bug is specific to a particular implementation approach, and not a reason to object to delegations in particular; it’s just that complexity adds risk attack exposure, and maintenance cost.)

keymanagementscenarios.md
mtrmac: > The only complexity that is added is in one-time setup of root keys and the use of these keys in the event of a compromise. The verification can all be done transparently to users by adding it to cli tools.

There’s also the implementation complexity. E.g. Notary v1 had a bug where a signed “targets” sub-delegation was actually allowed to override a delegation/trusted key for any other role including root. (Sure, that one bug is specific to a particular implementation approach, and not a reason to object to delegations in particular; it’s just that complexity adds risk attack exposure, and maintenance cost.)

trishankatdatadog: > There’s also the implementation complexity. E.g. Notary v1 had a bug where a signed “targets” sub-delegation was actually allowed to override a delegation/trusted key for any other role including root. (Sure, that one bug is specific to a particular implementation approach, and not a reason to object to delegations in particular; it’s just that complexity adds risk attack exposure, and maintenance cost.)

We're all here to fix Notary-v1, which includes broken implementation (not resulting from bad TUF specification, BTW) ...therefore, we should remove flexibility _and_ reduce security for those who want them?

mtrmac: (This has all been a parenthetical comment; I do need we need to resolve this eventually, but in this context the the main point was the impracticality of assuming many root keys help much.

Dropping root key rotation does not remove flexibility; the opposite, it allows using any existing configuration management tools (which necessarily exist, if only to bootstrap trust).

We are here to fix Notary v1, which was, AFAICT, primarily not adopted because it was too complex to use/set up, due to assumptions ~hard-coded both in the implementation and the cryptographic design, and due to the large amount of Notary-specific tooling/infrastructure that was required to use it (and some of that infrastructure didn’t even exist); not, to my knowledge, because it had too few cryptographic protections and users were afraid to use it.

We’ve been dancing around this trade-off for months now, and seem to be no closer to resolving it, sure.)

keymanagementscenarios.md
```diff
@@ -0,0 +1,105 @@
+Key management for container signing can be broadly categorized into three general use cases:
+- Key Setup/Signing (How are keys set up? How are keys accessed by Notary v2 client when signatures are generated?)
+- Trust store configuration (How do runtime environments determine which keys to trust?)
+- Revocation (How do key owners create/revoke keys? How do runtime environments get this information?)
+
```
dbilling, line 5: Should there be a definition section?  specifically I am looking for definitions of (or a pointer to the definition of) the root key, delegate keys, and subordinate keys as used in hthis doc, as well as a description of the key chaining concepts used in this doc.   it's really hard to follow the doc without knowing the meaning of these terms in the notary v2 context and how the chaining capabilities work.  (i.e. how many levels of chaining are supported, etc) These concepts are critical to get right as they are some of the areas where the notary v1 was most lacking.

keymanagementscenarios.md
```diff
@@ -0,0 +1,105 @@
...
+Personas:
+- Publisher: User who builds and signs containers.
+    - Publisher Admin: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Admin users (i.e. security administrator) will be responsible for configuring roots, provide access to use or generate delegate keys, and make decisions for key revocation.
+    - Publisher Builder [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Builders (i.e. developers, build hosts) will be responsible for building container images. 
+    - Publisher Signer [TODO: better wording]: In some scenarios a publisher will include a group of users (i.e. teams or enterprises). Signers (i.e. developers, signing hosts) will be responsible for signing container images.
+- Deployer: User who deploys containers.
+
+Signing use cases:
+- Local Key Store
+    - Publisher creates new root key on local machine. Keys are stored in key store or plain text on local machine (Need to identify keystores we want to support. We can preclude signing with root keys to improve security posture).
```
dbilling, line 15: can this say "creates or reuses a previously created root key"?   Stating creates only seems to imply each registry/repo gets a different root key.      (same comment applies to all cases below....)

keymanagementscenarios.md
```diff
@@ -0,0 +1,105 @@
...
+    - Publisher Admin configures signing machine with credentials and if needed location of delegate keys.
+    - Build machine (Publisher Builder only in this context) runs docker build and passes objects to be signed to signing host (the build machine does not have access to signing keys. This prevents signing keys from being exposed through potential vulnerabilities in build logic).
+    - Signing host generates signatures with keys in configured location.
+- Signing in Air Gapped Environments with Shared Root 
+    - Publisher Admin creates new root, uploads root public key to registry, and shares root public key (similar justification to earlier scenarios).
+    - Publisher Admin creates new subordinate key in air gapped environment on a host, USB Token, Network HSM, or Cloud Key Management Service chaining to root (one time operation enables the root to be used by runtime environments outside of the air gapped environment to validate signatures generated inside the environment).
```
dbilling, line 58: If the concept of subordinate keys exists, and the chaining can occur offline between root and subordinate, I don't understand the need for the "one time operation" concept describe here.  Can you provide more detail?

keymanagementscenarios.md
trishankatdatadog: > We’ve been dancing around this trade-off for months now, and seem to be no closer to resolving it, sure.)

There is no necessary trade-off. Notary-v1 made some implementation choices that made usability harder, which we can fix now while keeping flexibility in key management (e.g., by requiring the registry to manage one root, timestamp, and snapshot for all repos instead of per repo) instead of downright dropping it. I agree that we're no closer to resolving this issue.

keymanagementscenarios.md
mnm678: I think it would be good to define what is meant by a root key here, so we know what this root key is trusted to sign

PR #29, opened 2020-08-07
Move goals and stakeholders to root readme

SteveLasker:


SteveLasker:
looking for 👀 to move the goals up to the root readme, in the streamlined format.
We'll have separate PRs to amend the goals.

PR #32, opened 2020-08-26
Verification by reference

SteveLasker:
Copied from google doc: Notary v2 - Verification Persistence by Reference
Continuation of #22 

Signed-off-by: Steve Lasker stevelasker@hotmail.com


SteveLasker:
Merging as it reflects an intended state, that we want to reference through status updates. As we evolve the reverse lookup design to possibly support a new `artifact.manifest` we can update.

PR #45, opened 2021-02-13
Feb 2021 Notary v2 Status

SteveLasker:
Signed-off-by: Steve Lasker <stevenlasker@hotmail.com>


notary-status.md
```diff
@@ -0,0 +1,183 @@
...
+1. [*Author a Notary v2 Spec*](#author-a-notary-v2-spec-(2021-02))
+
+#### Define Requirements (2021-02)
+
+[Baseline requirements](https://github.com/notaryproject/requirements#requirements) have been defined, enabling focused prototyping. Additional requirements are being discussed as PRs:
+
+- [Add ephemeral clients as a core scenario #25](https://github.com/notaryproject/requirements/issues/25)
+- [Should tags be signed in addition to digests? #43](https://github.com/notaryproject/requirements/issues/43)
+
+#### Build Prototypes (2021-02)
```
dlorenc, line 90: Is this meant to imply we're going to be prototyping for two years?

notary-status.md
```diff
@@ -0,0 +1,183 @@
...
+- Pushing to an instance of the [CNCF Distribution][cncf-distribution] based registry, with exploratory [prototype enhancements](https://github.com/notaryproject/distribution/pull/2).
+- Discovery of signatures, based on a [verification by reference](https://github.com/notaryproject/requirements/blob/main/verification-by-reference.md) design.
+- Validation of a signature with the `nv2` client, using the public key pair used to sign the artifact.
+
+Details of the September 2020 Prototype:
+
+- [Video of prototype-1](https://youtu.be/ciNW19F_T_8?t=367)  
+- [Steps to reproduce the prototype-1 nv2 client](https://github.com/notaryproject/nv2/tree/prototype-1/docs/nv2)
+
+The prototypes have enabled participants and subject matter experts to identify additional areas of focus, such as [policy management capabilities](https://github.com/notaryproject/requirements/issues/42). Policy management will be added as an area of focus.
```
dlorenc, line 106: This feels like a huge expansion in scope IMO.

I think there's probably a difference between prototyping to ensure policy management can coexist with the signature formats/retrieval/etc. and making it a core part of the notary project.

Which one is implied here?

dlorenc:
Thanks for writing this up @SteveLasker! There's one other question I've been trying to wrap my head around here, apologies if this is stated somewhere else and I've missed it - but what exactly _is Notary v2_?

Is it a set of changes to the OCI specification? A signature specification unto itself? A set of tools for working with signatures? All of the above? Something completely different?

As an outsider/casual observer trying to understand where Notary v2 fits/will fit into my workflow of "I just want to sign a container image", I've had a hard time understanding what this will all look like when complete.

notary-status.md
SteveLasker: lol, it was intended to create a unique header/anchor 2012-February
Once I split the overview and specific status reports, I’ll change as we’ll only have a single heading per report.

notary-status.md
SteveLasker: it’s a good check. 
There have been a number of great questions that have come up, such as:
- How many keys should be used to verify an artifact
- Should the CN of the cert match the registry the artifact is currently in?
  - one in the ACME Rockets environment, the wabbit networks key can be used to verify the content, but the registry would match the ACME Rockets CN. 
  - issues around key revocation, such as delays in air-gapped environments

The Policy Management area is a bucket for these topics as we don’t believe they should be core to key validation, rather options in policy management. To your scope question, I don’t expect this would replace OPA/GateKeeper, rather integrate with these type of policy management solutions.

SteveLasker:
Thanks @dlorenc,
Fresh eyes are always a great checkpoint on our docs. 

Notary v2 is focused on signing artifacts in a registry, with the ability to move within and across registries. 

The goal is to enable the above. Rather than focus overly narrowly to just enable signing, we’re looking at generalization changes to existing specs to enable Notary v2 requirements. 

The ability to enable additional scenarios are beneficial. 

> "I just want to sign a container image", I've had a hard time understanding what this will all look like when complete.

See the nv2 verify prototype-1 for this example. However, I’ll take this as an action item to highlight the sign & verify steps in an update to the status report.

dlorenc:
I'm still trying to make sure I understand the end state here. Is "Notary v2" a tool or service I will use directly? Is it a specification that registries implement separately from OCI? From your answer, it sounds like it's none of these things, it is simply an effort to make sure certain things are possible.

I'll try repeating that back to make sure I understand:

Notary v2 is an effort to establish a set of core requirements and scenarios, then an effort to make sure those scenarios are enabled by working with other specifications, projects and tools to make any necessary changes.

Does that sound close?

notary-status.md
SteveLasker: Update pushed. Please re-open if I didn't capture the question.

SteveLasker:
> Is "Notary v2" a tool or service I will use directly? Is it a specification that registries implement separately from OCI? From your answer, it sounds like it's none of these things, it is simply an effort to make sure certain things are possible.

Notary v2 is, will be, a spec for:
- Defining a Notary v2 signature

To support the e2e workflows, the Notary v2 spec will define dependencies on:
- OCI Distribution for persisting and retrieving Notary v2 signatures
- OCI Artifacts for linking Notary v2 signatures to the artifact they are signing.

I've pushed updates and changed from **draft** to **ready for review**

SteveLasker:
Trying to merge in the requirements with a status report, I realized this is better suited in a notaryproject root repo for the overall notaryproject org. 
Please see: https://github.com/notaryproject/notaryproject/pull/1

PR #76, opened 2021-06-21
Attacker goals

mnm678:
supersedes #35

Adds an attacker goals and out of scope section to the threat model.

### DIFF
--- a/threatmodel.md
+++ b/threatmodel.md
@@ -10,3 +10,8 @@ It is assumed that an attacker may perform one or more the following actions:
 
 While it is not always possible to protect against all scenarios, the system should to the extent possible mitigate and/or reduce the damage caused by a successful attack, detect the occurrence of an attack and notify appropriate parties, yet remain usable for parties operating the system.  Furthermore, the system should recover from successful attacks in a way that presents low operational overhead and risk to users.
 
+Attacker Goals:
+1. Trying to have a party install a malicious image under the attackers control.
+2. Trying to have a party install an outdated image.  For example, one with known security vulnerabilities.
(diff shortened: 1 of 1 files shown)


threatmodel.md
```diff
@@ -6,7 +6,20 @@ It is assumed that an attacker may perform one or more the following actions:
...
 
+Attacker Goals:
+1. To have a party deploy a malicious artifact under the attacker's control.
+2. To have a party deploy an artifact using a digest or tag that does not have a currently valid signature.  For example, a previous version of an artifact with a signature that was revoked by the signer, or an old version associated with a tag.
+3. To have a party deploy a different artifact than the one requested, including an older version when the latest is requested.
+4. Disrupt the verification of artifact signatures, for example by making the current version of metadata unavailable.
+5. Prevent a party from learning about updates to currently installed artifacts.
+6. Convince a party to download large amounts of data, such as signatures or metadata, that interfere with the party's system.
```
mnm678: We discussed removing this in favor of the registry-defined data size, but upon reflection I think there could be value in including this limit in signed and verifiable metadata (so that an attacker couldn't alter it in transit, etc). But I'd appreciate input from folks on the registry side.

mnm678: I moved this requirement to out of scope, with a note that the registry should consider this attack (like  opencontainers/distribution-spec#260)

sudo-bmitch (approved):

SteveLasker (approved):
LGTM

yizha1:
@mnm678 Would you mind closing this PR since no activities for more than 1 year? You can create a new issue to describe the problem if needed. Thanks.

PR #102, opened 2021-10-08
Signature specification

priteshbandi:
**Summary**
First draft of signature specification.
- Added support for multiple signature format.
- Added specs for JWS JSON serialization signature format. This is intended to be default format for alpha releases. The signature format for RC is TBD.

**issues for future work**

- https://github.com/notaryproject/notaryproject/issues/103
- https://github.com/notaryproject/notaryproject/issues/106
- https://github.com/notaryproject/notaryproject/issues/105

 **How is this PR different from #93?**
- #93 was mostly about requirements and discussions/comparisons around various signature formats while this PR is specification.
- #93 only supported one signature format while this PR advocates about supporting multiple.


signature-specification.md
```diff
@@ -0,0 +1,230 @@
...
+The document consists of following sections:
+* **[Storage](#storage)**: Describes how signatures are stored in OCI registry
+* **[Signature Envelope](#signature-envelope)**: Describes how signatures are created.
+
+
+## Storage
+This section describes how Notary v2 signatures are stored in the OCI Distribution conformant registry.
+Notary v2 uses [ORAS artifact manifest](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md) to store the signature in the repository. The signature artifact manifest consists of a signature type, a reference to the manifest of the artifact being signed, a blob referencing to the signature, and a collection of annotations.
+
+-<img src="media/signature-specification.jpg" width="550" height="450">
```
shizhMSFT: Replace the HTML tag with a Markdown tag for the image.

signature-specification.md
```diff
@@ -0,0 +1,230 @@
...
+
+- **`artifactType`**: Used to identify the Notary signature artifact. The supported value is `application/vnd.cncf.notary.v2`.
+- **`blobs`**: A collection consisting of only one [artifact descriptor](https://github.com/oras-project/artifacts-spec/blob/main/descriptor.md) that refers to a signature envelope.
+   - **`mediaType`**: Defines the media type of signature envelope blob. The supported value is `application/jose+json`
```
shizhMSFT: For the media type `application/jose+json`, will `application/vnd.cncf.notary.signature.v2.jws+json` be better?

signature-specification.md
```diff
@@ -0,0 +1,230 @@
...
+* Cryptographic signatures(s): The digital signatures computed on payload and signed attributes.
+
+A signature envelope is `e = {m, v, u, s}` where `s` is signature.
+
+Notary v2 supports [JWS JSON Serialization](https://datatracker.ietf.org/doc/html/rfc7515) as signature envelope format with some additional constraints but is extensible to support any other signature envelope format.
+
+### Payload
+Notary v2 requires Payload to be the [descriptor](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#properties) of the target manifest that is being signed.  
+1. Descriptor MUST contain `mediaType`, `digest`, `size` fields.
+1. Descriptor MAY contain `artifactType` field for artifact manifests, or the `config.mediaType` for `oci.image` based manifests or annotations.
```
shizhMSFT: All optional fields can go to the **signed attribtues**.

signature-specification.md
```diff
@@ -0,0 +1,230 @@
...
+               ...
+           },
+           "custom" : {
+               "customKey1": "customValue1",
+               "customKey2": "customValue2"
+               ...
+           }
+       }
+   },
+   "iat": 0123456789101,
```
shizhMSFT: nit: numbers do not start with `0`.

signature-specification.md
```diff
@@ -0,0 +1,230 @@
...
+       }
+   ]
+}
+```
+If a JWS contains only one signature as above, the JWS can be flattened as
+
+```json
+{
+   "payload": "<Base64Url(JWSPayload)>",
+   "protected": "<Base64Url(ProtectedHeaders)>"
```
shizhMSFT: nit: missing `,` at the EOL.

SteveLasker:
Incorporates #93

signature-specification.md
```diff
@@ -0,0 +1,230 @@
...
+     "size": 32654
+   }
+ ],
+ "subject": {
+   "mediaType": "application/vnd.oci.image.manifest.v1+json",
+   "digest": "sha256:73c803930ea3ba1e54bc25c2bdc53edd0284c62ed651fe7b00369da519a3c333",
+   "size": 16724
+ },
+ "annotations": {
+   "org.cncf.notary.signature.subject": "wabbit-networks.io"
```
SteveLasker, line 37: Noting we'll want to update the annotation, per #103

SteveLasker, line 35: Tracking the issue to finalize the annotation details: #103

signature-specification.md
priteshbandi: Added `application/jose+json` because its already registered in [iana](https://www.iana.org/assignments/media-types/media-types.xhtml). Also, its actually an serialized JWS JSON signature.

signature-specification.md
priteshbandi: Technically yes but the idea behind of payload and signed attributes is different. Payload is the actual content we want to sign(integrity protect) while attributes is metadata for the signature or payload and in this case annotation, artifactType or  any other  descriptor data is a payload.

signature-specification.md
priteshbandi: Github won't recognize the syntax of [](.. =400) or [](... =400px), so to resize the image I've to use the `` syntax.

signature-specification.md
SteveLasker: After re-reading this, the `mediaType` hers is the blob of the signature. We had similar feedback in OCI Artifact conversations to use existing `mediaTypes` when possible. So, `application/jose~json` seems good, particularly as it's already registered with iana.org

signature-specification.md
```diff
@@ -0,0 +1,228 @@
...
+This document describes how Notary v2 signatures are created and stored.
+The document consists of following sections:
+* **[Storage](#storage)**: Describes how signatures are stored in OCI registry
+* **[Signature Envelope](#signature-envelope)**: Describes how signatures are created.
+
+
+## Storage
+This section describes how Notary v2 signatures are stored in the OCI Distribution conformant registry.
+Notary v2 uses [ORAS artifact manifest](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md) to store the signature in the repository. The signature artifact manifest consists of a signature type, a reference to the manifest of the artifact being signed, a blob referencing to the signature, and a collection of annotations.
+<img src="media/signature-specification.jpg" width="550" height="450">
```
SteveLasker: A picture helps with clarity. 
A few clarifications:
- The manifest that contains the signature is of type `application/vnd.cncf.oras.artifact.manifest.v1+json`
- The top box suggests it can only sign an OCI image manifest. I'd suggest changing this to simply be "Manifest". See feedback below for the `subject` property.
- Suggest taking the top box within the two lower boxes and make them a header on the box. I was reading it as the content, not the description of the content.

signature-specification.md
```diff
@@ -0,0 +1,228 @@
...
+
+- **`artifactType`**: Used to identify the Notary signature artifact. The supported value is `application/vnd.cncf.notary.v2`.
+- **`blobs`**: A collection consisting of only one [artifact descriptor](https://github.com/oras-project/artifacts-spec/blob/main/descriptor.md) that refers to a signature envelope.
+   - **`mediaType`**: Defines the media type of signature envelope blob. The supported value is `application/jose+json`
+- **`subject`**: An artifact descriptor referencing to image manifest that is signed.
```
SteveLasker: The `subject`, by which the signature is associated with can be any manifest type which means Notary v2 can sign any artifact from container images to anything else that follows the conventions in [OCI Artifacts authors guidance](https://github.com/opencontainers/artifacts/blob/master/artifact-authors.md), a multi-arch manifest of images, or other artifacts that use the ORAS Artifact manifest for things like SBoMs, scan results and other reference types

Suggest: A REQUIRED [artifact descriptor](https://github.com/oras-project/artifacts-spec/blob/main/descriptor.md) referencing the signed manifest, including, but not limited to ([oci image manifest](https://github.com/opencontainers/image-spec/blob/main/manifest.md), [oci image index](https://github.com/opencontainers/image-spec/blob/main/image-index.md), [oras-artifact manifest](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md)).

signature-specification.md
```diff
@@ -0,0 +1,228 @@
...
+* **[Storage](#storage)**: Describes how signatures are stored in OCI registry
+* **[Signature Envelope](#signature-envelope)**: Describes how signatures are created.
+
+
+## Storage
+This section describes how Notary v2 signatures are stored in the OCI Distribution conformant registry.
+Notary v2 uses [ORAS artifact manifest](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md) to store the signature in the repository. The signature artifact manifest consists of a signature type, a reference to the manifest of the artifact being signed, a blob referencing to the signature, and a collection of annotations.
+<img src="media/signature-specification.jpg" width="550" height="450">
+
+- **`artifactType`**: Used to identify the Notary signature artifact. The supported value is `application/vnd.cncf.notary.v2`.
```
SteveLasker: This REQUIRED property references the Notary version of the signature: `application/vnd.cncf.notary.v2`

signature-specification.md
```diff
@@ -0,0 +1,228 @@
...
+   "size": 16724
+ },
+ "annotations": {
+   "org.cncf.notary.signature.subject": "wabbit-networks.io"
+ }
+}
+```
+
+### Signature Discovery
+The client should be able to discover all the signatures belonging to an artifact(such as image manifest) by using [ORAS Manifest Referrers API](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md). ORAS Manifest Referrers API returns a paginated list of all artifacts belonging to a target artifact(such as image manifest). The implementation can filter Notary signature artifacts by either using ORAS Manifest Referrers API or using custom logic on the client. Each Notary signature artifact refers to a signature envelope blob.
```
SteveLasker: > (such as image manifest)

( such as container images, sboms)

signature-specification.md
```diff
@@ -0,0 +1,228 @@
...
+}
+```
+
+### Signature Discovery
+The client should be able to discover all the signatures belonging to an artifact(such as image manifest) by using [ORAS Manifest Referrers API](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md). ORAS Manifest Referrers API returns a paginated list of all artifacts belonging to a target artifact(such as image manifest). The implementation can filter Notary signature artifacts by either using ORAS Manifest Referrers API or using custom logic on the client. Each Notary signature artifact refers to a signature envelope blob.  
+
+
+## Signature Envelope
+The Signature Envelope is a standard data structure for creating a signed message. A signature envelope consists of the following components:
+* Payload(m): The data that is integrity protected - e.g. container image descriptor.
```
SteveLasker: > e.g. container image descriptor

descriptor of the artifact being signed

signature-specification.md
```diff
@@ -0,0 +1,228 @@
...
+* Signed attributes(v): The signature metadata that is integrity protected - e.g. signature expiration time, signing time, etc.
+* Unsigned attributes(u): This OPTIONAL property represents signature metadata that is not integrity protected - e.g. timestamp, certificates, etc.
+* Cryptographic signatures(s): The digital signatures computed on payload and signed attributes.
+
+A signature envelope is `e = {m, v, u, s}` where `s` is signature.
+
+Notary v2 supports [JWS JSON Serialization](https://datatracker.ietf.org/doc/html/rfc7515) as signature envelope format with some additional constraints but is extensible to support any other signature envelope format.
+
+### Payload
+Notary v2 requires Payload to be the [descriptor](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#properties) of the target manifest that is being signed.
```
SteveLasker: > of the target manifest that is being signed.  

of the `subject` manifest that is being signed.

signature-specification.md
```diff
@@ -0,0 +1,228 @@
...
+Notary v2 requires Payload to be the [descriptor](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#properties) of the target manifest that is being signed.  
+1. Descriptor MUST contain `mediaType`, `digest`, `size` fields.
+1. Descriptor MAY contain `artifactType` field for artifact manifests, or the `config.mediaType` for `oci.image` based manifests or annotations.
+```json
+{
+   "mediaType": "application/vnd.oci.image.manifest.v1+json",
+   "digest": "sha256:73c803930ea3ba1e54bc25c2bdc53edd0284c62ed651fe7b00369da519a3c333",
+   "size": 16724
+}
+```
```
SteveLasker, line 68: Suggest adding an example where artifactType would be added
```json
{
   "mediaType": "sbom/example",
   "digest": "sha256:e6688e911f15729983816cb784a4ae88c7405593c30bc88f6216e24c3363bf26",
   "size": 837
}
```

signature-specification.md
```diff
@@ -0,0 +1,228 @@
...
+Notary v2 is using one private claim(notary) and two public claims(iat and exp). An example of the claim is described below
+```json
+{
+   "notary": {  
+       "subject": {
+           "mediaType": "application/vnd.oci.image.manifest.v1+json",
+           "digest": "sha256:73c803930ea3ba1e54bc25c2bdc53edd0284c62ed651fe7b00369da519a3c333",
+           "size": 16724
+       },
+       "attributes": {
```
sudo-bmitch: The attributes could be compressed into annotations on the descriptor, and we could remove an extra layer in this payload. That would have the advantage of passing those annotations through to anything using the descriptor downstream.

signature-specification.md
```diff
@@ -0,0 +1,228 @@
...
+           "header": {
+               "timestamp": "<Base64Url(TimeStampToken)>",
+               "x5c": ["<Base64(DER(leafCert))>", "<Base64(DER(intermediateCACert))>", "<Base64(DER(rootCert))>"]
+           },
+           "signature": "Base64Url( sign( ASCII( <Base64Url(JWSPayload)>.<Base64Url(ProtectedHeader)> )))"  
+       }
+   ]
+}
+```
+If a JWS contains only one signature as above, the JWS can be flattened as
```
sudo-bmitch: Are there any scenarios where multiple signatures would be pushed together, rather than separate artifacts?

sudo-bmitch:
Can we clarify what is being decided in this PR vs #93

signature-specification.md
priteshbandi: Addressed

signature-specification.md
priteshbandi: incorporated feedback

signature-specification.md
priteshbandi: incorporated feedback

signature-specification.md
priteshbandi: incorporated feedback...removed attributes and used annotations for this.

signature-specification.md
priteshbandi: As of now we don't have scenario for multiple signature. removed JWS JSON format that supports multiple signature i.e. we only support flattened JWS JSON format.

signature-specification.md
```diff
@@ -0,0 +1,218 @@
+# Signature Specification
+
+This document describes how Notary v2 signatures are created and stored.
+The document consists of following sections:
```
gokarnm:
```suggestion
The document has following sections:
```

signature-specification.md
```diff
@@ -0,0 +1,218 @@
...
+* **[Storage](#storage)**: Describes how signatures are stored in OCI registry
+* **[Signature Envelope](#signature-envelope)**: Describes how signatures are created.
+
+
+## Storage
+This section describes how Notary v2 signatures are stored in the OCI Distribution conformant registry.
+Notary v2 uses [ORAS artifact manifest](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md) to store the signature in the repository. The manifest that contains the signature is of type `application/vnd.cncf.oras.artifact.manifest.v1+json`. The signature artifact manifest consists of a signature type, a reference to the manifest of the artifact being signed, a blob referencing to the signature, and a collection of annotations.
+![Signature storage inside registry](media/signature-specification.svg)
+
+- **`artifactType`**(*string*): This REQUIRED property references the Notary version of the signature: `application/vnd.cncf.notary.v2`.
```
gokarnm: Shouldn't this be more specific artifact type like `application/vnd.cncf.notary.v2.signature`

signature-specification.md
```diff
@@ -0,0 +1,218 @@
...
+
+This document describes how Notary v2 signatures are created and stored.
+The document consists of following sections:
+* **[Storage](#storage)**: Describes how signatures are stored in OCI registry
+* **[Signature Envelope](#signature-envelope)**: Describes how signatures are created.
+
+
+## Storage
+This section describes how Notary v2 signatures are stored in the OCI Distribution conformant registry.
+Notary v2 uses [ORAS artifact manifest](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md) to store the signature in the repository. The manifest that contains the signature is of type `application/vnd.cncf.oras.artifact.manifest.v1+json`. The signature artifact manifest consists of a signature type, a reference to the manifest of the artifact being signed, a blob referencing to the signature, and a collection of annotations.
```
gokarnm:
```suggestion
Notary v2 uses [ORAS artifact manifest](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md) to store the signature in the repository. The media type of the signature manifest is `application/vnd.cncf.oras.artifact.manifest.v1+json`. The signature manifest has an artifact type which specifies it's a Notary V2 signature, a reference to the manifest of the artifact being signed, a blob referencing the signature, and a collection of annotations.
```

signature-specification.md
```diff
@@ -0,0 +1,218 @@
...
+![Signature storage inside registry](media/signature-specification.svg)
+
+- **`artifactType`**(*string*): This REQUIRED property references the Notary version of the signature: `application/vnd.cncf.notary.v2`.
+- **`blobs`**(*array of objects*): This REQUIRED property contains collection of only one [artifact descriptor](https://github.com/oras-project/artifacts-spec/blob/main/descriptor.md) referencing signature envelope.
+   - **`mediaType`**(*string*): This REQUIRED property contains media type of signature envelope blob. The supported value is `application/jose+json`
+- **`subject`**(*descriptor*): A REQUIRED artifact descriptor referencing the signed manifest, including, but not limited to image manifest, image index, oras-artifact manifest.
+- **`annotations`**(*string-string map*): This OPTIONAL property contains arbitrary metadata for the artifact manifest. It can be used to store information about the signature.
+```json
+{
+ "artifactType": "application/vnd.cncf.notary.v2",
```
gokarnm: We can update this if we agree on the artifactType

signature-specification.md
```diff
@@ -0,0 +1,218 @@
...
+## Signature Envelope
+The Signature Envelope is a standard data structure for creating a signed message. A signature envelope consists of the following components:
+* Payload(m): The data that is integrity protected - e.g. descriptor of the artifact being signed.
+* Signed attributes(v): The signature metadata that is integrity protected - e.g. signature expiration time, signing time, etc.
+* Unsigned attributes(u): This OPTIONAL property represents signature metadata that is not integrity protected - e.g. timestamp, certificates, etc.
+* Cryptographic signatures(s): The digital signatures computed on payload and signed attributes.
+
+A signature envelope is `e = {m, v, u, s}` where `s` is signature.
+
+Notary v2 supports [JWS JSON Serialization](https://datatracker.ietf.org/doc/html/rfc7515) as signature envelope format with some additional constraints but is extensible to support any other signature envelope format.
```
gokarnm: I'd add a line saying support for additional signatures in Notation lib requires updates to Notation lib.

signature-specification.md
```diff
@@ -0,0 +1,218 @@
...
+   "size": 16724
+}
+```
+
+```json
+{
+   "mediaType": "sbom/example",
+   "digest": "sha256:9834876dcfb05cb167a5c24953eba58c4ac89b1adf57f28f2f9d09af107ee8f0",
+   "size": 32654
+}
```
gokarnm, line 75: Will this also have an artifactType as described in previous section.

signature-specification.md
```diff
@@ -0,0 +1,218 @@
...
+       }
+   },
+   "iat": 1234567891000,
+   "exp": 1234567891011
+}
+```
+The payload contains the subject manifest and other attributes that have to be integrity protected.
+
+* **`notary`**(*string-string map*): This is a REQUIRED top-level node and a private claim encapsulating the notary v2 data.
+* **`subject`**(*descriptor*): A REQUIRED manifest that needs to be integrity protected.
```
gokarnm: Do you need to document this again, it was already covered in Payload section, with artifactType etc. being optional.

signature-specification.md
```diff
@@ -0,0 +1,218 @@
...
+Notary v2 supports only three protected headers: alg, cty, and crit.
+```json
+{
+   "alg": "RS256",
+   "cty": "application/vnd.cncf.notary.v2.jws.v1",
+   "crit":["cty"]
+}
+```
+* **`alg`**(*string*): This REQUIRED property defines which algorithm was used to generate the signature. JWS needs an algorithm(alg) to be present in the header, so we have added it as a protected header.
+* **`cty`**(*string*): The REQUIRED property content-type(cty) is used to declare the media type of the secured content(the payload). This will be used to version different variotions of JWS signature. The supported value is `application/vnd.cncf.notary.v2.jws.v0`.
```
gokarnm: The version is different than the example.

gokarnm:
```suggestion
* **`cty`**(*string*): The REQUIRED property content-type(cty) is used to declare the media type of the secured content(the payload). This will be used to version different variations of JWS signature. The supported value is `application/vnd.cncf.notary.v2.jws.v0`.
```

signature-specification.md
priteshbandi: it can have it but its optional

signature-specification.md
priteshbandi: yeah to support multiple `artifactType` for notary it makes sense to add `.signature` suffix.

gokarnm (approved):

NiazFK (approved):
LGTM

signature-specification.md
```diff
@@ -0,0 +1,215 @@
...
+```
+* **`alg`**(*string*): This REQUIRED property defines which algorithm was used to generate the signature. JWS needs an algorithm(alg) to be present in the header, so we have added it as a protected header.
+* **`cty`**(*string*): The REQUIRED property content-type(cty) is used to declare the media type of the secured content(the payload). This will be used to version different variations of JWS signature. The supported value is `application/vnd.cncf.notary.v2.jws.v1`.
+* **`crit`**(*array of strings*): This REQUIRED property lists the headers that implementation MUST understand and process. The value MUST be `["cty"]`.
+
+**UnprotectedHeaders**:
+Notary v2 supports only two unprotected headers: timestamp and x5c.
+```
+{
+   "timestamp": "<Base64Url(TimeStampToken)>",
```
shizhMSFT: Timestamp should just be `Base64` encoded, not `Base64URL` encoded.

signature-specification.md
```diff
@@ -0,0 +1,215 @@
...
+   "x5c": ["<Base64(DER(leafCert))>", "<Base64(DER(intermediateCACert))>", "<Base64(DER(rootCert))>"]
+}
+```
+* **`timestamp`**(*string*): This OPTIONAL property is used to store time stamp token. Only [RFC3161]([rfc3161](https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2)) compliant TimeStampToken are supported.
+* **`x5c`**(*array of strings*): This REQUIRED property contains the list of X.509 certificate or certificate chain([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the JWS.
```
shizhMSFT: You may mention:

> The certificate containing the public key corresponding to the key used to digitally sign the JWS MUST be the first certificate.

as stated in [RFC 7515 4.1.6](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.6).

signature-specification.md
```diff
@@ -0,0 +1,215 @@
...
+**Signature Envelope**:
+The final signature envelope comprises of Claims, ProtectedHeaders, UnprotectedHeaders, and signature.
+
+Since Notary v2 restricts one signature per signature envelope, the compliant signature envelope MUST be in flattened JWS JSON format.
+```json
+{
+   "payload": "<Base64Url(JWSPayload)>",
+   "protected": "<Base64Url(ProtectedHeaders)>",
+   "header": {
+     "timestamp": "<Base64Url(TimeStampToken)>",
```
shizhMSFT: `base64` instead of `base64url` for `timestamp`.

signature-specification.md
```diff
@@ -0,0 +1,215 @@
...
+Notary v2 implementation MUST enforce the following constraints on signature generation and verification:
+1. `alg` header value MUST NOT be `none` or any symmetric-key algorithm such as `HMAC`.
+2. `alg` header value MUST be same as that of signature algorithm identified using signing certificate's public key algorithm and size.
+3. `alg` header values for various signature algorithms:
+   | Signature Algorithm    | "alg" Param Value |
+   | --------    | -------- |
+   | ECDSA on secp256r1 with SHA-256 | ES256     |
+   | ECDSA on secp256r1 with SHA-512 | ES256     |
+   | RSASSA-PSS with SHA-256   | PS256     |
+   | RSASSA-PSS with SHA-512   | PS512     |
```
shizhMSFT: Need to format the table.

shizhMSFT: Missing `ES384` and `PS384`.

signature-specification.md
```diff
@@ -0,0 +1,215 @@
...
+
+**Implementation Constraints**
+Notary v2 implementation MUST enforce the following constraints on signature generation and verification:
+1. `alg` header value MUST NOT be `none` or any symmetric-key algorithm such as `HMAC`.
+2. `alg` header value MUST be same as that of signature algorithm identified using signing certificate's public key algorithm and size.
+3. `alg` header values for various signature algorithms:
+   | Signature Algorithm    | "alg" Param Value |
+   | --------    | -------- |
+   | ECDSA on secp256r1 with SHA-256 | ES256     |
+   | ECDSA on secp256r1 with SHA-512 | ES256     |
```
shizhMSFT: This should be `ES512`, shouldn't it?

signature-specification.md
```diff
@@ -0,0 +1,215 @@
...
+The signing certificate's public key algorithm and size MUST be used to determine the signature algorithm.
+
+|Public Key Algorithm| Key Size (bits)   | Signature Algorithm  |
+| -------- | --------    | -------- |
+|RSA| 2048 | RSASSA-PSS with SHA-256     |
+|RSA| 3072 | RSASSA-PSS with SHA-384     |
+|RSA| 4096 | RSASSA-PSS with SHA-512     |
+|ECDSA| 256   | ECDSA on secp256r1 with SHA-256     |
+|ECDSA| 384   | ECDSA on secp384r1 with SHA-384     |
+|ECDSA| 512   | ECDSA on secp521r1 with SHA-512     |
```
shizhMSFT: Table format

signature-specification.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+**Implementation Constraints**
+Notary v2 implementation MUST enforce the following constraints on signature generation and verification:
+1. `alg` header value MUST NOT be `none` or any symmetric-key algorithm such as `HMAC`.
+2. `alg` header value MUST be same as that of signature algorithm identified using signing certificate's public key algorithm and size.
+3. `alg` header values for various signature algorithms:
+   | Signature Algorithm    | "alg" Param Value |
+   | --------    | -------- |
+   | ECDSA on secp256r1 with SHA-256 | ES256     |
```
shizhMSFT: Can we have a more friendly name for `secp256r1` from SECG? like `P-256` from NIST?

Also, we need to provide a reference for readers. [RFC 8422 Appendix A](https://datatracker.ietf.org/doc/html/rfc8422#appendix-A) is a good one.

signature-specification.md
shizhMSFT: Pedantically, `ECDSA` is not a public key algorithm where `DSA` stands for **Digital Signature Algorithm**. Therefore, we should use `EC` here for **Elliptic Curve** keys.

shizhMSFT (commented):
General comments on formating. We should always do backquotes for words like `alg`, `iat`, and `exp`... even if they are in the description.

signature-specification.md
SteveLasker: @priteshbandi, does this need to change?

SteveLasker:
@priteshbandi, looks like a few minor clarification items are still open.

signature-specification.md
priteshbandi: Not all algos and curves are in NIST and that's the reason refrained from using NIST curves. 
Added reference to RFC 8422 Appendix A  for discovery.

signature-specification.md
priteshbandi: That was a miss, updated

signature-specification.md
priteshbandi: Added

signature-specification.md
priteshbandi: That was a miss, updated

signature-specification.md
priteshbandi: Yep...updated

signature-specification.md
priteshbandi: Added

SteveLasker (approved):
LGTM

PR #107, opened 2021-10-17
TrustStore and TrustPolicy specification

priteshbandi:
### Summary
This is initial draft for TrustStore and TrustPolicy specification.
- No scoping support
- Only x509 certificate support.
- Enforced complete certificate chain for signing certificate.
- Added high level signature evaluation process.

### References:
* https://github.com/notaryproject/notaryproject/discussions/100
* https://github.com/notaryproject/notaryproject/discussions/98
* More explanation on some decision and FAQs can be found at https://hackmd.io/lyVjt9s1QYWPwhK0gvfAxw#FAQ


trust-store-trust-policy-specification.md
```diff
@@ -0,0 +1,154 @@
+# Trust Store and Trust Policy Specification
+This document describes how Notary v2 signatures are evaluated for trust. The document comprises of following sections:
```
shizhMSFT:
```suggestion
This document describes how Notary v2 signatures are evaluated for trust. The document consists of following sections:
```

trust-store-trust-policy-specification.md
```diff
@@ -0,0 +1,154 @@
...
+* **`enforceWithFailClose`**: This means implementation MUST perform validation and if validation fails because the endpoint is not reachable, the implementation MUST log an error and MUST NOT fail the validation.
+* **`warn`**: This means implementation MUST perform the validation and if validation fails(because of any reason) the implementation MUST log an error and MUST NOT fail validation.
+* **`skip`**: This means implementation MUST NOT perform the validation.
+
+### Extended Validation
+The implementation must allow the user to execute custom validations. These custom validation MUST have access to all the information available in the signature envelope like payload, signed attributes, unsigned attributes, and signature.
+
+## Signature Evaluation
+Precondition: The artifact is signed, trust store and trust policies are present.
+1. Get the signing algorithm(hash+encyption) from the signing identity and validate that the signing algorithm is valid and allow-listed.
```
shizhMSFT:
```suggestion
1. Get the signing algorithm(hash+encryption) from the signing identity and validate that the signing algorithm is valid and allow-listed.
```

trust-store-trust-policy-specification.md
```diff
@@ -0,0 +1,154 @@
...
+
+If revocation validations are enforced implementation MUST perform the following validations:
+1. If signing identity is a certificate
+   1. then signing certificate and certificate-chain MUST NOT be revoked.
+   1. and if the signing certificate and certificate chain are not expired then the implementation MUST ignore the revocation check for timestamping signature even if it's present in the signature.
+   1. and if the signing certificate and certificate chain is expired then the implementation MUST validate that signature is timestamped and timestamping signature is valid. Also, validate that timestamping certificate and certificate chain MUST NOT be revoked.
+
+The implementation MUST support both [OCSP](https://datatracker.ietf.org/doc/html/rfc6960) and [CRL](https://datatracker.ietf.org/doc/html/rfc5280) based revocations. Since revocation check requires network call and network call can fail because of a variety of reasons such as revocation endpoint is unavailable, network connectivity issue, DDOS attack, etc the implementation MUST support both `fail-open` or `fail-close` use cases.
```
shizhMSFT:
```suggestion
The implementation MUST support both [OCSP](https://datatracker.ietf.org/doc/html/rfc6960) and [CRL](https://datatracker.ietf.org/doc/html/rfc5280) based revocations. Since revocation check requires network call and network call can fail because of a variety of reasons such as revocation endpoint is unavailable, network connectivity issue, DDoS attack, etc the implementation MUST support both `fail-open` or `fail-close` use cases.
```

trust-store-trust-policy-specification.md
```diff
@@ -0,0 +1,154 @@
...
+The trust store is represented as JSON data structure as shown below:
+
+```json
+{
+    "version": "1.0",
+    "trustStores": {
+        "trust-store-name-1": {
+            "identities": {
+                "x509Certs": [
+                    "-----BEGIN CERTIFICATE-----certificate1-----END CERTIFICATE-----",
```
shizhMSFT: A valid PEM file should have line returns.
```suggestion
                    "-----BEGIN CERTIFICATE-----\ncertificate1\n-----END CERTIFICATE-----",
```

trust-store-trust-policy-specification.md
```diff
@@ -0,0 +1,154 @@
...
+            }
+        }
+    }
+}
+```
+Property Description:
+* **`version`**(*string*): This REQUIRED property is the version of the trust store. The supported value is `1.0`
+* **`trustStores`**(*object*): This REQUIRED property represents the parent node containing multiple trust stores. Each trust store is identified by the key associated with it like 'trust-store-name-1', 'trust-store-name-2'.
+   * **`identities`**(*object*): This REQUIRED property represents the collection of different types of identities. There are two types of identifies that Notary v2 supports: x509 certificates and x509 timestamping certificates.
+       * **`x509Certs`**(*array of strings*): This REQUIRED property specifies a list of x509 certificates in PEM format. The collection MUST contain at least one certificate.
```
shizhMSFT: Instead of PEM, should we consider DER? Basically, PEM is just a base64 wrapper of the DER content.

trust-store-trust-policy-specification.md
```diff
@@ -0,0 +1,154 @@
...
+       * **`tsaX509Certs`**(*array of strings*): This Optional property specifies a list of x509 timestamping certificates in PEM format. If the `tsaX509Certs` key is present then collection MUST contain at least one timestamping certificate.
```
shizhMSFT:
```suggestion
       * **`tsaX509Certs`**(*array of strings*): This OPTIONAL property specifies a list of x509 timestamping certificates in PEM format. If the `tsaX509Certs` key is present then collection MUST contain at least one timestamping certificate.
```

trust-store-trust-policy-specification.md
```diff
@@ -0,0 +1,154 @@
...
+### Artifact Integrity
+The presence of a trust policy indicates that implementation MUST validate that artifact is signed and has not been altered.
+
+### Artifact Expiry
+Trust policy allows users to define how the system should behave when the artifact's signature is expired or signing identity is expired or timestamping identity is expired.
+
+If artifact expiry validations are enforced the implementation MUST perform the following validations:
+1. If signature expiry is present then signature MUST NOT be expired.
+1. If signing identity is certificate and   
+    1. signing certificate and certificate chain are not expired then the implementation MUST ignore the timestamping signature even if it is present in the signature.
```
shizhMSFT, line 76: I remember there was an option to force validate the timestamping siganture even if the signing certificate and the certificate chain are not expired. Is that still the case?

trust-store-trust-policy-specification.md
```diff
@@ -0,0 +1,154 @@
...
+    1. signing certificate and certificate chain are expired then the implementation MUST validate that signature is timestamped and timestamping signature is valid. Also, validate that timestamping certificate and certificate chain MUST NOT be expired.
```
shizhMSFT: What if the signing certificate is not expired but the certificate chain is expired?
```suggestion
    1. signing certificate or certificate chain are expired then the implementation MUST validate that signature is timestamped and timestamping signature is valid. Also, validate that timestamping certificate and certificate chain MUST NOT be expired.
```

trust-store-trust-policy-specification.md
```diff
@@ -0,0 +1,154 @@
...
+        "revocationValidations": {
+            "signingIdentityRevocation": "enforceWithFailOpen/enforceWithFailClose/warn/skip",
+            "timestampRevocation": "enforceWithFailOpen/enforceWithFailClose/warn/skip"
+        }
+    }
+}
+```
+Property descriptions
+* **`version`**(*string*): This REQUIRED property is the version of the trust policy. The supported value is `1.0`.
+* **`trustPolicies`**(*array of objects*): This REQUIRED property represents a collection of trust policies. As of now Notary v2 only supports one trust policy.
```
shizhMSFT: `trustPolicies` is an `object` or a `string-object map`. It is not an `array of objects`.

trust-store-trust-policy-specification.md
```diff
@@ -0,0 +1,154 @@
...
+       * **`timestampExpiry`**(*string*): This REQUIRED property specifies what implementation must do if timestamping certificate and certificate-chain are expired. Supported values are `enforce` and `warn`.
+   * **`revocationValidations`**(*object*): This REQUIRED property represents collection of artifact revocation related validations.
+       * **`signingIdentityRevocation`**(*string*): This REQUIRED property specifies whether implementation should check for signing identity(certificate and certificate-chain) revocation status or not and what implementation must do if this revocation check fails. Supported values are `enforceWithFailOpen`, `enforceWithFailClose`, `warn` and `skip`.
+       * **`timestampRevocation`**(*string*): This REQUIRED property specifies whether implementation should check for timestamping certificate and certificate-chain revocation status or not and what implementation must do if this revocation check fails. Supported values are `enforceWithFailOpen`, `enforceWithFailClose`, `warn` and `skip`.
```
shizhMSFT: We probably want to specify the default value of these properties.

shizhMSFT: or recommanded values

shizhMSFT: Do we still have `signatureRevocation`?

trust-store-trust-policy-specification.md
```diff
@@ -0,0 +1,154 @@
...
+### Extended Validation
+The implementation must allow the user to execute custom validations. These custom validation MUST have access to all the information available in the signature envelope like payload, signed attributes, unsigned attributes, and signature.
+
+## Signature Evaluation
+Precondition: The artifact is signed, trust store and trust policies are present.
+1. Get the signing algorithm(hash+encyption) from the signing identity and validate that the signing algorithm is valid and allow-listed.
+1. Get the public key from the signing identity and validate the artifact integrity using the public key and signing algorithm identified in the previous step.
+1. Get and validate TrustStore and TrustPolicy for correctness.
+1. Get the signing identity from the signed artifact and validate it against the identities configured in the trust store. The signing identity should match or leads to at least one of the trusted identities configured in the trust store.
+    1. If signing identity is certificate then validate that certificate and certificate-chain leads to self-signed root.
```
shizhMSFT, line 150: This line seems redundant.

trust-store-trust-policy-specification.md
```diff
@@ -0,0 +1,154 @@
...
+
+We don't support n out m signature requirement verification scheme. Validation succeeds if verification succeeds for at least one signature. 
+
+**Q3.** Should local revocation and TSA servers are listed in the trust policy to support disconnected environments?
+
+Not natively supported but a user can configure `revocationValidations` to `skip` and then use extended validations to check for revocation.
+
+**Q4.** Why do we need to include a complete certificate chain(leading to root) in the signature? 
+
+Without a complete certificate chain, the implementation won't be able to perform an exhaustive revocation check, which will lead to security issues, and that's the reason for enforcing complete certificate chain.
```
shizhMSFT, line 175: What if all revocation checks are skipped? In this case, do we still need a complete certificate chain?

shizhMSFT, line 175: Why do we need to do an exhaustive revocation check even for the certificates specified in the trust store?

media/trust-store-trust-policy-evaluation.svg
```diff
@@ -0,0 +1,66 @@
...
+@startuml
+start
+if(Is artifact signature valid?) then (yes)
```
shizhMSFT, line 4: We should validate the signing identity first, and then use the public key in the signing identity to verify the artifact signature although the overall validation will still fail due to bad signing identity after the success validation of the artifact signature with a bad public key. However, this may also lead to poor performance as there are lots of verification against the signing identity.

Therefore, we should validate the artifact signature first and then the signing identity but we need to document the reasoning behind it.

media/trust-store-trust-policy-evaluation.svg
```diff
@@ -0,0 +1,66 @@
...
+        if(Is signing identity trusted?) then (No)
+            stop
+        else (Yes)
+            if(Is 'signatureExpiry' validation enforced?) then (Yes)
+                 if(Is signature expired?) then  (Yes)
+                     stop
+                 else(No)
+                 endif
+            else(No)
+            endif
```
shizhMSFT, line 15: Shoud this be checked before the `signing identity`?

media/trust-store-trust-policy-evaluation.svg
```diff
@@ -0,0 +1,66 @@
...
+@startuml
+start
+if(Is artifact signature valid?) then (yes)
+	if(Is TrustStore and TrustPolicy valid?) then (Yes)
```
shizhMSFT, line 5: This should not be in the signature evaluation. Instead, this should be loaded when the verification service starts.

shizhMSFT (commented):
Suggest running a markdown linter for style checking.

trust-store-trust-policy-specification.md
priteshbandi: These are just placeholders to hint that cert is in pem format...There are multiple things here that doesn't confirm to [rfc7468](https://www.rfc-editor.org/rfc/rfc7468.html#section-3) like `\r\n` at end of line or content being not base64text,etc. 

Will try to make it as close as possible to rfc .

trust-store-trust-policy-specification.md
priteshbandi: So, skipping revocation checks or not is an artifact consumer decision while including a complete certificate chain in the signature envelope is the producer's decision. If a producer doesn't include a complete certificate chain then the consumer has no choice(and is vulnerable) that’s the reason we are enforcing a complete cert chain.

But if a use case arises where a producer cannot include a complete certificate chain we can revisit this decision.

priteshbandi: The certificate specified in the trust-store can get revoked moreover, the CA that issued that certificate can also get revoked(rare but possible event) thus resulting in distrust of all the cert that is issued.

https://hackmd.io/lyVjt9s1QYWPwhK0gvfAxw#Open-Questions explains this scenario in more detail.

trust-store-trust-policy-specification.md
priteshbandi: Nop, didn't include signature revocation in the initial version because we don't know how will it work, there is no specification around it. Once, we have the signature revocation specification, we will update the trust-store and trust policy spec to support it.

priteshbandi: I did think about that, the ideal default value(to provide max security) for all validations would be `enforce` but then that users will start using the default value without understanding its implications and one fine day their system will stop working because the signature expired or some cert was revoked or because of some other reason.
Not having default value forces users to think about the implications of validations in there system. 
Other way to deal with this is by using profiles where this trust policy will be pre-configured with default values; will cross that bridge when we get there.

trust-store-trust-policy-specification.md
priteshbandi: We are planning to use the timestamp as a counter signature that will allow signatures to be verified even after the signing certificate has expired or revoked([CRL](https://datatracker.ietf.org/doc/html/rfc5280#section-5.3.2)'s and [OCSP](https://datatracker.ietf.org/doc/html/rfc6960#section-4.4.5) has Invalidity Date; although non-critical extension but that's a separate discussion). By using timestamp the verifier can reliably know the time at which signature was affixed and thereby trusting the signature if it was valid at that time.

While force validating timestamp signature(if present) does add some additional layer of protection even when the certificate is not expired but it is easy to evade since the timestamp is an unsigned attribute and can be removed without invalidating the signature. Also as per requirement doc https://github.com/notaryproject/notaryproject/pull/92 [An artifact signature SHOULD be time-stamped with a timestamp signature](https://github.com/notaryproject/notaryproject/pull/92/files#diff-cf457aa764d8b910ca9d4d853f485a99596170d26149e2873ec9d57801483519R42), the timestamp is optional.

If we want to force validate the timestamp the reliable way to implement is by providing an option in trust policy to check for presence and validity of timestamp and fail the validation if that if the timestamp is not present.

I don't have any strong opinion here.

trust-store-trust-policy-specification.md
priteshbandi: DER is a binary format and is not user-friendly, while PEM is ASCII. With PEM users can copy-paste the cert in Truststore; With DER user will have to point to some file and then trust store has to read it from that file(json is not binary friendly), this indirection is not a good user experience.

```json
{
  "version": "1.0",
  "trustStores": {
    "trust-store-name-1": {
      "identities": {
        "x509Certs": [
          "-----BEGIN CERTIFICATE-----\nMIIFBjCCBK2gAwIBAgIQDovzdw2S0Zbwu2H5PEFmvjAKBggqhkjOPQQDAjBnMQsw\nCQYDVQQGEwJVUzEXMBUGA1UEChMORGlnaUNlcnQsIEluYy4xPzA9BgNVBAMTNkRp\nZ2lDZXJ0IEhpZ2ggQXNzdXJhbmNlIFRMUyBIeWJyaWQgRUNDIFNIQTI1NiAyMDIw\nIENBMTAeFw0yMTAzMjUwMDAwMDBaFw0yMjAzMzAyMzU5NTlaMGYxCzAJBgNVBAYT\nAlVTMRMwEQYDVQQIEwpDYWxpZm9ybmlhMRYwFAYDVQQHEw1TYW4gRnJhbmNpc2Nv\nMRUwEwYDVQQKEwxHaXRIdWIsIEluYy4xEzARBgNVBAMTCmdpdGh1Yi5jb20wWTAT\nBgcqhkjOPQIBBggqhkjOPQMBBwNCAASt9vd1sdNJVApdEHG93CUGSyIcoiNOn6H+\nudCMvTm8DCPHz5GmkFrYRasDE77BI3q5xMidR/aW4Ll2a1A2ZvcNo4IDOjCCAzYw\nHwYDVR0jBBgwFoAUUGGmoNI1xBEqII0fD6xC8M0pz0swHQYDVR0OBBYEFCexfp+7\nJplQ2PPDU1v+MRawux5yMCUGA1UdEQQeMByCCmdpdGh1Yi5jb22CDnd3dy5naXRo\ndWIuY29tMA4GA1UdDwEB/wQEAwIHgDAdBgNVHSUEFjAUBggrBgEFBQcDAQYIKwYB\nBQUHAwIwgbEGA1UdHwSBqTCBpjBRoE+gTYZLaHR0cDovL2NybDMuZGlnaWNlcnQu\nY29tL0RpZ2lDZXJ0SGlnaEFzc3VyYW5jZVRMU0h5YnJpZEVDQ1NIQTI1NjIwMjBD\nQTEuY3JsMFGgT6BNhktodHRwOi8vY3JsNC5kaWdpY2VydC5jb20vRGlnaUNlcnRI\naWdoQXNzdXJhbmNlVExTSHlicmlkRUNDU0hBMjU2MjAyMENBMS5jcmwwPgYDVR0g\nBDcwNTAzBgZngQwBAgIwKTAnBggrBgEFBQcCARYbaHR0cDovL3d3dy5kaWdpY2Vy\ndC5jb20vQ1BTMIGSBggrBgEFBQcBAQSBhTCBgjAkBggrBgEFBQcwAYYYaHR0cDov\nL29jc3AuZGlnaWNlcnQuY29tMFoGCCsGAQUFBzAChk5odHRwOi8vY2FjZXJ0cy5k\naWdpY2VydC5jb20vRGlnaUNlcnRIaWdoQXNzdXJhbmNlVExTSHlicmlkRUNDU0hB\nMjU2MjAyMENBMS5jcnQwDAYDVR0TAQH/BAIwADCCAQUGCisGAQQB1nkCBAIEgfYE\ngfMA8QB2ACl5vvCeOTkh8FZzn2Old+W+V32cYAr4+U1dJlwlXceEAAABeGq/vRoA\nAAQDAEcwRQIhAJ7miER//DRFnDJNn6uUhgau3WMt4vVfY5dGigulOdjXAiBIVCfR\nxjK1v4F31+sVaKzyyO7JAa0fzDQM7skQckSYWQB3ACJFRQdZVSRWlj+hL/H3bYbg\nIyZjrcBLf13Gg1xu4g8CAAABeGq/vTkAAAQDAEgwRgIhAJgAEkoJQRivBlwo7x67\n3oVsf1ip096WshZqmRCuL/JpAiEA3cX4rb3waLDLq4C48NSoUmcw56PwO/m2uwnQ\nprb+yh0wCgYIKoZIzj0EAwIDRwAwRAIgK+Kv7G+/KkWkNZg3PcQFp866Z7G6soxo\na4etSZ+SRlYCIBSiXS20Wc+yjD111nPzvQUCfsP4+DKZ3K+2GKsERD6d\n-----END CERTIFICATE-----"
        ]
      }
    }
  }
}
```

media/trust-store-trust-policy-evaluation.svg
priteshbandi: Will add this to FAQ.

media/trust-store-trust-policy-evaluation.svg
priteshbandi: yes but while validating we will need to make sure that implementation can understand the config.

media/trust-store-trust-policy-evaluation.svg
priteshbandi: The intention here is to not parse payload until we know we can trust the payload.

trust-store-trust-policy-specification.md
priteshbandi: Added this line to be verbose for certificate(s) scenario.

trust-store-trust-policy-specification.md
shizhMSFT: How about paths pointing to certificate files? That seems more user friendly.

trust-store-trust-policy-specification.md
shizhMSFT: > While force validating timestamp signature(if present) does add some additional layer of protection even when the certificate is not expired but **it is easy to evade since the timestamp is an unsigned attribute and can be removed without invalidating the signature**. 

This makes sense.

trust-store-trust-policy-specification.md
shizhMSFT: Can you document the above reasoning behind scene?

shizhMSFT (approved):
LGTM

trust-store-trust-policy-specification.md
priteshbandi: Yes, we can support both in-lining(PEM) and referencing(both PEM and DER) file but I was thinking more from storage and retrieval perspective where it would be easier to store trust  as one file than storing it as multiple files.

gokarnm:
LGTM

NiazFK (approved):
LGTM

SteveLasker (approved):
LGTM

PR #110, opened 2021-10-19
refine signature spec

### LINKED ISSUE #109: Inconsistent Description in Signature Specification
There are some inconsistent descriptions in the signature specification.

In the `Payload` section:
- `descriptor` points to an [OCI image spec descriptor](https://github.com/opencontainers/image-spec/blob/main/descriptor.md). However, it allows `artifactType`, which is from an [ORAS artifact spec descriptor](https://github.com/oras-project/artifacts-spec/blob/main/descriptor.md). The `descriptor` can even have `config.mediaType`, which does not exist in any descriptor specs.
- `notary` claim is a **string-object map**, not a **string-string map**.

In the `Supported Signature Envelopes` section:
- "ECDSA on secp256r1 with SHA-512" should be "ECDSA on **secp521r1** with SHA-512"

shizhMSFT:
Refine the formating of the signature spec markdown doc and resolves #109.

### DIFF
--- a/signature-specification.md
+++ b/signature-specification.md
@@ -202,7 +202,7 @@ Since Notary v2 restricts one signature per signature envelope, the compliant si
    | RSASSA-PSS with SHA-512         | PS512             |
    | ECDSA on secp256r1 with SHA-256 | ES256             |
    | ECDSA on secp384r1 with SHA-384 | ES384             |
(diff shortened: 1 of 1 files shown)


sudo-bmitch (approved):

priteshbandi (approved):
LGTM

SteveLasker (approved):
LGTM

PR #122, opened 2022-01-11
Added signing and verification workflows

priteshbandi:
## Summary
- Added signing and verification workflows
- Added detailed steps for signature verification
  - Added certificate requirements
  - Added certificate revocation

### Important call-outs
- In verification, we are evaluating timestamp(if present) irrespective of whether the signing certificate is expired or not. 
  - IMO if a user has added a timestamp to the signature then they want the verification system to use that timestamp.
  - Also, always evaluating timestamp(if present) avoid modes. e.g., if the timestamp is present but it's corrupted then the verification will continue to pass until the signing certificate hasn't expired but when the signing certificate expires, verification will start failing.
- During signature verification, we don’t support multiple certificate chains leading to multiple roots. If multiple certificate chains are leading to trusted roots, Notary v2 randomly picks one(usually, the one which is encountered first).
- Notary v2 doesn’t support building certificate chain by retrieving certificates from locations specified in the authority information access extension (AIA).

HackmdDoc: https://hackmd.io/@pritesh/signing-verfication-workflow

### DIFF
--- a/signature-specification.md
+++ b/signature-specification.md
@@ -18,7 +18,7 @@ The signature manifest has an artifact type which specifies it's a Notary V2 sig
 
 - **`artifactType`** (*string*): This REQUIRED property references the Notary version of the signature: `application/vnd.cncf.notary.v2.signature`.
 - **`blobs`** (*array of objects*): This REQUIRED property contains collection of only one [artifact descriptor](https://github.com/oras-project/artifacts-spec/blob/main/descriptor.md) referencing signature envelope.
-   - **`mediaType`** (*string*): This REQUIRED property contains media type of signature envelope blob. The supported value is `application/jose+json`
+  - **`mediaType`** (*string*): This REQUIRED property contains media type of signature envelope blob. The supported value is `application/jose+json`
 - **`subject`** (*descriptor*): A REQUIRED artifact descriptor referencing the signed manifest, including, but not limited to image manifest, image index, oras-artifact manifest.
 - **`annotations`** (*string-string map*): This OPTIONAL property contains arbitrary metadata for the artifact manifest. It can be used to store information about the signature.
 
@@ -204,7 +204,7 @@ Since Notary v2 restricts one signature per signature envelope, the compliant si
 4. Signing certificate MUST be a valid codesigning certificate.
 5. Only JWS JSON flattened format is supported. See 'Signature Envelope' section.
 
-## Signature Algorithms
+## Signature Algorithm Requirements
 
 The implementation MUST support the following set of algorithms:
 1. RSASSA-PSS with SHA-256
@@ -229,6 +229,28 @@ The signing certificate's public key algorithm and size MUST be used to determin
 | EC                   | 256             | ECDSA on secp256r1 with SHA-256 |
 | EC                   | 384             | ECDSA on secp384r1 with SHA-384 |
 | EC                   | 512             | ECDSA on secp521r1 with SHA-512 |
+
+### Certificate Requirements
+
+The signing certificate MUST meet the following minimum requirements:
+
+- The signing certificate MUST satisfy the following constraints:
+  - The keyUsage extension MUST be present and MUST be marked critical. The bit positions for digitalSignature MUST be set ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3)).
+  - The extKeyUsage extension MUST be present and its value MUST be id-kp-codeSigning ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12)).
+  - If the basicConstraints extension is present, the cA field MUST be set false ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9)).
+- The certificate MUST abide by the following key length restrictions:
+  - For RSA public key, the key length MUST be 2048 bits or higher.
+  - For ECDSA public key, the key length MUST be 256 bits or higher.
+
+The timestamping certificate MUST meet the following minimum requirements:
+
+- The timestamping certificate MUST satisfy the following constraints:
+  - The keyUsage extension MUST be present and MUST be marked critical. The bit positions for digitalSignature MUST be set ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3)).
+  - The extKeyUsage extension MUST be present and and MUST be marked critical. The value of extension MUST be id-kp-timeStamping ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12)).
+  - If the basicConstraints extension is present, the cA field MUST be set false ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9)).
+- The certificate MUST abide by the following key length restrictions:
+  - For RSA public key, the key length MUST be 2048 bits or higher.
+  - For ECDSA public key, the key length MUST be 256 bits or higher.
 
 ## FAQ
 
--- a/signature-specification.md
+++ b/signature-specification.md
@@ -232,22 +232,20 @@ The signing certificate's public key algorithm and size MUST be used to determin
 
 ### Certificate Requirements
 
-The signing certificate MUST meet the following minimum requirements:
+The **signing certificate** MUST meet the following minimum requirements:
 
-- The signing certificate MUST satisfy the following constraints:
-  - The keyUsage extension MUST be present and MUST be marked critical. The bit positions for digitalSignature MUST be set ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3)).
-  - The extKeyUsage extension MUST be present and its value MUST be id-kp-codeSigning ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12)).
-  - If the basicConstraints extension is present, the cA field MUST be set false ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9)).
+- The keyUsage extension MUST be present and MUST be marked critical. The bit positions for digitalSignature MUST be set ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3)).
+- The extKeyUsage extension MUST be present and its value MUST be id-kp-codeSigning ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12)).
+- If the basicConstraints extension is present, the cA field MUST be set false ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9)).
 - The certificate MUST abide by the following key length restrictions:
   - For RSA public key, the key length MUST be 2048 bits or higher.
   - For ECDSA public key, the key length MUST be 256 bits or higher.
 
-The timestamping certificate MUST meet the following minimum requirements:
+The **timestamping certificate** MUST meet the following minimum requirements:
 
-- The timestamping certificate MUST satisfy the following constraints:
-  - The keyUsage extension MUST be present and MUST be marked critical. The bit positions for digitalSignature MUST be set ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3)).
-  - The extKeyUsage extension MUST be present and and MUST be marked critical. The value of extension MUST be id-kp-timeStamping ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12)).
-  - If the basicConstraints extension is present, the cA field MUST be set false ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9)).
+- The keyUsage extension MUST be present and MUST be marked critical. The bit positions for digitalSignature MUST be set ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3)).
+- The extKeyUsage extension MUST be present and and MUST be marked critical. The value of extension MUST be id-kp-timeStamping ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12)).
+- If the basicConstraints extension is present, the cA field MUST be set false ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9)).
 - The certificate MUST abide by the following key length restrictions:
   - For RSA public key, the key length MUST be 2048 bits or higher.
   - For ECDSA public key, the key length MUST be 256 bits or higher.
--- a/signing-and-verification-workflow.md
+++ b/signing-and-verification-workflow.md
@@ -1,0 +1,56 @@
+# Signing and Verification Workflow
+
+This document describes how Notary v2 signs and verifies OCI artifacts.
+
+## Signing workflow
+
+The user wants to sign an OCI artifact and push the signature to a repository.
+
+### Prerequisite
+
+- User has access to the signing certificate and private key or a remote signing service through a notation plug-in.
+
+### Steps
+
+1. **Generate signature:** Using notation CLI or any other compliant signing tool, sign the OCI artifact. The signing tool should follow the following guideline.
+    1. Verify that the signing certificate is valid and satisfies [certificate requirements](./signature-specification.md#certificate-requirements).
+    1. Verify that the signing algorithm satisfies [algorithm requirements](./signature-specification.md#signature-algorithm-requirements).
+    1. Generate signature.
+        1. Generate signature using signature formats specified in [supported signature envelopes](./signature-specification.md#supported-signature-envelopes).
+        1. If the user wants to timestamp the signature, obtain an [RFC-3161](https://datatracker.ietf.org/doc/html/rfc3161.html) compliant timestamp for the signature generated in the previous step. Otherwise, continue to the next step.
+            1. Verify that the timestamp signing certificate satisfies [certificate requirements](./signature-specification.md#certificate-requirements).
+            1. Verify that the timestamp signing algorithm satisfies [algorithm requirements](./signature-specification.md#signature-algorithm-requirements).
+        1. Embed timestamp to the signature envelope.
+1. **Push the signature envelope:** Push the signature envelope generated in the previous step to the repository.
+1. **Generate signature artifact manifest:** As decribed in [signature specification](./signature-specification.md#storage) create the Notary v2 signature artifact manifest for the signature envelope generated in step 1.
+1. **Push signature artifact manifest:** Push Notary v2 signature artifact manifest to the repository.
+
+The user pushes the OCI artifact to the repository before the signature generation process as the signature reference must exist for the signature push to succeed. See: [ORAS Artifact Push Validation](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md#push-validation) for more info.
+
+## Verification workflow
+
+The user wants to pull an OCI artifact only if they are signed by a trusted publisher and the signature is valid.
+
+### Prerequisites
+
+- User has a fully qualified reference to an OCI artifact they want to pull. If the fully qualified artifact reference contains a tag then the user needs to resolve this tag to a digest.
+
+    E.g. If fully qualified reference is `wabbit-networks.io/software:Latest` where `Latest` is a tag pointing to an artifact. The user must resolve `Latest` tag to a digest and construct a new artifact reference using the resolved digest `wabbit-networks.io/software@sha256:${digest}`.
+- User has configured [trust store and trust policy](./trust-store-trust-policy-specification.md) required for signature verification.
+
+### Steps
+
+1. **Should Notary v2 verify the signature? :** Depending upon [trust-policy](./trust-store-trust-policy-specification.md#trust-policy) configuration, determine whether Notary v2 needs to verify the signature or not. If signature verification should be skipped for the given artifact, skip the below steps and directly jump to step 4.
+1. **Get signature artifact descriptors:** Using the ORAS [Manifest Referrers API](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md) download the Notary v2 signature artifact descriptors. The  `artifactType` parameter is set to Notary v2 signature's artifact type `application/vnd.cncf.notary.v2.signature`. Note that all ORAS implementations may not support filtering using `artifactType`.
+1. For each signature artifact descriptor, perform the following steps:
+    1. **Get signature artifact manifest:** Download the Notary v2 signature's artifact manifest for the given artifact descriptor.
+    1. **Filter signature artifact manifest:**
+        1. Filter out the unsupported signature formats by comparing the signature envelope format type (`[descriptors].descriptor.mediaType`) in the signature manifest, with the supported formats defined in [signature specification](./signature-specification.md#storage).
+        1. Depending upon the trust-store configuration, further filter out signature manifest. why? How? **TBD**(See [signature filtering](https://hackmd.io/@pritesh/signature_filtering)).
+        1. If the artifact manifest is filtered out, skip the below steps and move to the next signature artifact descriptor(step 2.1). If all signature artifact descriptors have already been processed, fail the signature verification and exit.
+    1. **Get and verify signatures:** On the filtered Notary v2 signature artifact manifest, perform the following steps:
+        1. Download the signature envelope.
+        1. Verify the signature envelope using trust-store and trust-policy as mentioned in [signature evaluation](./trust-store-trust-policy-specification.md#signature-evaluation) section.
+        1. If signature verification fails, skip the below steps and move to the next signature artifact descriptor(step 2.1). If all signature artifact manifests have already been processed, fail the signature verification and exit.
+        1. If signature verification succeeds, compare the digest derived from the given OCI artifact reference with the signed digest present in the signature envelope's payload. If digests are equal, signature verification is considered successful. Otherwise, move to the next signature artifact descriptor(step 2.1). If all signature artifact descriptors have already been processed, fail the signature verification and exit.
+1. **Get OCI artifact:** Using the verified digest, download the OCI artifact. This step is not in the purview of Notary v2.
--- a/trust-store-trust-policy-specification.md
+++ b/trust-store-trust-policy-specification.md
@@ -157,8 +157,8 @@ Property descriptions
 Value descriptions
 
 - **`enforce`**: This means implementation MUST perform validation and throw an error if validation fails.
-- **`enforceWithFailOpen`**: This means implementation MUST perform validation and if validation fails because the endpoint is not reachable, the implementation MUST throw an error and MUST fail the validation.
-- **`enforceWithFailClose`**: This means implementation MUST perform validation and if validation fails because the endpoint is not reachable, the implementation MUST log an error and MUST NOT fail the validation.
+- **`enforceWithFailOpen`**: This means implementation MUST perform validation and if validation fails because the endpoint is not reachable, the implementation MUST log an error and MUST NOT fail the validation.
+- **`enforceWithFailClose`**: This means implementation MUST perform validation and if validation fails because the endpoint is not reachable, the implementation MUST throw an error and MUST fail the validation.
 - **`warn`**: This means implementation MUST perform the validation and if validation fails(because of any reason) the implementation MUST log an error and MUST NOT fail validation.
 - **`skip`**: This means implementation MUST NOT perform the validation.
 
@@ -224,7 +224,7 @@ The implementation must allow the user to execute custom validations. These cust
 
 If the certificate revocation trust-store setting is set to `skip`, skip the below steps. Otherwise, check for revocation status for certificate and certificate chain.
 
-1. If the revocation status of any of the certificates cannot be determined (revocation unavailable) and `signingIdentityRevocation` is set to either `enforceWithFailClose` or `warn` then log a warning and skip the below steps. Otherwise, fail the signature validation and exit.
+1. If the revocation status of any of the certificates cannot be determined (revocation unavailable) and `signingIdentityRevocation` is set to either `enforceWithFailOpen` or `warn` then log a warning and skip the below steps. Otherwise, fail the signature validation and exit.
 1. If any of the certificates are revoked and `signingIdentityRevocation` is set to either `enforceWithFailOpen` or `enforceWithFailClose` then fail signature validation and exit else log a warning.
 
 Starting from Root to leaf certificate, for each certificate in the certificate chain, perform the following steps to check its revocation status:
--- a/trust-store-trust-policy-specification.md
+++ b/trust-store-trust-policy-specification.md
@@ -256,7 +256,7 @@ To check the revocation status of a certificate against CRL, the following steps
 1. Verify the CRL signature.
 1. Verify that the CRL is valid (not expired). A CRL is considered expired if the current date is after the `NextUpdate` field in the CRL.
 1. Look up the certificate’s serial number in the CRL.
-    1. If the certificate’s serial number is listed in the CRL, look for `InvalidityDate`. If the invalidity date is present and timestamp signature is also present then if the invalidity date is before the timestamping date, the certificate is considered revoked. If the invalidity date is not present in CRL, the certificate is considered revoked.
+    1. If the certificate’s serial number is listed in the CRL, look for `InvalidityDate`. If CRL has invalidity date and artifact signature is timestamped then compare the invalidity date with the timestamping date. If the invalidity date is before the timestamping date, the certificate is considered revoked. If the invalidity date is not present in CRL, the certificate is considered revoked.
     1. If the CRL is expired and the certificate is listed in the CRL for any reason other than `certificate hold`, the certificate is considered revoked.
     1. If the certificate is not listed in the CRL or the revocation reason is `certificate hold`, a new CRL is retrieved if the current time is past the time in the `NextUpdate` field in the current CRL. The new CRL is then checked to determine if the certificate is revoked. If the original reason was `certificate hold`, the CRL is checked to determine if the certificate is unrevoked by looking for the `RemoveFromCRL` revocation code.
 
--- a/trust-store-trust-policy-specification.md
+++ b/trust-store-trust-policy-specification.md
@@ -181,21 +181,110 @@ The implementation must allow the user to execute custom validations. These cust
 
 ## Signature Evaluation
 
-Precondition: The artifact is signed, trust store and trust policies are present.
+### Prerequisites
 
-1. Get the signing algorithm (hash+encryption) from the signing identity and validate that the signing algorithm is valid and allow-listed.
-1. Get the public key from the signing identity and validate the artifact integrity using the public key and signing algorithm identified in the previous step.
-1. Get and validate TrustStore and TrustPolicy for correctness.
-1. Find the trust policy that is applicable for the given artifact. In the trust policy if `skipSignatureVerification` is set to `true`, then the signature evaluation must be skipped.
-1. Get the signing identity from the signed artifact and validate it against the identities configured in the trust store of trust policy determined in step 4. The signing identity must match or lead to at least one of the trusted identities configured in the trust store.
-    1. If signing identity is certificate then validate that the certificate and certificate-chain leads to self-signed root.
-1. Perform [artifact expiry](#artifact-expiry) validations based on trust policy.
-1. Perform [artifact revocation](#artifact-revocation) validations based on trust policy
-1. Perform extended validation(If any).
+- User has configured [trust store](#trust-store) and [trust policy](#trust-policy).
 
-Here is high level uml diagram for signature evaluation:
+### Steps
 
-![uml diagram showing signature evaluation](./media/trust-store-trust-policy-evaluation.svg)
+1. **Validate that the signature envelope format is supported.**
+    1. Parse the signature envelope content based on the signature envelope type specified in the `[descriptors].descriptor.mediaType` attribute of the signature artifact manifest.
+    1. Validate that the content type indicated by the `cty` property value of protected headers in the signature envelope is supported.
+1. **Validate the signature envelope integrity.**
+    1. Get the signing certificate from the parsed [signature envelope](https://github.com/notaryproject/notaryproject/blob/7b7d283038/signature-specification.md#signature-envelope).
+    1. Get the signing algorithm(hash+encryption) from the signing certificate and validate that the signing algorithm satisfies [algorithm requirements](./signature-specification.md#signature-algorithm-requirements)
+    1. Using the public key of the signing certificate and signing algorithm identified in the previous step, validate the integrity of the signature envelope.
+1. **Validate the signature against trust policy and trust store.**
+    1. Using the `scope` configured in trust policies, get the applicable trust policy. (Implementations might have this value precomputed, added it for completeness)
+    1. For the applicable trust policy, **validate trust-store:**
+        1. Validate that signature envelope contains complete certificate chain that start from a code signing certificate and terminate with the root certificate. Also, validate that code signing certificate satisfies [certificate requirements](./signature-specification.md#certificate-requirements).
+        1. For each the trust-stores configured in applicable trust-policy perform the following steps.
+            1. Validate that certificate and certificate-chain lead to a trusted certificate configured in the `x509Certs` field of trust-store.
+            1. If the above verification succeeds then continue to the next step else iterate over the next trust store. If all of the trust stores have been evaluated then fail the signature validation and exit.
+    1. **Validate trust policy:**
+        1. If signature expiry is present in the signature envelope, using the local machine’s current time(in UTC) check whether the signature is expired or not. If the signature is not expired, continue to the next step. Otherwise, if `signatureExpiry` is set to `Enforce` then fail the signature validation and exit else log a warning and continue to the next step.
+        1. Check for the timestamp signature in the signature envelope.
+            1. If the timestamp exists, continue with the next step. Otherwise, store the local machine's current time(in  UTC) in variables `timeStampLowerLimit` and `timeStampUpperLimit` and continue with step 3.3.c.
+            1. Validate that the timestamp hash in `TSTInfo.messageImprint` matches the hash of the signature to which the timestamp was applied.
+            1. Validate that the timestamp signing certificate satisfies [certificate requirements](./signature-specification.md#certificate-requirements).
+            1. Validate that the timestamp signing algorithm satisfies [algorithm requirements](./signature-specification.md#signature-algorithm-requirements).
+            1. Validate the `signing-certificate`([RFC-2634](https://tools.ietf.org/html/rfc2634)) or `signing-certificate-v2`([RFC-5126](https://tools.ietf.org/html/rfc5126#section-5.7.3.2)) attribute of timestamp CMS.
+            1. Check whether timestamping certificate and certificate chain are valid (not expired) or not. If timestamping certificate and certificate-chain are not expired, continue to the next step. Otherwise, if `timestampExpiry` in trust-policy is configured to `Enforce` then fail the signature validation and exit, else log a warning and continue to the next step.
+            1. Validate that timestamp certificate and certificate chain leads to a trusted TSA certificate configured in trust policy.
+            1. Validate timestamp certificate and certificate chain revocation status  using [certificate revocation evaluation](#certificate-revocation-evaluation) section as per `timestampRevocation` setting in trust-policy
+            1. Retrieve the timestamp's time from `TSTInfo.genTime`.
+            1. Retrieve the timestamp's accuracy.  If the accuracy is explicitly specified in `TSTInfo.accuracy`, use that value.  If the accuracy is not explicitly specified and `TSTInfo.policy` is the baseline time-stamp policy([RFC-3628](https://tools.ietf.org/html/rfc3628#section-5.2)), use accuracy of 1 second.  Otherwise, use accuracy of 0.
+            1. Calculate the timestamp range using the lower and upper limits per [RFC-3161 section 2.4.2](https://tools.ietf.org/html/rfc3161#section-2.4.2) and store the limits as `timeStampLowerLimit` and `timeStampUpperLimit` variables respectively.
+        1. Check that the time range from `timeStampLowerLimit` to `timeStampUpperLimit` timestamp is entirely within the certificate's validity period.  If the time range is entirely within the signing certificate and certificate chain's validity period, continue to the next step.  Otherwise, If `signingIdentityExpiry` in trust-policy is configured to `Enforce` then fail the signature validation and exit else log a warning and continue to the next step.
+        1. Validate signing identity(certificate and certificate chain) revocation status using [certificate revocation evaluation](#certificate-revocation-evaluation) section as per `signingIdentityRevocation` setting in trust-policy.
+        1. Perform extended validation using the applicable(if any) plugin.
+        1. If you have reached this step then treat the OCI artifact signature as a valid signature.
+
+### Certificate Revocation Evaluation
+
+If the certificate revocation trust-store setting is set to `skip`, skip the below steps. Otherwise, check for revocation status for certificate and certificate chain.
+
+1. If the revocation status of any of the certificates cannot be determined (revocation unavailable) and `signingIdentityRevocation` is set to either `enforceWithFailClose` or `warn` then log a warning and skip the below steps. Otherwise, fail the signature validation and exit.
+1. If any of the certificates are revoked and `signingIdentityRevocation` is set to either `enforceWithFailOpen` or `enforceWithFailClose` then fail signature validation and exit else log a warning.
+
+Starting from Root to leaf certificate, for each certificate in the certificate chain, perform the following steps to check its revocation status:
+
+- If the certificate being validated doesn't include information OCSP or CRLs then no revocation check is performed and the certificate is considered valid(not revoked).
+- If the certificate being validated includes either OCSP or CRL information, then the one which is present is used for revocation check.
+- If both OCSP URLs and CDP URLs are present, then OCSP is preferred over CRLs. If revocation status cannot be determined using OCSP because of any reason such as unavailability then fallback to using CRLs for revocation check.
+
+#### CRLs
+
+There are two types of CRLs (per RFC 3280), Base CRLs and Delta CRls.
+
+- **Base CRLs:** Contains the revocation status of all certificates that have been issued by a given CA. BaseCRLs are signed by the certificate issuing CAs.
+- **Delta CRLs:** Contains only certificates that have changed status since the last base CRL was published. Delta CRLs are signed by the certificate issuing CAs.
+- **Indirect CRLs:** Special Case in which BaseCRLs and Delta CRLs are not signed by the certificate issuing CAs, instead they are signed with a completely different certificate.
+
+Notary v2 MUST support BaseCRLs and Delta CRLs. Notary v2 MAY support Indirect CRLs. Notary v2 supports only HTTP CRL URLs.
+
+Implementations MAY add support for caching CRLs and OCSP response to improve availability, latency and avoid network overhead.
+
+##### CRL Download
+
+CRL download location (URL) can be obtained from the certificate's CRL Distribution Point (CDP) extension. If the certificate contains multiple CDP locations then each location download is attempted in sequential order. For each CDP location, Notary V2 will try to download the CRL for the default threshold of 10 seconds. If the CRL cannot be downloaded within the timeout threshold the revocation result will be "revocation unavailable". The user may be able to configure this threshold.
+
+##### Revocation Checking with CRL
+
+To check the revocation status of a certificate against CRL, the following steps must be performed:
+
+1. Verify the CRL signature.
+1. Verify that the CRL is valid (not expired). A CRL is considered expired if the current date is after the `NextUpdate` field in the CRL.
+1. Look up the certificate’s serial number in the CRL.
+    1. If the certificate’s serial number is listed in the CRL, look for `InvalidityDate`. If the invalidity date is present and timestamp signature is also present then if the invalidity date is before the timestamping date, the certificate is considered revoked. If the invalidity date is not present in CRL, the certificate is considered revoked.
+    1. If the CRL is expired and the certificate is listed in the CRL for any reason other than `certificate hold`, the certificate is considered revoked.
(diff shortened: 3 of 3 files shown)


signing-and-verification-workflow.md
```diff
@@ -0,0 +1,52 @@
+# Signing and Verification Workflow
+
+This document describes how Notary v2 signs and verifies OCI artifacts.
+
+## Signing workflow
+
+The user wants to sign a locally available OCI artifact and push it to a repository(along with signature).
```
SteveLasker: I think we can remove "locally" as this is one option. I realize this has evolved a bit from the original requirements.
For example, the artifact can be pushed to the registry, and a signature is subsequently pushed to the registry. 
The sign local is an option, but not required.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Prerequisite
+
+- User has access to the signing certificate and private key.
```
SteveLasker: I've seen concern with the term "access". The signing cert may be in a remote, secured location, the user "has access", but can't take the private certificate.

Suggestion:
- User has access to the signing certificate private key or a remote signing service through a notation plug-in.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Steps
+
+1. **Generate signature:** Using notation CLI or any other compliant signing tool, sign the OCI artifact. The signing tool should follow the following guideline.
+    1. Verify that the signing certificate is valid and satisfies [certificate requirements](./signature-specification.md#certificate-requirements).
```
SteveLasker, line 16: While I agree in concept, is this a requirement of the notation client? Is the notation client expected to "verify..."? Or, are we saying

> The signing certificate MUST satisfy the [certificate requirements](./signature-specification.md#certificate-requirements).

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,52 @@
...
+    1. Verify that the signing algorithm satisfies [algorithm requirements](./signature-specification.md#signature-algorithm-requirements).
```
SteveLasker, line 17: > The signing algorithm MUST implement the Notary v2 [algorithm requirements](./signature-specification.md#signature-algorithm-requirements).

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,52 @@
...
+    1. Generate signature.
+        1. Generate signature using signature formats specified in [supported signature envelopes](./signature-specification.md#supported-signature-envelopes).
+        1. Obtain an [RFC-3161](https://datatracker.ietf.org/doc/html/rfc3161.html) compliant timestamp for the signature generated in the previous step. This step is optional but recommended.
+            1. Verify that the timestamp signing certificate satisfies [certificate requirements](./signature-specification.md#certificate-requirements).
```
SteveLasker, line 21: > If provided, the timestamp signing certificate MUST satisfy [certificate requirements](./signature-specification.md#certificate-requirements).

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,52 @@
...
+            1. Verify that the timestamp signing algorithm satisfies [algorithm requirements](./signature-specification.md#signature-algorithm-requirements).
```
SteveLasker, line 22: > The timestamp signing algorithm MUST satisfy the Notary v2 [algorithm requirements](./signature-specification.md#signature-algorithm-requirements).

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,52 @@
...
+        1. Embed timestamp to the signature envelope.
+1. **Push the signature envelope:** Push the signature envelope generated in the previous step to the repository.
+1. **Generate signature artifact manifest:** As decribed in [signature specification](./signature-specification.md#storage) create the Notary v2 signature artifact manifest for the signature envelope generated in step 1.
+1. **Push signature artifact manifest:** Push Notary v2 signature artifact manifest to the repository.
+
+The user can push the OCI artifact to the repository before, after, or in between the signature generation process.
```
SteveLasker: As a signature is pushed, it includes a `subject` property of the artifact it's signing. That artifact must exist in the registry for the push to succeed. See: [ORAS Artifact Push Validation](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md#push-validation)

Suggested text:
> The user pushes the OCI artifact to the repository before the signature generation process as the signature reference must exist for the signature push to succeed. See: [ORAS Artifact Push Validation](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md#push-validation) for more info.

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
 
-1. Get the signing algorithm(hash+encryption) from the signing identity and validate that the signing algorithm is valid and allow-listed.
-1. Get the public key from the signing identity and validate the artifact integrity using the public key and signing algorithm identified in the previous step.
-1. Get and validate TrustStore and TrustPolicy for correctness.
-1. Get the signing identity from the signed artifact and validate it against the identities configured in the trust store. The signing identity should match or leads to at least one of the trusted identities configured in the trust store.
-    1. If signing identity is certificate then validate that certificate and certificate-chain leads to self-signed root.
-1. Perform [artifact expiry](#artifact-expiry) validations based on trust policy.
-1. Perform [artifact revocation](#artifact-revocation) validations based on trust policy
-1. Perform extended validation(If any).
+- User has configured trust-store and trust-policy.
```
SteveLasker: Can we add links to the trust store and policy specs?

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
 
-Here is high level uml diagram for signature evaluation:
+### Steps
 
-![uml diagram showing signature evaluation](./media/trust-store-trust-policy-evaluation.svg)
```
SteveLasker, line 198: Visuals are always helpful. Perhaps a slightly simplified version, but it would be great to bring a simplified version back.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Verification workflow
+
+The user wants to pull OCI artifacts only if they are signed by the trusted publisher and the signature is valid.
```
SteveLasker: Nit:

> The user wants to pull an OCI artifact only if they are signed by a trusted publisher and the signature is valid.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Prerequisites
+
+- User has a digest of OCI artifact they want to download. If the user wants to use a tag to identify the artifact then that tag must be resolved to digest.
```
SteveLasker: We support tag references, which are converted to digests. We don't want to force the user to do the conversion themselves.
Suggestion:
> User has a fully qualified reference to an OCI artifact they want to pull. If a tag is provided, the tag is looked up and converted to a digest.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,52 @@
...
+- User has configured [trust store and trust policy](./trust-store-trust-policy-specification.md) required for signature verification.
+
+### Steps
+
+1. **Get signature artifact descriptors:** Using the ORAS [Manifest Referrers API](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md) download  Notary v2 signature artifact descriptors. The ORAS referrers API only needs artifact digest and artifact type(optional) as input. The Notary v2 signature's artifact type is `application/vnd.cncf.notary.v2.signature`.
```
SteveLasker: Nit:
> **Get signature artifact descriptors:** Using the ORAS [Manifest Referrers API](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md) download the Notary v2 signature artifact descriptors. The ORAS referrers API only needs the artifact digest and `artifactType` (optional) as input. The Notary v2 signature's artifact type is `application/vnd.cncf.notary.v2.signature`.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,52 @@
...
+1. **Get signature artifact manifests:** For each signature artifact descriptor, perform the following steps:
+    1. **Get signature artifact manifests:** Download the Notary v2 signature's artifact manifests for the given artifact descriptor.
+    1. **Filter signature artifact manifests:**
+        1. Based on the value of the `blobs->mediaType` attribute and supported signature envelope format, filter out signature manifests.
```
SteveLasker: I'm not quite following the step here.
Once the user has the manifest, they download the blob for the signature payload. Today, we only have one format, so I'm not sure we need to call out filtering as we haven't identified if we'll have multiple blobs in a single signature or multiple signatures for different formats.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,52 @@
...
+        1. Depending upon the trust-store configuration, further filter out signature manifests. why? How? **TBD**(See [signature filtering](https://hackmd.io/@pritesh/signature_filtering)).
```
SteveLasker: I think this is referring to picking the right signature, based on the policy.
Example: the `net-monitor:v1` image has a wabbit-networks, docker hub and acme-rockets signature. Which one(s) are required for validation, based on the trust store policy.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,52 @@
...
+    1. **Get and validate signatures:** On filtered Notary v2 signature artifact manifest, perform the following steps:
+        1. Download the signature envelope.
+        1. Validate the signature envelope using trust-store and trust-policy as mentioned in [signature evaluation](./trust-store-trust-policy-specification.md#signature-evaluation) section.
+        1. If signature validation fails, skip the below steps and move to the next signature artifact descriptor(step 2.1). If all signature artifact manifests have already been processed, fail the signature verification and exit.
```
SteveLasker: What constitutes a signature validation success/failure? Is this where we should call out checking for revocation lists, and/or timestamp validations?

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+    1. Validate that the content type indicated by the `cty` property value of protected headers in the signature envelope is supported.
+1. **Validate the signature envelope integrity.**
+    1. Get the signing algorithm(hash+encryption) from the signing identity(such as a certificate) and validate that signing algorithm satisfies [algorithm requirements](./signature-specification.md#signature-algorithm-requirements)
+    1. Get the public key from the signing identity and validate the signature envelope integrity using the public key and signing algorithm identified in the previous step.
+1. **Validate the signature against trust policy and trust store.**
+    1. Using the `scope` configured in trust policies, get the applicable trust policy. (Implementations might have this value precomputed, added it for completeness)
+    1. For the applicable trust policy, **validate trust-store:**
+        1. Validate that the signing certificate and certificate chain found in the signature envelope leads to a self-signed root and satisfies [certificate requirements](./signature-specification.md#certificate-requirements).
```
SteveLasker: > leads to a self-signed root
? Do we need to mention self-signed? or can we say, "leads to a root certificate."

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+            1. Validate that the timestamp hash in `TSTInfo.messageImprint` matches the hash of the signature to which the timestamp was applied.
+            1. Validate that the timestamp signing certificate satisfies [certificate requirements](./signature-specification.md#certificate-requirements).
+            1. Validate that the timestamp signing algorithm satisfies [algorithm requirements](./signature-specification.md#signature-algorithm-requirements).
+            1. Validate the `signing-certificate`([RFC-2634](https://tools.ietf.org/html/rfc2634)) or `signing-certificate-v2`([RFC-5126](https://tools.ietf.org/html/rfc5126#section-5.7.3.2)) attribute.
+            1. Check whether timestamping certificate and certificate chain are valid(not expired) or not. If timestamping certificate and certificate-chain are not expired, continue to the next step. Otherwise, if `timestampExpiry` in trust-policy is configured to `Enforce` then fail the signature validation and exit, else log a warning and continue to the next step.
```
SteveLasker: Nit:
1. Check whether timestamping certificate and certificate chain are valid (not expired) or not. If timestamping certificate and certificate-chain are not expired, continue to the next step. Otherwise, if `timestampExpiry` in the trust-policy is configured to `enforce` then fail the signature validation and exit, else log a warning and continue to the next step.

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+        1. Validate signing identity(certificate and certificate chain) revocation status using [certificate revocation evaluation](#certificate-revocation-evaluation) section as per `signingIdentityRevocation` setting in trust-policy.
+        1. Perform extended validation using the applicable(if any) plugin.
+        1. If you have reached this step then treat the OCI artifact signature as a valid signature.
+
+### Certificate Revocation Evaluation
+
+If the certificate revocation trust-store setting is set to `skip`, skip the below steps. Otherwise, check for revocation status for certificate and certificate chain.
+
+1. If the revocation status of any of the certificates cannot be determined(revocation unavailable) and  `signingIdentityRevocation` is set to either `enforceWithFailClose` or `warn` then log a warning and skip the below steps. Otherwise, fail the signature validation and exit.
```
SteveLasker: Nit: 
> If the revocation status of any of the certificates cannot be determined (revocation unavailable) and  `signingIdentityRevocation` is set to either `enforceWithFailClose` or `warn` then log a warning and skip to the below steps. Otherwise, fail the signature validation and exit.

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+
+Starting from Root to leaf certificate, for each certificate in the certificate chain, perform the following steps to check its revocation status:
+
+- If the certificate being validated doesn't include information OCSP or CRLs then no revocation check is performed and the certificate is considered valid(not revoked).
+- If the certificate being validated includes either OCSP or CRL information, then the one which is present is used for revocation check.
+- If both OCSP URLs and CDP URLs are present, then OCSP is preferred over CRLs. If revocation status cannot be determined using OCSP because of any reason such as unavailability then fallback to using CRLs for revocation check.
+
+#### CRLs
+
+There are two types of CRLs(per RFC 3280), Base CRLs and Delta CRls.
```
SteveLasker: Nit:
There are two types of CRLs (per RFC 3280), Base CRLs and Delta CRls.

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+
+- **Base CRLs:** Contains the revocation status of all certificates that have been issued by a given CA. BaseCRLs are signed by the certificate issuing CAs.
+- **Delta CRLs:** Contains only certificates that have changed status since the last base CRL was published. Delta CRLs are signed by the certificate issuing CAs.
+- **Indirect CRLs:** Special Case in which BaseCRLs and Delta CRLs are not signed by the certificate issuing CAs, instead they are signed with a completely different certificate.
+
+Notary v2 MUST support BaseCRLs and Delta CRLs. Notary v2 MAY support Indirect CRLs. Notary v2 supports only HTTP CRL URLs.
```
SteveLasker, line 244: Just a note: 
There are lots of ideas around how to improve the CRL models, but I'll leave those for later and we'll lean into the existing infra and work to improve it as incremental work.

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+
+Implementations MAY add support for caching CRLs and OCSP response to improve availability, latency and avoid network overhead.
+
+##### CRL Download
+
+CRL download location(URL) can be obtained from the certificate's CRL Distribution Point(CDP) extension. If the certificate contains multiple CDP locations then each location download is attempted in sequential order. For each CDP location, NotaryV2 will try to download the CRL for the default threshold of 10 seconds. If the CRL cannot be downloaded within the timeout threshold the revocation result will be "revocation unavailable". The user may be able to configure this threshold.
```
SteveLasker: Nit:
> CRL download location (URL) can be obtained from the certificate's CRL Distribution Point (CDP) extension.

> For each CDP location, Notary V2 will try to download the CRL for the default threshold of 10 seconds.

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+
+##### Revocation Checking with CRL
+
+To check the revocation status of a certificate against CRL, the following steps must be performed:
+
+1. Verify the CRL signature.
+1. Verify that the CRL is valid(not expired). A CRL is considered expired if the current date is after the `NextUpdate` field in the CRL.
```
SteveLasker: Nit `space`()

> Verify that the CRL is valid (not expired).

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+
+- If the delta CRL cannot be retrieved for some reason, the revocation result will be "revocation unavailable".
+- If the delta CRL’s indicator is less than the current base CRL, the revocation result returned will be "revocation unavailable"
+- To check revocation of a certificate in Delta CRLs follow steps similar to [Revocation Checking with CRL](#revocation-checking-with-crl).
+
+#### OCSP
+
+##### OCSP Download
+
+OCSP URLs can be obtained from the certificate's authority information access(AIA) extension as defined in [RFC-2560](https://datatracker.ietf.org/doc/html/rfc2560). Notary V2 will wait for a default threshold of 5 seconds to receive an OCSP response. If OCSP response is not available within the timeout threshold the revocation result will be "revocation unavailable". The user may be able to configure this threshold.
```
SteveLasker: Space nit:
> Verify that the CRL is valid (not expired).

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+
+##### Revocation Checking with OCSP
+
+To check the revocation status of a certificate using OCSP, the following steps must be performed.
+
+1. Verify the signature of the OCSP response. This step also includes verifying the OSCP response signing certificate is valid and trusted. The OCSP signing certificate must be issued by the same CA as the certificate being verified or the OCSP response must be signed by the issuing CA.
+1. Verify that the OCSP response is valid(not expired). A CRL is considered expired if the current date is after the `NextUpdate` field in the CRL.
```
SteveLasker: Nit: 
> 1. Verify that the OCSP response is valid (not expired).

SteveLasker (commented):
Thanks @priteshbandi for all the details. Mostly nits, with a bit more details in some places.

trust-store-trust-policy-specification.md
priteshbandi: Nope, roots are always self-signed. Will update

signing-and-verification-workflow.md
priteshbandi: updated to remove locally.

trust-store-trust-policy-specification.md
priteshbandi: Yep, visuals are helpful but in this case it was becoming complicated so had to remove. I can try to create simplified version of it.

signing-and-verification-workflow.md
priteshbandi: Added it for backward compatibility so that in the future if we have multiple signature formats we don't break existing clients.

signing-and-verification-workflow.md
priteshbandi: Yep, that's correct.

signing-and-verification-workflow.md
priteshbandi: That's specified in the previous step and the hyperlink points to the signature evaluation process.

> Validate the signature envelope using trust-store and trust-policy as mentioned in [signature evaluation](./trust-store-trust-policy-specification.md#signature-evaluation) section

signing-and-verification-workflow.md
priteshbandi: This document depicts all the steps required for signature generation and verification. Also, the compliant client MUST perform these steps. 

I am not sure how to combine workflow and specification/requirement so wrote the complete doc as workflow/control-flow document. Also, its easier to read and follow workflow than specification.

signature-specification.md
```diff
@@ -230,6 +230,22 @@ The signing certificate's public key algorithm and size MUST be used to determin
 | EC                   | 384             | ECDSA on secp384r1 with SHA-384 |
 | EC                   | 512             | ECDSA on secp521r1 with SHA-512 |
 
+### Certificate Requirements
+
+The signing certificate MUST meet the following minimum requirements:
+
+- The certificate MUST be valid for the id-kp-codeSigning purpose([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12)).
```
gokarnm: I'd also clarify values for `keyUsage digitalSignature` and `basicConstraints CA`.

signature-specification.md
```diff
@@ -230,6 +230,22 @@ The signing certificate's public key algorithm and size MUST be used to determin
...
+- The certificate MUST abide by the following key length restrictions:
+  - For RSA public key, the key length MUST be 2048 bits or higher.
+  - For ECDSA public key, the key length MUST be 256 bits or higher.
+
+The timestamping certificate MUST meet the following minimum requirements:
+
+- The certificate MUST be valid for the id-kp-timeStamping purpose([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12)).
```
gokarnm: Same as previous comment about additional cert attributes to validate.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,53 @@
...
+- User has access to the signing certificate and private key or a remote signing service through a notation plug-in.
+
+### Steps
+
+1. **Generate signature:** Using notation CLI or any other compliant signing tool, sign the OCI artifact. The signing tool should follow the following guideline.
+    1. Verify that the signing certificate is valid and satisfies [certificate requirements](./signature-specification.md#certificate-requirements).
+    1. Verify that the signing algorithm satisfies [algorithm requirements](./signature-specification.md#signature-algorithm-requirements).
+    1. Generate signature.
+        1. Generate signature using signature formats specified in [supported signature envelopes](./signature-specification.md#supported-signature-envelopes).
+        1. Obtain an [RFC-3161](https://datatracker.ietf.org/doc/html/rfc3161.html) compliant timestamp for the signature generated in the previous step. This step is optional but recommended.
```
gokarnm: This should be based on if user configured use of TSA. One way would be to add additional config options in Notation key definition to enable TSA and specify TSA server.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,53 @@
...
+
+The user pushes the OCI artifact to the repository before the signature generation process as the signature reference must exist for the signature push to succeed. See: [ORAS Artifact Push Validation](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md#push-validation) for more info.
+
+## Verification workflow
+
+The user wants to pull an OCI artifact only if they are signed by a trusted publisher and the signature is valid.
+
+### Prerequisites
+
+- User has a fully qualified reference to an OCI artifact they want to pull. If a tag is provided, the tag is looked up and converted to a digest.
```
gokarnm: Notation will accept a url with tag and construct the descriptor from downloaded manifest, the registry maps tag to digest and send the artifact manifest as response.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,53 @@
...
+- User has configured [trust store and trust policy](./trust-store-trust-policy-specification.md) required for signature verification.
+
+### Steps
+
+1. **Get signature artifact descriptors:** Using the ORAS [Manifest Referrers API](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md) download the Notary v2 signature artifact descriptors. The ORAS referrers API only needs the artifact digest and `artifactType`(optional) as input. The Notary v2 signature's artifact type is `application/vnd.cncf.notary.v2.signature`.
```
gokarnm: I'd reword to say that Notation will call referrers API with digest and artifactType as `application/vnd.cncf.notary.v2.signature`, we need describe the referrers api which is already linked.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,53 @@
...
+1. **Get signature artifact manifests:** For each signature artifact descriptor, perform the following steps:
+    1. **Get signature artifact manifests:** Download the Notary v2 signature's artifact manifests for the given artifact descriptor.
+    1. **Filter signature artifact manifests:**
```
gokarnm: s/manifests/manifest ?

signing-and-verification-workflow.md
gokarnm: Can you elaborate in the in the workflow about why this is done.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,53 @@
...
+        1. Based on the value of the `blobs->mediaType` attribute and supported signature envelope format, filter out signature manifests.
+        1. Depending upon the trust-store configuration, further filter out signature manifests. why? How? **TBD**(See [signature filtering](https://hackmd.io/@pritesh/signature_filtering)).
+        1. If the artifact manifest is filtered out, skip the below steps and move to the next signature artifact descriptor(step 2.1). If all signature artifact descriptors have already been processed, fail the signature verification and exit.
+    1. **Get and validate signatures:** On filtered Notary v2 signature artifact manifest, perform the following steps:
```
gokarnm: Nit, s/On filtered/On the filtered
s/validate/verify - for this whole section.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,53 @@
...
+        1. Download the signature envelope.
+        1. Validate the signature envelope using trust-store and trust-policy as mentioned in [signature evaluation](./trust-store-trust-policy-specification.md#signature-evaluation) section.
+        1. If signature validation fails, skip the below steps and move to the next signature artifact descriptor(step 2.1). If all signature artifact manifests have already been processed, fail the signature verification and exit.
+        1. If signature validation succeeds, compare the given digest with the signed digest present in the signature envelope's payload. If digests are equal, signature verification is considered successful. Otherwise, move to the next signature artifact descriptor(step 2.1). If all signature artifact descriptors have already been processed, fail the signature verification and exit.
```
gokarnm: It's unclear what "given digest" mean, can be more descriptive here.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,53 @@
...
+1. **Get OCI artifact:** Using the given digest, download the OCI artifact. This step is not in the purview of Notary v2.
```
gokarnm: s/given/verified

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
-1. Perform [artifact revocation](#artifact-revocation) validations based on trust policy
-1. Perform extended validation(If any).
+- User has configured [trust store](#trust-store) and [trust policy](#trust-policy).
 
-Here is high level uml diagram for signature evaluation:
+### Steps
 
-![uml diagram showing signature evaluation](./media/trust-store-trust-policy-evaluation.svg)
+1. **Validate that the signature envelope format is supported.**
+    1. Parse the signature envelope content based on the signature envelope type specified in the signature artifact manifest.
```
gokarnm: Are you referring to the signature blob media type here?

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+    1. Validate that the content type indicated by the `cty` property value of protected headers in the signature envelope is supported.
+1. **Validate the signature envelope integrity.**
+    1. Get the signing algorithm(hash+encryption) from the signing identity(such as a certificate) and validate that signing algorithm satisfies [algorithm requirements](./signature-specification.md#signature-algorithm-requirements)
```
gokarnm: Unclear which certificate is used , is this from the envelope cert chain?

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+    1. Get the public key from the signing identity and validate the signature envelope integrity using the public key and signing algorithm identified in the previous step.
+1. **Validate the signature against trust policy and trust store.**
+    1. Using the `scope` configured in trust policies, get the applicable trust policy. (Implementations might have this value precomputed, added it for completeness)
+    1. For the applicable trust policy, **validate trust-store:**
+        1. Validate that the signing certificate and certificate chain found in the signature envelope leads to a root certificate and satisfies [certificate requirements](./signature-specification.md#certificate-requirements).
+        1. For all the trust-stores configured in applicable trust-policy perform the following steps.
+            1. Validate that certificate and certificate-chain lead to a trusted certificate configured in the `x509Certs` field of trust-store.
+            1. If above verification succeeds then continue to the next step else iterate over the next set of trust stores. If all of the trust stores have been evaluated then fail the signature validation and exit.
```
gokarnm: "iterate over the next set of trust store" => "next trust store" or "next set of root in trust store"

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+    1. **Validate trust policy:**
+        1. If signature expiry is present in the signature envelope, using the local machine’s current time(in UTC) check whether the signature is expired or not. If the signature is not expired, continue to the next step. Otherwise, if `signatureExpiry` is set to `Enforce` then fail the signature validation and exit else log a warning and continue to the next step.
+        1. Check for the timestamp in the signature envelope.
```
gokarnm: s/timestamp/timestamp signature?

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+            1. If the timestamp exists, continue with the next step. Otherwise, store the local machine's current time(in  UTC) in variables `timeStampLowerLimit` and `timeStampUpperLimit` and continue with step 3.3.3.
+            1. Validate that the timestamp hash in `TSTInfo.messageImprint` matches the hash of the signature to which the timestamp was applied.
+            1. Validate that the timestamp signing certificate satisfies [certificate requirements](./signature-specification.md#certificate-requirements).
+            1. Validate that the timestamp signing algorithm satisfies [algorithm requirements](./signature-specification.md#signature-algorithm-requirements).
+            1. Validate the `signing-certificate`([RFC-2634](https://tools.ietf.org/html/rfc2634)) or `signing-certificate-v2`([RFC-5126](https://tools.ietf.org/html/rfc5126#section-5.7.3.2)) attribute.
```
gokarnm: Is this only applicable to TSA cert?

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+            1. Check whether timestamping certificate and certificate chain are valid (not expired) or not. If timestamping certificate and certificate-chain are not expired, continue to the next step. Otherwise, if `timestampExpiry` in trust-policy is configured to `Enforce` then fail the signature validation and exit, else log a warning and continue to the next step.
```
gokarnm, line 212: Let's discuss if we really need to enforce this.

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,109 @@ The implementation must allow the user to execute custom validations. These cust
...
+            1. Retrieve the timestamp's generalized time from `TSTInfo.genTime`.
+            1. Retrieve the timestamp's accuracy.  If the accuracy is explicitly specified in `TSTInfo.accuracy`, use that value.  If the accuracy is not explicitly specified and `TSTInfo.policy` is the baseline time-stamp policy([RFC-3628](https://tools.ietf.org/html/rfc3628#section-5.2)), use accuracy of 1 second.  Otherwise, use accuracy of 0.
+            1. Calculate the timestamp range using the lower and upper limits per [RFC-3161 section 2.4.2](https://tools.ietf.org/html/rfc3161#section-2.4.2) and store the limits in variables `timeStampLowerLimit` and `timeStampUpperLimit`, respectively.
+        1. Check that the time range from `timeStampLowerLimit` to `timeStampUpperLimit` timestamp is entirely within the certificate's validity period.  If the time range is entirely within the signing certificate and certificate chain's validity period, continue to the next step.  Otherwise, If `signingIdentityExpiry` in trust-policy is configured to `Enforce` then fail the signature validation and exit else log a warning and continue to the next step.
```
gokarnm, line 218: Would like to understand this better, and if we should apply `signingIdentityExpiry` config here. Also, I think we should TSA verification details into another section and just summarize here.

gokarnm (changes requested):

signing-and-verification-workflow.md
priteshbandi: Updated.

trust-store-trust-policy-specification.md
priteshbandi: yep, updated sentence to reflect that.

trust-store-trust-policy-specification.md
priteshbandi: clarified that its signing certificate.

signing-and-verification-workflow.md
gokarnm:
```suggestion
1. **Get signature artifact descriptors:** Using the ORAS [Manifest Referrers API](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md) download the Notary v2 signature artifact descriptors. The  `artifactType` parameter is set to Notary v2 signature's artifact type `application/vnd.cncf.notary.v2.signature`. Note that all ORAS implementations may not support filtering using `artifactType`.
```

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,53 @@
...
+    1. **Get and verify signatures:** On the filtered Notary v2 signature artifact manifest, perform the following steps:
+        1. Download the signature envelope.
+        1. Validate the signature envelope using trust-store and trust-policy as mentioned in [signature evaluation](./trust-store-trust-policy-specification.md#signature-evaluation) section.
+        1. If signature validation fails, skip the below steps and move to the next signature artifact descriptor(step 2.1). If all signature artifact manifests have already been processed, fail the signature verification and exit.
+        1. If signature validation succeeds, compare the digest derived from the given OCI artifact reference with the signed digest present in the signature envelope's payload. If digests are equal, signature verification is considered successful. Otherwise, move to the next signature artifact descriptor(step 2.1). If all signature artifact descriptors have already been processed, fail the signature verification and exit.
+1. **Get OCI artifact:** Using the verified digest, download the OCI artifact. This step is not in the purview of Notary v2.
```
shizhMSFT, line 56: The signed payload of a signature contains the descriptor of the target OCI artifact, which includes digest, size, and media type.

Instead of `using the verified digest`, we should say `using the verified descriptor`.

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,53 @@
...
+            1. Verify that the timestamp signing certificate satisfies [certificate requirements](./signature-specification.md#certificate-requirements).
+            1. Verify that the timestamp signing algorithm satisfies [algorithm requirements](./signature-specification.md#signature-algorithm-requirements).
+        1. Embed timestamp to the signature envelope.
+1. **Push the signature envelope:** Push the signature envelope generated in the previous step to the repository.
+1. **Generate signature artifact manifest:** As decribed in [signature specification](./signature-specification.md#storage) create the Notary v2 signature artifact manifest for the signature envelope generated in step 1.
+1. **Push signature artifact manifest:** Push Notary v2 signature artifact manifest to the repository.
+
+The user pushes the OCI artifact to the repository before the signature generation process as the signature reference must exist for the signature push to succeed. See: [ORAS Artifact Push Validation](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md#push-validation) for more info.
+
+## Verification workflow
```
shizhMSFT, line 30: If `skipSignatureVerification` in the corresponding trust policy is set to `true`, do we need to do the verification workflow at all?

priteshbandi: No, we don't need to. I was waiting for scoping PR to get merged but as we have LGTM on the scoping PR will update this PR to account for `skipSignatureVerification`

signing-and-verification-workflow.md
priteshbandi: Have mentioned digest instead of descriptor because we haven't verified the descriptor inside the signature envelope with the actual OCI artifact's descriptor (the one stored in the registry).

signing-and-verification-workflow.md
```diff
@@ -0,0 +1,54 @@
...
+1. **Get signature artifact descriptors:** Using the ORAS [Manifest Referrers API](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md) download the Notary v2 signature artifact descriptors. The  `artifactType` parameter is set to Notary v2 signature's artifact type `application/vnd.cncf.notary.v2.signature`. Note that all ORAS implementations may not support filtering using `artifactType`.
+1. For each signature artifact descriptor, perform the following steps:
+    1. **Get signature artifact manifest:** Download the Notary v2 signature's artifact manifest for the given artifact descriptor.
+    1. **Filter signature artifact manifest:**
+        1. Filter out the unsupported signature envelope format by comparing the value of the `blobs->mediaType` attribute of the signature manifest with the supported formats defined in [signature specification](./signature-specification.md#storage).
```
gokarnm: https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md 

```suggestion
        1. Filter out the unsupported signature formats by comparing the signature envelope format type (`[descriptors].descriptor.mediaType`) in the signature manifest, with the supported formats defined in [signature specification](./signature-specification.md#storage).
```

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,110 @@ The implementation must allow the user to execute custom validations. These cust
...
-1. Perform [artifact revocation](#artifact-revocation) validations based on trust policy
-1. Perform extended validation(If any).
+- User has configured [trust store](#trust-store) and [trust policy](#trust-policy).
 
-Here is high level uml diagram for signature evaluation:
+### Steps
 
-![uml diagram showing signature evaluation](./media/trust-store-trust-policy-evaluation.svg)
+1. **Validate that the signature envelope format is supported.**
+    1. Parse the signature envelope content based on the signature envelope type specified in the `blobs->mediaType` attribute of the signature artifact manifest.
```
gokarnm:
```suggestion
    1. Parse the signature envelope content based on the signature envelope type specified in the `[descriptors].descriptor.mediaType` attribute of the signature artifact manifest.
```

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,110 @@ The implementation must allow the user to execute custom validations. These cust
...
+1. **Validate the signature envelope integrity.**
+    1. Get the signing certificate from the parsed [signature envelope](https://github.com/notaryproject/notaryproject/blob/7b7d283038/signature-specification.md#signature-envelope).
+    1. Get the signing algorithm(hash+encryption) from the signing certificate and validate that the signing algorithm satisfies [algorithm requirements](./signature-specification.md#signature-algorithm-requirements)
+    1. Using the public key of the signing certificate and signing algorithm identified in the previous step, validate the integrity of the signature envelope.
+1. **Validate the signature against trust policy and trust store.**
+    1. Using the `scope` configured in trust policies, get the applicable trust policy. (Implementations might have this value precomputed, added it for completeness)
+    1. For the applicable trust policy, **validate trust-store:**
+        1. Validate that the signing certificate and certificate chain found in the signature envelope leads to a root certificate and satisfies [certificate requirements](./signature-specification.md#certificate-requirements).
```
gokarnm: Can we reword, I confused this with cert chain validation against TS.

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,110 @@ The implementation must allow the user to execute custom validations. These cust
...
+        1. For all the trust-stores configured in applicable trust-policy perform the following steps.
```
gokarnm:
```suggestion
        1. For each the trust-stores configured in applicable trust-policy perform the following steps.
```

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,110 @@ The implementation must allow the user to execute custom validations. These cust
...
+            1. Validate that certificate and certificate-chain lead to a trusted certificate configured in the `x509Certs` field of trust-store.
+            1. If the above verification succeeds then continue to the next step else iterate over the next trust store. If all of the trust stores have been evaluated then fail the signature validation and exit.
+    1. **Validate trust policy:**
+        1. If signature expiry is present in the signature envelope, using the local machine’s current time(in UTC) check whether the signature is expired or not. If the signature is not expired, continue to the next step. Otherwise, if `signatureExpiry` is set to `Enforce` then fail the signature validation and exit else log a warning and continue to the next step.
+        1. Check for the timestamp signature in the signature envelope.
+            1. If the timestamp exists, continue with the next step. Otherwise, store the local machine's current time(in  UTC) in variables `timeStampLowerLimit` and `timeStampUpperLimit` and continue with step 3.3.3.
```
gokarnm:
```suggestion
            1. If the timestamp exists, continue with the next step. Otherwise, store the local machine's current time(in  UTC) in variables `timeStampLowerLimit` and `timeStampUpperLimit` and continue with step 3.3.c.
```

gokarnm (changes requested):

trust-store-trust-policy-specification.md
```diff
@@ -141,20 +141,110 @@ The implementation must allow the user to execute custom validations. These cust
...
+        1. Validate signing identity(certificate and certificate chain) revocation status using [certificate revocation evaluation](#certificate-revocation-evaluation) section as per `signingIdentityRevocation` setting in trust-policy.
+        1. Perform extended validation using the applicable(if any) plugin.
+        1. If you have reached this step then treat the OCI artifact signature as a valid signature.
+
+### Certificate Revocation Evaluation
+
+If the certificate revocation trust-store setting is set to `skip`, skip the below steps. Otherwise, check for revocation status for certificate and certificate chain.
+
+1. If the revocation status of any of the certificates cannot be determined (revocation unavailable) and `signingIdentityRevocation` is set to either `enforceWithFailClose` or `warn` then log a warning and skip the below steps. Otherwise, fail the signature validation and exit.
```
sudo-bmitch: Should this be `enforceWithFailOpen`?

sudo-bmitch (commented):
We may want to make a clear separation between the policies that will likely be implemented in notation, from the process to verify a signature what would also be implemented in other tools that want to verify a notary signature with their own policy checks.

signing-and-verification-workflow.md
priteshbandi: Reworded to `1. If the user wants to timestamp the signature, obtain an [RFC-3161](https://datatracker.ietf.org/doc/html/rfc3161.html) compliant timestamp for the signature generated in the previous step. Otherwise, continue to the next step.`

signing-and-verification-workflow.md
priteshbandi: Because of security concerns around the value of resolved tag being different during signature verification and actual pull, for this PR we won't support automatic conversion of the tag to digest inside notary.

trust-store-trust-policy-specification.md
priteshbandi: No, it should be `enforceWithFailClose`,


- **`enforceWithFailOpen`**: This means implementation MUST perform validation and if validation fails because the endpoint is not reachable, the implementation MUST throw an error and MUST fail the validation.
- **`enforceWithFailClose`**: This means implementation MUST perform validation and if validation fails because the endpoint is not reachable, the implementation MUST log an error and MUST NOT fail the validation.

gokarnm:
LGTM!

trust-store-trust-policy-specification.md
sudo-bmitch: Perhaps I'm misunderstanding, or our usage of the terminology is confusing here. The traditional usage of "fail open" means that you only throw a warning but do not stop workflows from running when the security check is unable to be performed. It sounds like we're saying "fail open" means to fail the validation which would block images from being deployed in the common use case.

priteshbandi: Yep, you are correct our usage of terminology is incorrect. Recently I've been reading a lot about [circuit breakers](https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern#Different_States_of_Circuit_Breaker) where open state represents failure scenario and close state represents business as usual. Perhaps I got confused and interpolated that with fail open and fail close of networking.
Have fixed it in the latest commit.

signature-specification.md
```diff
@@ -230,6 +230,28 @@ The signing certificate's public key algorithm and size MUST be used to determin
 | EC                   | 384             | ECDSA on secp384r1 with SHA-384 |
 | EC                   | 512             | ECDSA on secp521r1 with SHA-512 |
 
+### Certificate Requirements
+
+The signing certificate MUST meet the following minimum requirements:
+
+- The signing certificate MUST satisfy the following constraints:
+  - The keyUsage extension MUST be present and MUST be marked critical. The bit positions for digitalSignature MUST be set ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3)).
```
SteveLasker: These two lines read as redundant. Can these be merged or differentiated if needed?

signature-specification.md
```diff
@@ -230,6 +230,28 @@ The signing certificate's public key algorithm and size MUST be used to determin
...
+  - The extKeyUsage extension MUST be present and its value MUST be id-kp-codeSigning ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12)).
+  - If the basicConstraints extension is present, the cA field MUST be set false ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9)).
+- The certificate MUST abide by the following key length restrictions:
+  - For RSA public key, the key length MUST be 2048 bits or higher.
+  - For ECDSA public key, the key length MUST be 256 bits or higher.
+
+The timestamping certificate MUST meet the following minimum requirements:
+
+- The timestamping certificate MUST satisfy the following constraints:
```
SteveLasker: Same here. This and the line above read as redundant.

SteveLasker (commented):
Overall LGTM, with a bit of redundant text on lines 237-238 and 245-247.

trust-store-trust-policy-specification.md
```diff
@@ -181,21 +181,110 @@ The implementation must allow the user to execute custom validations. These cust
...
+        1. Validate signing identity(certificate and certificate chain) revocation status using [certificate revocation evaluation](#certificate-revocation-evaluation) section as per `signingIdentityRevocation` setting in trust-policy.
+        1. Perform extended validation using the applicable(if any) plugin.
+        1. If you have reached this step then treat the OCI artifact signature as a valid signature.
+
+### Certificate Revocation Evaluation
+
+If the certificate revocation trust-store setting is set to `skip`, skip the below steps. Otherwise, check for revocation status for certificate and certificate chain.
+
+1. If the revocation status of any of the certificates cannot be determined (revocation unavailable) and `signingIdentityRevocation` is set to either `enforceWithFailOpen` or `warn` then log a warning and skip the below steps. Otherwise, fail the signature validation and exit.
+1. If any of the certificates are revoked and `signingIdentityRevocation` is set to either `enforceWithFailOpen` or `enforceWithFailClose` then fail signature validation and exit else log a warning.
```
SteveLasker, line 228: Am I reading this as `enforceWithFailOpen` or `enforceWithFailClose` have the same result?

trust-store-trust-policy-specification.md
```diff
@@ -181,21 +181,110 @@ The implementation must allow the user to execute custom validations. These cust
...
+- **Delta CRLs:** Contains only certificates that have changed status since the last base CRL was published. Delta CRLs are signed by the certificate issuing CAs.
+- **Indirect CRLs:** Special Case in which BaseCRLs and Delta CRLs are not signed by the certificate issuing CAs, instead they are signed with a completely different certificate.
+
+Notary v2 MUST support BaseCRLs and Delta CRLs. Notary v2 MAY support Indirect CRLs. Notary v2 supports only HTTP CRL URLs.
+
+Implementations MAY add support for caching CRLs and OCSP response to improve availability, latency and avoid network overhead.
+
+##### CRL Download
+
+CRL download location (URL) can be obtained from the certificate's CRL Distribution Point (CDP) extension. If the certificate contains multiple CDP locations then each location download is attempted in sequential order. For each CDP location, Notary V2 will try to download the CRL for the default threshold of 10 seconds. If the CRL cannot be downloaded within the timeout threshold the revocation result will be "revocation unavailable". The user may be able to configure this threshold.
```
SteveLasker, line 250: Do we want to leave (`The user may be able to configure this threshold.`) open ended for now? Just seems to lead to what should be a link for how this is done today, or a new issue to track for how the user can configure it.
Or, we can simply remove that last sentence and do a subsequent PR when and how configuration is enabled.

trust-store-trust-policy-specification.md
```diff
@@ -181,21 +181,110 @@ The implementation must allow the user to execute custom validations. These cust
...
+
+##### Revocation Checking with CRL
+
+To check the revocation status of a certificate against CRL, the following steps must be performed:
+
+1. Verify the CRL signature.
+1. Verify that the CRL is valid (not expired). A CRL is considered expired if the current date is after the `NextUpdate` field in the CRL.
+1. Look up the certificate’s serial number in the CRL.
+    1. If the certificate’s serial number is listed in the CRL, look for `InvalidityDate`. If the invalidity date is present and timestamp signature is also present then if the invalidity date is before the timestamping date, the certificate is considered revoked. If the invalidity date is not present in CRL, the certificate is considered revoked.
```
SteveLasker: Is this accurate, or complete?

> If the invalidity date is present and timestamp signature is also present then if the invalidity date is before the timestamping date, the certificate is considered revoked. 

Should this be:

> If the invalidity date is present and timestamp signature is also present then if the invalidity date is before the current date and before the timestamping date, the certificate is considered revoked.

trust-store-trust-policy-specification.md
```diff
@@ -181,21 +181,110 @@ The implementation must allow the user to execute custom validations. These cust
...
+
+- If the delta CRL cannot be retrieved for some reason, the revocation result will be "revocation unavailable".
+- If the delta CRL’s indicator is less than the current base CRL, the revocation result returned will be "revocation unavailable"
+- To check revocation of a certificate in Delta CRLs follow steps similar to [Revocation Checking with CRL](#revocation-checking-with-crl).
+
+#### OCSP
+
+##### OCSP Download
+
+OCSP URLs can be obtained from the certificate's authority information access (AIA) extension as defined in [RFC-2560](https://datatracker.ietf.org/doc/html/rfc2560). Notary V2 will wait for a default threshold of 5 seconds to receive an OCSP response. If OCSP response is not available within the timeout threshold the revocation result will be "revocation unavailable". The user may be able to configure this threshold.
```
SteveLasker, line 277: > The user may be able to configure this threshold.

Same comment as above. If we have a way to configure today, let's add a link. If not, we should just remove this sentence and see if/when we need to configure and make that change as a batch pr.

SteveLasker (commented):
A few more comments on trust-store-trust-policy-specification.md

trust-store-trust-policy-specification.md
priteshbandi: Yep, when the notary can identify the revocation status of a certificate enforceWithFailOpen or enforceWithFailClose will have the same result. The behavior differs when notary is unable to determine certificate revocation status.

trust-store-trust-policy-specification.md
priteshbandi: Yes, keeping it open-ended is intentional. We want the implementations to have some sane defaults and we also want to allow implementations to override these values with user-provided values.

trust-store-trust-policy-specification.md
priteshbandi: So there is no standard here. Basically we are saying that implementation MUST use default threshold of 5 and implementation MAY provide functionality for user to override this values.

trust-store-trust-policy-specification.md
priteshbandi: I have paraphrased the sentence for better readability.
`If CRL has invalidity date and artifact signature is timestamped then compare the invalidity date with the timestamping date. If the invalidity date is before the timestamping date, the certificate is considered revoked.`

Basically we are comparing the signature's timestamping date(signing date) with the certificate's effective revocation date.

signature-specification.md
priteshbandi: Merged the lines.

signature-specification.md
```diff
@@ -230,6 +230,28 @@ The signing certificate's public key algorithm and size MUST be used to determin
 | EC                   | 384             | ECDSA on secp384r1 with SHA-384 |
 | EC                   | 512             | ECDSA on secp521r1 with SHA-512 |
 
+### Certificate Requirements
+
+The signing certificate MUST meet the following minimum requirements:
+
+- The signing certificate MUST satisfy the following constraints:
+  - The keyUsage extension MUST be present and MUST be marked critical. The bit positions for digitalSignature MUST be set ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3)).
+  - The extKeyUsage extension MUST be present and its value MUST be id-kp-codeSigning ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12)).
```
briankr-ms: Recommend removing this requirement, as not all signed artifacts will be "code".

priteshbandi: The idea here is to not allow the use of certificates (like timestamping, TLS) that are not meant for singing, to sign the artifacts.
Do you have a recommendation on how we can achieve this while removing `id-kp-codeSigning`?

---

>Recommend removing this requirement, as not all signed artifacts will be "code".

What kind of signed artifacts won't be a code, can you please share some examples? 

Also found an interesting draft https://www.ietf.org/archive/id/draft-ietf-lamps-documentsigning-eku-00.html

priteshbandi: Opened an issue to track this https://github.com/notaryproject/notaryproject/issues/130

trust-store-trust-policy-specification.md
SteveLasker: Ok, so the notary spec says the user may configure it. 
The current `notation` cli doesn't yet implement it, but the spec says it may.

SteveLasker (approved):
LGTM

PR #124, opened 2022-01-27
Plugin interface for signing. 

gokarnm:
Related issue notaryproject/roadmap/issues/26.

* Includes plugin interfaces for signing workflow
* Addresses feedback on plugin mechanism and versioning
* Added validations performed by Notation for plugin responses
* Added error contract and error codes
* [hackmd doc](https://hackmd.io/jhYbA4aNRFW7nmp7gPTIBA?view)

### DIFF
--- a/specs/plugin-extensibility.md
+++ b/specs/plugin-extensibility.md
@@ -1,0 +1,286 @@
+# Notation Extensibility for Signing and Verification
+
+Keys and associated certificates used for signing artifacts using Notary could be available to users through varied solutions that provide secure key generation, storage and cryptographic operations. Some are well established with standards like [PIV](https://csrc.nist.gov/projects/piv/piv-standards-and-supporting-documentation) and [PKCS #11](https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/os/pkcs11-base-v3.0-os.html) implemented by hardware tokens, smart cards. More recent options which use varied authentication and API protocols are remote key management services and signing services by third party vendors and cloud service providers. Notation will support a few built-in integrations with standard providers, and will provide plugin interfaces for users, and vendors to implement their own integrations with the solutions they use. This allows a plugin publisher to implement, test, release and patch their solutions independent of Notation’s development and release cycle. This document provides specification for the plugin model, and interfaces to implement. This specification aims to work both for [existing](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md) and future signature formats adopted by Notary.
+
+## Terminology
+
+* **Plugin Publisher** - A user, organization, open source project or 3rd party vendor that creates a Notation plugin for internal or public distribution.
+* **Plugin** - A component external to Notation that can integrate as one of the steps in  Notation’s workflow for signature generation or verification.
+* **Default provider** - Signing and verification mechanisms built into Notation itself to provide default experience without requiring to install/configure additional plugins. ***[We are yet to define what is included in the default experience]***.
+
+## Plugin mechanism
+
+### Requirements
+
+* A plugin publisher MUST be able to distribute and patch a Notation plugin independently of Notation’s release cycle.
+* The plugin mechanism MUST work across commonly used OS platforms like Linux, Windows and macOS.
+* The plugin interface contract MUST be versioned. This version is different from Notation's release version. It will be used to accomodate for additional capabilities and contract changes post initial release of Notation.
+* A plugin MAY implement a subset of capabilities (features) available in plugin contract. E.g A plugin may implement signing feature, but not verification.
+* Notation and plugins MAY be updated independently in an environment.
+  * Notation MUST work with a plugin that implements a matching or lower minor version of the plugin contract. Notation SHALL NOT support using a plugin with higher version of plugin contract.
+  * A plugin MUST support a single plugin contract version, per major version.
+
+Notation will invoke plugins as executable, pass parameters using command line arguments, and use standard IO streams to pass request/response payloads. This mechanism is used as Go language (used to develop [Notation library](https://github.com/notaryproject/notation-go-lib)) does not have a [in built support](https://github.com/golang/go/issues/19282) to load and execute plugins that works across OS platforms. Other mechanisms like gRPC require every plugin to be implemented as a service/daemon.
+
+### Plugin installation and config
+
+* Plugin publisher will provide instructions to download and install the plugin. Plugins intended for public distribution should also include instructions for users to verify the authenticity of the plugin.
+* Each plugin executable and dependencies are installed under directory `~/.notation/plugins/{plugin-name}` with an executable under that directory `~/.notation/plugins/{plugin-name}/notation-{plugin-name}`. The executable can be a shim which calls plugin dependecies installed elsewhere on the file system.
+* To use a plugin for signing, the user associates the plugin as part of registering a signing key. E.g.
+  * `notation key add --name "mysigningkey" --id "keyid" --plugin "com.example.nv2plugin"`
+  * In the example, the command registers a signing key in `/notation/config.json`, where `mysigningkey` is a friendly key name to refer during signing operation from the CLI, `id` is an key identifier known to plugin that is used for signing, and the value of `plugin` specifies it's using a plugin located at `~/.notation/plugins/com.example.nv2plugin/notation-com.example.nv2plugin`.
+  
+```jsonc
+{
+  "signingKeys": {  
+    "default": "",  
+    "keys": [
+      {  
+        "name": "mysigningkey",  
+        "id" : "keyid",
+        "plugin": "com.example.nv2plugin"
+      }  
+    ]  
+  }
+}
+```
+
+### Plugin contract
+
+* Notation will invoke the plugin executable for each command (e.g. sign, verify), pass inputs through `stdin` and get output through `stdout` and `stderr`.
+* The command will be passed as the first argument to the plugin e.g. `notary-{plugin-name} <command>`. A JSON request is passed using `stdin`. The plugin is expected to return a JSON response through `stdout` with a `0` exit code for successful response, and a non-zero exit code with a JSON error response in `stderr` for error response. Each command defines its request, response and error contract. To avoid any additional content like debug or info level logging from dependencies and inbuilt libraries, the plugin implementation should redirect any output to `stdout` on initialization, and only send the JSON response away from `stdout` when the command execution completes. E.g. For golang, set [`os.Stdout`](https://pkg.go.dev/os#pkg-variables) to point to a log file.
+* Every request JSON will contain a `contract-version` top level attribute whose value will indicate the plugin contract version. Contract version is revised when there are changes to command request/response, new plugin commands are introduced, and supported through Notation.
+* For an error response, every command returns a non-zero exit code 1, with an OPTIONAL JSON error response in `stderr`. It is recommended to return an error response to help user troubleshoot the error.  There is no need to send different exit codes for different error conditions, as Notation (which will call the plugin and parse the response) will use `error-code` in the error response to interpret different error conditions if needed. Notation will attempt to parse the error response in `stderr` when exit code is 1, else treat it as a general error for any other non-zero exit codes. An implementation can 
+
+```jsonc
+{
+  // Each command defines expected error codes
+  "error-code" : "<error code>",  
+  // Plugin defined error message.
+  "error-message" : "User friendly error message"
+
+  // Optional plugin defined additional key value pairs related to the error.
+  "error-metadata" : {
+    "key1" : "value1",
+    "key2" : "value2"
+  }
+}
+```
+
+### Plugin metadata
+
+* Every plugin MUST implement a metadata discovery command called `discover`. 
+* Notation will invoke this command as part of signing and verification workflows to discover the capabilities of the plugin and subsequently invoke commands corresponding to the capabilities. Notation will invoke the `discover` command every time a signing or verification workflow is invoked to discover metadata. Notation will not cache the response of `discover` command as it involves invalidating cache when a plugin is updated, detecting a plugin update can be non trivial. The `discover` command is expected to return only static data, and plugin implementors should avoid making remote calls in this command.
+
+***discover***
+
+*Request* - None
+
+*Response*
+All response attributes are required.
+
+```jsonc
+{
+  // Plugin name that matches its install dir
+  // e.g. "com.example.nv2plugin".
+  "plugin-name" : "<plugin name>",  
+  // Plugin friendly name.
+  "plugin-friendly-name" : "<friendly name>",
+  // Plugin publisher controlled version.
+  "plugin-version" : "<version>",
+  // Plugin webpage for support or documentation.
+  "plugin-url" : "<URL>",
+  
+  // List of contract versions supported by the plugin, one per major version
+  "supported-contract-versions" : [ ],
+  
+  // Currently one of 
+  // SIGNATURE_GENERATOR or
+  // SIGNATURE_ENVELOPE_GENERATOR
+  "capabilities" : [
+    
+  ]
+}
+```
+
+*plugin-name* - Plugin name uses reverse domain name notation to avoid plugin name collisions.
+
+*supported-contract-versions* - The list of contract versions supported by the plugin. Currently this list must include only one version, per major version. Post initial release, Notation may add new features through plugins, in the form of new commands (e.g. tsa-sign for timestamping), or additional request and response parameters. Notation will publish updates to plugin interface along with appropriate contract version update. Backwards compatible changes (changes for which older version of plugin continue to work with versions of Notation using newer contract version) like new optional parameters on existing contracts, and new commands will be supported through minor version contract updates, breaking changes through major version updates. Plugin `discover` command returns the contract version a plugin supports. Notation will evaluate the minimum plugin version required to satisfy a user's request, and reject the request if the plugin does not support the required version.
+
+*capabilities* - Non empty list of features supported by a plugin. Each capability such as `SIGNATURE_ENVELOPE_GENERATOR` requires one of more commands to be implemented by the plugin. When new features are available for plugins to implement, an implementation may choose to not implement it, and therefore will not include the feature in capabililies. Notation will evaluate the capability required to satisfy a user’s request, and reject the request if the plugin does not support the required capability.
+
+## Signing interfaces
+
+### Requirements
+
+* The interface MUST NOT be tied to a specific signature envelope format.
+
+Notation will support plugins to be developed against the following interfaces - *Signature Generator*, and *Signature Envelope Generator*. These interfaces target abstraction levels that satisfy most plugin integration scenarios.
+
+### Signature Generator
+
+This interface targets plugins that integrate with providers of basic cryptographic operations. E.g. Local PIV/PKCS#11 hardware tokens, remote KMS, or key vault services. Plugins that target this interface will only generate a raw signature given a payload to sign. Notation will package this signature into a signature envelope, and generate the signature manifest. Notation will also generate the TSA signature if indicated by the user. The plugin does not need to be signature envelope format aware, and will continue to work if Notary adopts additional signature formats.
+
+#### Signing workflow using plugin
+
+1. Given a user request to sign `oci-artifact`, with `keyName` (the friendly key name)
+1. Pull the image manifest using `oci-artifact` url, and construct a descriptor
+1. Append any user provided metadata and Notary metadata as descriptor annotations.
+1. Determine if the registered key uses a plugin
+1. Execute the plugin with `discover` command
+    1. If plugin supports capability `SIGNATURE_GENERATOR`
+        1. Generate the payload to be signed. For JWS this includes [JWS](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#supported-signature-envelopes) payload with `subject` claim as descriptor, additional JWS claims, and protected headers.
+        2. Execute the plugin with `generate-signature` command, set `request.keyDefinition` to key definition corresponding to `keyName`, `request.payload` to the generated payload.
+        3. Validate the generated signature, return an error if any of the checks fails.
+           1. Check if `response.signingAlgorithm` is one of [supported signing algorithms](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#algorithm-selection).
+           2. Check that the plugin did not modify `request.payload` before generating the signature, and the signature is valid for the given payload. Verify the hash of the `request.payload` against `response.signature`, using the public key of signing certificate (leaf certificate) in `response.certificateChain` along with the `response.signingAlgorithm`. This step does not include certificate chain validation (certificate chain leads to a trusted root configured in Notation), or revocation check.
+           3. Check that the `response.certificateChain` conforms to [Certificate Requirements](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#certificate-requirements).
+        4. Assemble the JWS Signature envelope using `response.signature`, `response.signingAlgorithm` and `response.certificateChain`. Notation may also generate and include timestamp signature in this step.
+        5. Generate a signature manifest for the given signature envelope.
+    2. Else if plugin supports capability `SIGNATURE_ENVELOPE_GENERATOR` *(covered in next section)*
+    3. Return an error
+
+***generate-signature***
+
+*Request*
+
+```jsonc
+{
+  "contract-version" : <major-version.minor-version>,
+
+  // Complete key definition from
+  // /notation/config.json /signingKeys/keys with matching key name
+  "keyDefinition" : <key definition>,
+
+  "payload" : <Base64 encoded payload to be signed>
+}
+```
+
+*keyDefinition* : Required field that contains the complete key definition, which includes friendly name (`name`), key identifier (`id`) and any additional custom fields required by the plugin.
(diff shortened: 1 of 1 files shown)


priteshbandi:
Can you please run markdown linter on the change.

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Plugin installation and config
+
+* Plugin publisher will provide instructions to download and install the plugin. Plugins intended for public distribution should also include instructions for users to verify the authenticity of the plugin.
+* Each plugin executable and dependencies are installed under directory `~/.notation/plugins/{plugin-name}` with an executable under that directory `~/.notation/plugins/{plugin-name}/notation-{plugin-name}`. The executable can be a shim which calls plugin dependecies installed elsewhere on the file system.
+* To use a plugin for signing, the user associates the plugin as part of registering a signing key. E.g.
+  * `notation key add --name "mysigningkey" --id "keyid" --plugin "com.amazonaws.signer.nv2plugin"`
```
rchincha, line 30: If there are multiple plugins in the plugins-dir, does `com.amazonaws.signer.nv2plugin` map to a particular plugin?

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+  * In the example, the command registers a signing key in `/notation/config.json`, where `mysigningkey` is a friendly key name to refer during signing operation, `id` is an AWS resource identifier that is used for signing, and the value of `plugin` specifies it's using a plugin located at `~/.notation/plugins/com.amazonaws.signer.nv2plugin/notation-com.amazonaws.signer.nv2plugin`.
```
rchincha: Nitpicking.

`com.amazonaws.signer.nv2plugin` repetitive in `~/.notation/plugins/com.amazonaws.signer.nv2plugin/notation-com.amazonaws.signer.nv2plugin

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+#### Signing workflow using plugin
+
+1. Given a user request to sign `oci-artifact`, with `key`
+1. Pull the image manifest using `oci-artifact` url, and construct a descriptor
+1. Append any user provided metadata and Notary metadata as descriptor annotations.
+1. Determine if the registered key uses a plugin
+1. Execute the plugin with `discover` command
+    1. If plugin supports capability `SIGNATURE_GENERATOR`
+        1. Generate the payload to be signed. For JWS this includes [JWS](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#supported-signature-envelopes) payload with `subject` claim as descriptor, additional JWS claims, and protected headers.
+        2. Execute the plugin with `generate-signature` command, set `request.key` to key definition, `request.payload` to the generated payload.
```
rchincha: `request.key` must be optional since signing keys may be private to the remote API.
You are really asking for two modes here:
1. I have the key but can't do crypto (at least not native to notation), so let plugin handle that
2. I don't the key and of course can't do crypto, so let plugin ship the payload and get it signed.

specs/plugin-extensibility.md
gokarnm: That was the intent, `/plugin` has a child dir called `{plugin-name}`, which contains an executable called `notation-{plugin-name}`.

specs/plugin-extensibility.md
gokarnm: The `request.key` is the key definition in notation config, here `name` is a friendly name to refer from CLI, `id` is the plugin/crypto specific identifier that allows plugin to know which key to use. They do not contain the value of signing key itself. I should probably change `request.key` to `request.keyDefinition` or similar to make this clearer.

``` json
"keys": [
      {  
        "name": "mysigningkey",  
        "id" : "keyid",
        "plugin": "com.example.nv2plugin"
      }  
```

SteveLasker:
@aramase, can you help here?

specs/plugin-extensibility.md
rchincha: `~/.notation/plugins/com.amazonaws.signer.nv2plugin/notation-com.amazonaws.signer.nv2plugin` => `~/.notation/plugins/com.amazonaws.signer/nv2`


Your call though.

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+* A plugin MAY implement a subset of capabilities (features) available in plugin contract. E.g A plugin may implement signing feature, but not verification.
+* Notation and plugins MAY be updated independently in an environment.
+  * Notation MUST work with a plugin that implements a matching or lower minor version of the plugin contract. Notation SHALL NOT support using a plugin with higher version of plugin contract.
+  * A plugin MUST support a single plugin contract version, per major version.
+
+Notation will invoke plugins as executable, pass parameters using command line arguments, and use standard IO streams to pass request/response payloads. This mechanism is used as Go language (used to develop [Notation library](https://github.com/notaryproject/notation-go-lib)) does not have a [in built support](https://github.com/golang/go/issues/19282) to load and execute plugins that works across OS platforms. Other mechanisms like gRPC require every plugin to be implemented as a service/deamon.
```
priteshbandi:
```suggestion
Notation will invoke plugins as executable, pass parameters using command line arguments, and use standard IO streams to pass request/response payloads. This mechanism is used as Go language (used to develop [Notation library](https://github.com/notaryproject/notation-go-lib)) does not have a [in built support](https://github.com/golang/go/issues/19282) to load and execute plugins that works across OS platforms. Other mechanisms like gRPC require every plugin to be implemented as a service/daemon.
```

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Plugin installation and config
+
+* Plugin publisher will provide instructions to download and install the plugin. Plugins intended for public distribution should also include instructions for users to verify the authenticity of the plugin.
```
priteshbandi, line 27: >Plugins intended for public distribution should also include instructions for users to verify the authenticity of the plugin.

Should we say MUST instead of SHOULD include instructions for users to verify the authenticity of the plugin?
If a user cannot determine the authenticity of the plugin then they cannot trust plugin.

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+* Each plugin executable and dependencies are installed under directory `~/.notation/plugins/{plugin-name}` with an executable under that directory `~/.notation/plugins/{plugin-name}/notation-{plugin-name}`. The executable can be a shim which calls plugin dependecies installed elsewhere on the file system.
```
priteshbandi, line 28: nit: The `notation` in `notation-{plugin-name}` of `~/.notation/plugins/{plugin-name}/notation-{plugin-name}` is redundant.

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+      }  
+    ]  
+  }
+}
+```
+
+### Plugin contract
+
+* Notation will invoke the plugin executable for each command (e.g. sign, verify), pass inputs through `stdin` and get output through `stdout` and `stderr`.
+* The command will be passed as the first argument to the plugin e.g. `notary-{plugin-name} <command>`. A JSON request is passed using `stdin`. The plugin is expected to return a JSON response through `stdout` with a `0` exit code for successful response, and a non-zero exit code with a JSON error response in `stderr` for error response. The request, response and error contract for each command will be defined. To avoid any additional content like debug or info level logging from dependencies and inbuilt libraries, the plugin implementation should redirect any output to `stdout` on initialization, and only send the JSON response away from `stdout` when the command execution completes. E.g. For golang, set [`os.Stdout`](https://pkg.go.dev/os#pkg-variables) to point to a log file.
```
priteshbandi: > The request, response and error contract for each command will be defined.

Haven't we already defined this?

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+* Every request JSON will contain a `contract-version` top level attribute whose value will indicate the plugin contract version. Contract version is revised when there are changes to command request/response, new plugin commands are introduced, and supported through Notation.
+* For an error response, every command returns a non-zero exit code 1, with a standard JSON error response in `stderr`. There is no need to send different exit codes for different error conditions, as Notation (which will call the plugin and parse the response) will use `error-code` in the error response to intepret different error conditions if needed.
```
priteshbandi: What happens if the exit code is neither 0 nor 1? Should implementation throw an error or consider all non-zero exit codes as an error?

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+  "error-code" : "<error code>",  
+  // Plugin defined error message.
+  "error-message" : "User friendly error message"
+
+  // Optional plugin defined additional key value pairs related to the error.
+  "error-metadata" : {
+    "key1" : "value1",
+    "key2" : "value2"
+  }
+}
```
priteshbandi, line 67: Should returning error JSON be optional and notation should consider it as InternalFailure?
There are cases where the plugin might fail unexpectedly and might not be able to emit error JSON. Or the user might want to implement a plugin with minimal error handling.

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+  "plugin-name" : "<plugin name>",  
+  // Plugin friendly name.
+  "plugin-friendly-name" : "<friendly name>",
+  // Plugin publisher controlled version.
+  "plugin-version" : "<version>",
+  // Plugin webpage for support or documentation.
+  "plugin-url" : "<URL>",
+  
+  // Contract versions supported by the plugin
+  "supported-contract-versions" : "<contract-version>",
```
priteshbandi: nit: this should be an array.

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+  // SIGNATURE_ENVELOPE_GENERATOR
+  "capabilities" : [
+    
+  ]
+}
+```
+
+*plugin-name* - Plugin name uses reverse domain name notation to avoid plugin name collisions.
+
+*supported-contract-versions* - The list of contract versions supported by the plugin. Currently this list must include only one version, per major version. Post initial release, Notation may add new features through plugins, in the form of new commands (e.g. tsa-sign for timestamping), or additional request and response parameters. Notation will publish updates to plugin interface along with appropriate contract version update. Backwards compatible changes (changes for which older version of plugin continue to work with versions of Notation using newer contract version) like new optional parameters on existing contracts, and new commands will be supported through minor version contract updates, breaking changes through major version updates. Plugin `discover` command returns the contract version a plugin supports. Notation will evaluate the minimum plugin version required to satisfy a user's request, and reject the request if the plugin does not support the required version.
```
priteshbandi, line 108: > Currently this list must include only one version, per major version.

Does this mean plugin publishers should only specify a major version of the contract?

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+        2. Execute the plugin with `generate-signature` command, set `request.key` to key definition, `request.payload` to the generated payload.
+        3. Validate the generated signature, return an error if any of the checks fails.
+           1. Check if `response.signingAlgorithm` is one of [supported signing algorithms](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#algorithm-selection).
+           2. Check that the plugin did not modify `request.payload` before generating the signature, and the signature is valid for the given payload. Verify the hash of the `request.payload` against `response.signature`, using the public key of signing certificate (leaf certificate) in `response.certificateChain` along with the `response.signingAlgorithm`. This step does not include certificate chain validation (certificate chain leads to a trusted root configured in Notation), or revocation check.
+           3. Check that the `response.certificateChain` confirms to [Certificate Requirements].
```
priteshbandi: Missing hyperlink for `[Certificate Requirements]`?

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+*certificateChain* : Ordered list of certificates starting with leaf certificate and ending with root certificate.
+
+*Error codes*
+
+* VALIDATION_ERROR - Any of the required request fields was empty, or a value was malformed/invalid. Includes condition where the key referenced by `keyId` was not found.
+* UNSUPPORTED_CONTRACT_VERSION - The contract version used in the request is unsupported.
+* ACCESS_DENIED - Authentication/authorization error to use given key.
+* TIMEOUT - The operation to generate signature timed out and can be retried by Notation.
+* THROTTLED - The operation to generate signature was throttles and can be retried by Notation.
+* ERROR - Any general error that does not fall into previous error categories.
```
priteshbandi, line 196: s/ERROR/INTERNAL_ERROR ?

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+
+*certificateChain* : Ordered list of certificates starting with leaf certificate and ending with root certificate.
+
+*Error codes*
+
+* VALIDATION_ERROR - Any of the required request fields was empty, or a value was malformed/invalid. Includes condition where the key referenced by `keyId` was not found.
+* UNSUPPORTED_CONTRACT_VERSION - The contract version used in the request is unsupported.
+* ACCESS_DENIED - Authentication/authorization error to use given key.
+* TIMEOUT - The operation to generate signature timed out and can be retried by Notation.
+* THROTTLED - The operation to generate signature was throttles and can be retried by Notation.
```
priteshbandi, line 195: Shouldn't this be handled internally by the plugin? Why does notation need to know about access denied, timeout and throttling?

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+
+#### Signing workflow using plugin
+
+1. Given a user request to sign `image`, with `key`
+1. Pull the image manifest using `image` url, and construct a descriptor
+1. Append any user provided metadata and Notary metadata as descriptor annotations.
+1. Determine if the registered key uses a plugin
+1. Execute the plugin with `discover` command
+    1. If plugin supports capability `SIGNATURE_ENVELOPE_GENERATOR`
+        1. Execute the plugin with `generate-envelope` command, set `request.key` to key definition (from `/notation/config.json`), `request.payload` to base64 encoded descriptor, `request.payloadType` to `application/vnd.oci.descriptor.v1+json` and `request.signatureEnvelopeType` to `application/vnd.cncf.notary.v2.jws.v1`.
```
priteshbandi: Will Notation generate timestamp signature for `Signature Envelope Generator` plugin or its responsibility of plugin publisher?

specs/plugin-extensibility.md
gokarnm: Was waiting for your PR merge :), will update

specs/plugin-extensibility.md
gokarnm: There is limited things the spec can do here, unless there is a uniform mechanism to enforce it and Notation to verify. Plugins packages may be distributed for internal or public consumption and may be signed/unsigned or may use any mechanism to assert authenticity.

specs/plugin-extensibility.md
gokarnm: Reworded.
```suggestion
* The command will be passed as the first argument to the plugin e.g. `notary-{plugin-name} <command>`. A JSON request is passed using `stdin`. The plugin is expected to return a JSON response through `stdout` with a `0` exit code for successful response, and a non-zero exit code with a JSON error response in `stderr` for error response. Each command defines its request, response and error contract. To avoid any additional content like debug or info level logging from dependencies and inbuilt libraries, the plugin implementation should redirect any output to `stdout` on initialization, and only send the JSON response away from `stdout` when the command execution completes. E.g. For golang, set [`os.Stdout`](https://pkg.go.dev/os#pkg-variables) to point to a log file.
```

specs/plugin-extensibility.md
gokarnm: Good point, addressed.
```suggestion
* For an error response, every command returns a non-zero exit code 1, with an OPTIONAL JSON error response in `stderr`. It is recommended to return an error response to help user troubleshoot the error.  There is no need to send different exit codes for different error conditions, as Notation (which will call the plugin and parse the response) will use `error-code` in the error response to interpret different error conditions if needed. Notation will attempt to parse the error response in `stderr` when exit code is 1, else treat it as a general error for any other non-zero exit codes. An implementation can 
```

specs/plugin-extensibility.md
gokarnm: Addressed for previous comment.

specs/plugin-extensibility.md
gokarnm:
```suggestion
  // List of contract versions supported by the plugin, one per major version
  "supported-contract-versions" : [ ],
```

specs/plugin-extensibility.md
gokarnm:
```suggestion
           3. Check that the `response.certificateChain` conforms to [Certificate Requirements](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#certificate-requirements).
```

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+* VALIDATION_ERROR - Any of the required request fields was empty, or a value was malformed/invalid. Includes condition where the key referenced by `keyId` was not found, payload type or signature envelope type in the request is unsupported by the plugin.
+* UNSUPPORTED_CONTRACT_VERSION - The contract version used in the request is unsupported.
+* ACCESS_DENIED - Authentication/authorization error to use given key.
+* TIMEOUT - The operation to generate signature timed out and can be retried by Notation.
+* THROTTLED - The operation to generate signature was throttles and can be retried by Notation.
+* ERROR - Any general error that does not fall into previous error categories.
+
+## Verification extensibility
+
+**TBD**
```
sajayantony: Having a issue to track this before the next release would be good. Similarly for the **Threat Model** as well.

sajayantony:
If we create the required issues for TBD section would be good for initial Push.

sajayantony (approved):
Good for initial push with issues created.

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+  "error-metadata" : {
+    "key1" : "value1",
+    "key2" : "value2"
+  }
+}
+```
+
+### Plugin metadata
+
+* Every plugin MUST implement a metadata dicovery command called `discover`.
```
aramase:
```suggestion
* Every plugin MUST implement a metadata discovery command called `discover`. 
```

specs/plugin-extensibility.md
aramase, line 108: Do we also need to define the supportability contract for major and minor versions? How long will a particular minor version be supported in notation when there are multiple minor versions?

specs/plugin-extensibility.md
aramase:
```suggestion
           3. Check that the `response.certificateChain` conforms to [Certificate Requirements].
```

specs/plugin-extensibility.md
shizhMSFT:
```suggestion
Notation will invoke plugins as executable, pass parameters using command line arguments, and use standard IO streams to pass request/response payloads. This mechanism is used as Go language (used to develop [Notation library](https://github.com/notaryproject/notation-go-lib)) does not have a [built-in support](https://github.com/golang/go/issues/19282) to load and execute plugins that works across OS platforms. Other mechanisms like gRPC require every plugin to be implemented as a service/deamon.
```

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+
+*payload* : Required field that contains payload to be signed. This is currently the JWSPayload that includes the subject descriptor and other JWS claims.
+
+*Response*
+All response attributes are required.
+
+```json
+{    
+  // The same key id as passed in the request.
+  "keyId" : <key id>,
```
aramase, line 178: Should we use `key` here instead of `keyId` as the key definition is passed as part of the request?

specs/plugin-extensibility.md
aramase, line 196: These error codes are good to be surfaced out but IMO notation should not perform any retry action. In case of throttled requests, it's possible request can only be retried after a certain time defined in `retry-after` header. The plugin can decide to (1) retry the request internally before giving up and returning an error or (2) return the `retry-after` header info as part of error message, so the user can retry the request after the time has elapsed.

aramase: This also makes me think there should be a context defined with timeout as part of the request to the plugin. We can use [CommandContext](https://pkg.go.dev/os/exec#CommandContext) to ensure the request doesn't hang indefinitely

specs/plugin-extensibility.md
shizhMSFT: Similar pattern for executables can be found in [application bundles](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html).

specs/plugin-extensibility.md
```diff
@@ -0,0 +1,277 @@
...
+  "plugin-version" : "<version>",
+  // Plugin webpage for support or documentation.
+  "plugin-url" : "<URL>",
+  
+  // Contract versions supported by the plugin
+  "supported-contract-versions" : "<contract-version>",
+  
+  // Currently one of 
+  // SIGNATURE_GENERATOR or
+  // SIGNATURE_ENVELOPE_GENERATOR
```
aramase, line 99: If `SIGNATURE_GENERATOR` and `SIGNATURE_ENVELOPE_GENERATOR` are supported by the plugin which one takes precedence when `notation sign` is invoked?

specs/plugin-extensibility.md
shizhMSFT: nit: `jsonc` instead of `json` renders better.

```jsonc
{
  // Each command defines expected error codes
  "error-code" : "<error code>",  
  // Plugin defined error message.
  "error-message" : "User friendly error message"
  // Optional plugin defined additional key value pairs related to the error.
  "error-metadata" : {
    "key1" : "value1",
    "key2" : "value2"
  }
}

File: specs/plugin-extensibility.md [hunk 20]
```diff
@@ -0,0 +1,277 @@
...
+
+  // Complete key definition from
+  // /notation/config.json /signingKeys/keys with matching key name
+  "key" : <key definition>,
+
+  "payload" : <Base64 encoded payload to be signed>
+}
+```
+
+*key* : Required field that contains the complete key definition, which includes friendly name (`name`), key identifier (`id`) and any additional custom fields required by the plugin.
```
shizhMSFT: Could you provide an example request?

specs/plugin-extensibility.md
shizhMSFT:
```suggestion
        1. Execute the plugin with `generate-envelope` command, set `request.key` to key definition (from `/notation/config.json`), `request.payload` to base64 encoded descriptor, `request.payloadType` to `application/vnd.oci.descriptor.v1+json` and `request.signatureEnvelopeType` to a pre-defined type (default to `application/vnd.cncf.notary.v2.jws.v1`).
```

shizhMSFT (approved):
LGTM on this initial spec. We can add the specs for timestamping and verification in future PRs.

specs/plugin-extensibility.md
gokarnm: I'd prefer to keep the existing convention, I agree it's repetitive, though it does not directly impact the end user experience. Initially the executable name was `{plugin-name}`, then we prefixed notation following docker's convention for CLI plugins. @sajayantony let me know if you have any feedback to simplify this.

specs/plugin-extensibility.md
gokarnm: Currently we expect a signing plugin to implement one of the signing interfaces not both. Let me know if this does not work for a specific case.

specs/plugin-extensibility.md
gokarnm: @priteshbandi @aramase, plugin publishers specify the major.minor version of contract interface, for each major version. We are only putting in place mechanisms to support breaking and non breaking changes to plugin interface. These will not rev at the same rate as Notation updates. Notation should ideally support all minor versions of contract. The intent is plugin publishers can choose to implement the latest version of contract, but we don't want to break support for plugins implementing older minor versions of contract (within the major version).

specs/plugin-extensibility.md
gokarnm: The `key` in the request was actually the complete key definition, I'll rename that field to `keyDefinition` to be clearer.

specs/plugin-extensibility.md
gokarnm: This is supposed to be a catch all error that would cover any known, and unhandled plugin errors that do not fall into other categories.

gokarnm: Agreed, Notation should not retry. The errors would be surfaced to caller invoking Notation, and it can decide to retry if needed.

gokarnm: Can you provide more details about the context with timeout as part of the request, maybe with an example, I didn't understand that.

specs/plugin-extensibility.md
gokarnm: Good point, will address in next rev.

specs/plugin-extensibility.md
gokarnm: Done.

gokarnm:
Addressed review feedback, linked open issues for verification plugin and threat model, and squashed commits.

priteshbandi (approved):
LGTM

NiazFK (commented):
LGTM

NiazFK (approved):

aramase (approved):
lgtm

for some of the outstanding questions, we can follow up during implementation.

SteveLasker (approved):
LGTM

PR #131, opened 2022-02-04
Add support for signature filtering.

priteshbandi:
### Why we need this change

An OCI artifact can have multiple signatures associated with it. To efficiently perform the signature verification we need a mechanism to filter out the signatures based on the user trust criteria. Otherwise, we will have to download and evaluate all the associated signatures(until a match is found). Also, there is no upper bound on the number of signatures that can be associated with an artifact, this poses an availability risk as signature evaluation as signature verification workflow might timeout(if configured) or run for a long time.

### Supported filtering criteria
Filter based on SHA256 of signing certificate and certificate chain.

### Mechanism/Design
To filter out signatures based on the aforementioned trust criteria, we need to surface this trust criteria information in the signature artifact manifest. The only entry in the signature artifact manifest that allows arbitrary data is annotations, so we are using annotations to surface certificate(s) identifier i.e. SHA-256 fingerprints of certificate and certificate chain.


SteveLasker (approved):
LGTM

gokarnm:
LGTM!

gokarnm (approved):

signature-specification.md
```diff
@@ -20,7 +20,8 @@ The signature manifest has an artifact type which specifies it's a Notary V2 sig
...
   - **`mediaType`** (*string*): This REQUIRED property contains media type of signature envelope blob. The supported value is `application/jose+json`
 - **`subject`** (*descriptor*): A REQUIRED artifact descriptor referencing the signed manifest, including, but not limited to image manifest, image index, oras-artifact manifest.
-- **`annotations`** (*string-string map*): This OPTIONAL property contains arbitrary metadata for the artifact manifest. It can be used to store information about the signature.
+- **`annotations`** (*string-string map*): This REQUIRED property contains metadata for the artifact manifest. It is being used to store information about the signature. Keys using the `org.cncf.notary` namespace are reserved for use in Notary and MUST NOT be used by other specifications.
+  - **`org.cncf.notary.x509certs.fingerprint.sha256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprint of signing certificate and certificate chain used for signature generation. The list of fingerprints is present as a JSON array string.
```
shizhMSFT, line 24: nit: On the naming of `fingerprint` and `thumbprint`, which one is more popular or consistent?

shizhMSFT, line 24: nit: You may also want to note that the fingerprint is in HEX format.

shizhMSFT (approved):
LGTM

PR #132, opened 2022-02-04
Add support for verification of artifact signed using publicly trusted certificates.

priteshbandi:
### Introduction
Update trust policy to support verification of OCI artifacts signed using publicly trusted codesigning certificates. The certificates issued by CAs that are publicly trusted(abides by CAB forum guidelines and trusted by many operating systems) are referred to as publicly trusted certificates.

### Why do we need this change
The user MUST be able to use a codesigning certificate issued from publicly trusted CAs e.g. Digicert, Entrust, Verisign, etc to sign and verify OCI artifacts.
Publicly trusted CAs issue codesigning certificates to various entities from the same CA, the only way for a consumer to verify that the artifact came from a specific publisher is to pin(in trust-policy) on the publisher's signing certificate. Signing certificates have limited validity and it's recommended to rotate keys periodically. If consumer pins on the publisher's signing certificate, the rotation of the publisher's singing certificate will require all consumers to update the trust policy to pin on the new certificate.
Since there can be thousands of consumers for an artifact, asking each and every consumer to update their trust store is not scalable. Thus we need a scalable mechanism where a consumer can reliably trust an artifact signed using a publicly trusted certificate.

**Q. Why trust anchor is not scoped to a trusted certificate.**
**A.** If a user added certificates in the trusted roots, it means they trust those CAs to do the right thing(validations) when issuing certificates thus we don't need to tie trust anchor to trust root.

**Q. Why notary v2 supports only certificate subject in trust anchor?**
**A.** The Signing certificate's subject can uniquely identify an organization and none of the other mandatory attributes have this capability. Also, the specification is extensible to support other attributes.


SteveLasker (approved):
LGTM

gokarnm (approved):

shizhMSFT (approved):
LGTM

Issue #95, opened 2021-08-17
Key Discovery


SteveLasker:
We've scoped Notary v2 to not support TOFU.
How will clients acquire the public keys for different registries or repos?

mnm678:
Per the [TUF design](https://github.com/notaryproject/nv2/blob/prototype-tuf/tuf-design.md#design), if we can distribute root keys that do namespaced delegations for a registry or registries, we can reduce the number of keys that need to be distributed to end users. This reduced number of keys could be shared with client systems when they are provisioned using existing secret distribution methods, like SPIFFE/SPIRE.

letmaik:
For SCITT (see [draft spec](https://datatracker.ietf.org/doc/html/draft-birkholz-scitt-architecture-00.html)), the proposal is to use DID both as mechanism to establish issuer identifiers and to do key discovery (via DID resolution), while not being constrained to a single identity system but having extensibility through DID methods. For Notary to support this, signing envelopes would have to include the DID as issuer in the protected header plus a key ID.

iamsamirzon:
Key discovery/distribution can be done outside of Notary. Additionally, I believe we have discussed, that users can inspect a signature using Notation to start their discovery process.

SteveLasker:
Re-opening, as key discovery is a critical component to notary usability. We are currently focusing on the `notation inspect` [notation notaryproject/notation#238](https://github.com/notaryproject/notation/issues/238) command to provide the discovery method, for how to find the public key.

dtzar:
Steve - I believe we need to fill in more details on 238 because it is currently vague and not connected to this purpose as it is written.  Key discovery is vital for verification of course and we need to make this easy for users and producers.  Simply getting various metadata back about a signature not necessarily important in general, especially considering we have oras.

SteveLasker:
The focus here is we at least need to know where the public key is located. Then, users can download the public key, configure their trust stores and have an e2e story.  See https://github.com/notaryproject/roadmap/issues/74, and perhaps we can either copy the content from there, or re-use that issue for tracking.

iamsamirzon:
@SteveLasker  and @dtzar   - I think we are saying the same thing, that some sort of "Signature Inspection" can be used as a method to determine the publisher and the public key. There are already other issues/PR like you both pointed out https://github.com/notaryproject/notation/issues/238 and https://github.com/notaryproject/roadmap/issues/74. So could we close this one as a duplicate?

dtzar:
[...]

PR #139, opened 2022-03-02
COSE Signature Envelope

### LINKED ISSUE #117: Implement COSE as signing envelope
I'd like to see Notary v2 support [CBOR Object Signing and Encryption (COSE)](https://datatracker.ietf.org/doc/html/rfc8152) as the signing envelope for content, as it is flexible to support all of the Nv2 signing requirements including [IETF draft specs on support for x.509](https://datatracker.ietf.org/doc/html/draft-ietf-cose-x509-08). 

Also, COSE has broader alignments across the industry in CoMID/CoSWID, and standards for software manifests and remote attestation including C2PA .

Consider this as an update to [Issue #99 ](https://github.com/notaryproject/notaryproject/issues/99).

shizhMSFT:
Replace JWS with COSE for the envelope format.

Highlight: Unlike JWS, `alg` is not a required header in COSE. Therefore, we can remove it completely, making the signature envelope more secure.

Resolves #117

### DIFF
--- a/signature-specification.md
+++ b/signature-specification.md
@@ -149,8 +149,8 @@ Note: The above example is represented using the [extended CBOR diagnostic notat
 - **`crit`** (*array of integers or strings*): This REQUIRED property (label: `2`) lists the headers that implementation MUST understand and process.
   The array MUST contain `3` (`cty`), and `signingtime`. If `expiry` is presented, the array MUST also contain `expiry`.
 - **`cty`** (*string*): The REQUIRED property content-type (label: `3`) is used to declare the media type of the secured content (the payload).
-- **`signingtime`** (*integer* or tagged *datetime*): The REQUIRED property identifies the time at which the signature was generated.
-- **`expiry`** (*integer* or tagged *datetime*): This OPTIONAL property contains the expiration time on or after which the signature must not be considered valid.
+- **`signingtime`** (*datetime*): The REQUIRED property identifies the time at which the signature was generated.
+- **`expiry`** (*datetime*): This OPTIONAL property contains the expiration time on or after which the signature must not be considered valid.
 
 **UnprotectedHeaders**: Notary v2 supports two unprotected headers: `timestamp` and `x5chain`.
 
--- a/signature-specification.md
+++ b/signature-specification.md
@@ -17,7 +17,7 @@ The signature manifest has an artifact type that specifies it's a Notary V2 sign
 
 - **`artifactType`** (*string*): This REQUIRED property references the Notary version of the signature: `application/vnd.cncf.notary.v2.signature`.
 - **`blobs`** (*array of objects*): This REQUIRED property contains collection of only one [artifact descriptor](https://github.com/oras-project/artifacts-spec/blob/main/descriptor.md) referencing signature envelope.
-  - **`mediaType`** (*string*): This REQUIRED property contains media type of signature envelope blob. The supported value are `application/jose+json` and `application/cose`.
+  - **`mediaType`** (*string*): This REQUIRED property contains media type of signature envelope blob. The supported value is `application/cose`.
 - **`subject`** (*descriptor*): A REQUIRED artifact descriptor referencing the signed manifest, including, but not limited to image manifest, image index, oras-artifact manifest.
 - **`annotations`** (*string-string map*): This REQUIRED property contains metadata for the artifact manifest.
   It is being used to store information about the signature.
@@ -70,7 +70,7 @@ A signature envelope consists of the following components:
 
 A signature envelope is `e = {m, v, u, s}` where `s` is signature.
 
-Notary v2 supports [JWS JSON Serialization](https://datatracker.ietf.org/doc/html/rfc7515) and [COSE_Sign1_Tagged](https://datatracker.ietf.org/doc/html/rfc8152#section-4) as signature envelope formats with some additional constraints but makes provisions to support additional signature envelope format.
+Notary v2 supports [COSE_Sign1_Tagged](https://datatracker.ietf.org/doc/html/rfc8152#section-4) as signature envelope format with some additional constraints but makes provisions to support additional signature envelope format.
 
 ### Payload
 
@@ -80,9 +80,7 @@ Notary v2 requires Payload to be the content **descriptor** of the subject manif
 1. Descriptor MAY contain `annotations` and if present it MUST follow the [annotation rules](https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules). Notary v2 uses annotations for storing both Notary specific and user defined signed attributes. The prefix `org.cncf.notary` in annotation keys is reserved for use in Notary v2 and MUST NOT be used outside this specification.
 1. Descriptor MAY contain `artifactType` field for artifact manifests, or the `config.mediaType` for `oci.image` based manifests.
 
-#### JSON Examples
-
-The media type of the content descriptor is `application/vnd.cncf.oras.artifact.descriptor.v1+json`.
+Examples: The media type of the content descriptor is `application/vnd.cncf.oras.artifact.descriptor.v1+json`.
 
 ```jsonc
 {
@@ -101,58 +99,6 @@ The media type of the content descriptor is `application/vnd.cncf.oras.artifact.
    "digest": "sha256:9834876dcfb05cb167a5c24953eba58c4ac89b1adf57f28f2f9d09af107ee8f0",
    "size": 32654
 }
-```
-
-#### CBOR Examples
-
-The media type of the content descriptor is `application/vnd.cncf.oras.artifact.descriptor.v1+cbor`.
-
-```sql
-a4 -- Map of size 4
-   69       -- Key:   UTF-8 text: 9 bytes
-      6d 65 64 69 61 54 79 70 65                      -- mediaType
-   78 2a    -- Value: UTF-8 text: 42 bytes
-      61 70 70 6c 69 63 61 74 69 6f 6e 2f 76 6e 64 2e -- application/vnd.
-      6f 63 69 2e 69 6d 61 67 65 2e 6d 61 6e 69 66 65 -- oci.image.manife
-      73 74 2e 76 31 2b 6a 73 6f 6e                   -- st.v1+json
-   66       -- Key:   UTF-8 text: 6 bytes
-      64 69 67 65 73 74                               -- digest
-   78 47    -- Value: UTF-8 text: 71 bytes
-      73 68 61 32 35 36 3a 37 33 63 38 30 33 39 33 30 -- sha256:73c803930
-      65 61 33 62 61 31 65 35 34 62 63 32 35 63 32 62 -- ea3ba1e54bc25c2b
-      64 63 35 33 65 64 64 30 32 38 34 63 36 32 65 64 -- dc53edd0284c62ed
-      36 35 31 66 65 37 62 30 30 33 36 39 64 61 35 31 -- 651fe7b00369da51
-      39 61 33 63 33 33 33                            -- 9a3c333
-   64       -- Key:   UTF-8 text: 4 bytes
-      73 69 7a 65                                     -- size
-   19 41 54 -- Value: Integer: 16724
-   6b       -- Key:   UTF-8 text: 11 bytes
-      61 6e 6e 6f 74 61 74 69 6f 6e 73                -- annotations
-   a1 -- Value: Map of size 1
-      78 1a    -- Key:   UTF-8 text: 26 bytes
-         69 6f 2e 77 61 62 62 69 74 2d 6e 65 74 77 6f 72 -- io.wabbit-networ
-         6b 73 2e 62 75 69 6c 64 49 64                   -- ks.buildId
-      63       -- Value: UTF-8 text: 3 bytes
-         31 32 33                                        -- 123
-```
-
-```sql
-a3 -- Map of size 3
-   69       -- Key:   UTF-8 text: 9 bytes
-      6d 65 64 69 61 54 79 70 65                      -- mediaType
-   6c       -- Value: UTF-8 text: 12 bytes
-      73 62 6f 6d 2f 65 78 61 6d 70 6c 65             -- sbom/example
-   66       -- Key:   UTF-8 text: 6 bytes
-      64 69 67 65 73 74                               -- digest
-   78 47    -- Value: UTF-8 text: 71 bytes
-      73 68 61 32 35 36 3a 39 38 33 34 38 37 36 64 63 -- sha256:9834876dc
-      66 62 30 35 63 62 31 36 37 61 35 63 32 34 39 35 -- fb05cb167a5c2495
-      33 65 62 61 35 38 63 34 61 63 38 39 62 31 61 64 -- 3eba58c4ac89b1ad
-      66 35 37 66 32 38 66 32 66 39 64 30 39 61 66 31 -- f57f28f2f9d09af1
-      30 37 65 65 38 66 30                            -- 07ee8f0
-   64       -- Key:   UTF-8 text: 4 bytes
-      73 69 7a 65                                     -- size
-   19 7f 8e -- Value: Integer: 32654
 ```
 
 ### Signed Attributes
--- a/signature-specification.md
+++ b/signature-specification.md
@@ -244,17 +244,12 @@ To leverage JWS claims validation functionality already provided by libraries, w
 - **`x5c`** (*array of strings*): This REQUIRED property contains the list of X.509 certificate or certificate chain([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the JWS.
   The certificate containing the public key corresponding to the key used to digitally sign the JWS MUST be the first certificate.
 
-- **`timestamp`** (*string*): This OPTIONAL property is used to store time stamp token.
-  Only [RFC3161]([rfc3161](https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2)) compliant TimeStampToken are supported.
-- **`x5c`** (*array of strings*): This REQUIRED property contains the list of X.509 certificate or certificate chain([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the JWS.
-  The certificate containing the public key corresponding to the key used to digitally sign the JWS MUST be the first certificate.
-
 **Signature**: In JWS signature is calculated by combining JWSPayload and protected headers.
 The process is described below:
 
 1. Compute the Base64Url value of ProtectedHeaders.
-1. Compute the Base64Url value of JWSPayload.
-1. Build message to be signed by concatenating the values generated in step 1 and step 2 using '.'
+2. Compute the Base64Url value of JWSPayload.
+3. Build message to be signed by concatenating the values generated in step 1 and step 2 using '.'
 `ASCII(BASE64URL(UTF8(ProtectedHeaders)) ‘.’ BASE64URL(JWSPayload))`
 1. Compute the signature on the message constructed in the previous step by using the signature algorithm defined in the corresponding header element: `alg`.
 1. Compute the Base64Url value of the signature produced in the previous step.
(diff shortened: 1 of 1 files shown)


signature-specification.md
```diff
@@ -120,125 +120,105 @@ Notary v2 requires the signature envelope to support the following signed attrib
...
 Unless explicitly specified as OPTIONAL, all fields are required.
-Also, there shouldn’t be any additional fields other than ones specified in JWSPayload, ProtectedHeader, and UnprotectedHeader.
 
-**JWSPayload a.k.a. Claims**:
-Notary v2 is using one private claim (`notary`) and two public claims (`iat` and `exp`).
-An example of the claim is described below
+**Payload**: COSE signs the payload as defined in the [Payload](#payload) section.
 
-```json
+**ProtectedHeaders**: Notary v2 supports only the following protected headers:
```
briankr-ms: I would suggest clarifying that other headers can be included, but Notary v2 will not process them (and will fail if they are set as 'crit'). For example, some COSE implementations will always add 'alg' - it's fine if nv2 won't rely on this, but I don't believe it should be a failure if it is present.

signature-specification.md
```diff
@@ -120,125 +120,105 @@ Notary v2 requires the signature envelope to support the following signed attrib
...
 
-- **`subject`**(*descriptor*): A REQUIRED top-level node consisting of the manifest that needs to be integrity protected.
-  Please refer [Payload](#payload) section for more details.
-- **`iat`**(*number*): The REQUIRED property Issued-at(`iat`) identifies the time at which the signature was issued.
-- **`exp`**(*number*): This OPTIONAL property contains the expiration time on or after which the signature must not be considered valid.
+- **`crit`** (*array of integers or strings*): This REQUIRED property (label: `2`) lists the headers that implementation MUST understand and process.
+  The array MUST contain `3` (`cty`), and `signingtime`. If `expiry` is presented, the array MUST also contain `expiry`.
+- **`cty`** (*string*): The REQUIRED property content-type (label: `3`) is used to declare the media type of the secured content (the payload).
+- **`signingtime`** (*integer*): The REQUIRED property identifies the time at which the signature was generated.
+- **`expiry`** (*integer*): This OPTIONAL property contains the expiration time on or after which the signature must not be considered valid.
```
briankr-ms: Clarify that these are not just "integers", but tagged [CBOR date/time values](https://datatracker.ietf.org/doc/html/rfc7049#section-2.4.1). Consider supporting other time formats such as RFC3339 or future [CBOR time](https://datatracker.ietf.org/doc/html/draft-ietf-cbor-time-tag).

signature-specification.md
```diff
@@ -120,125 +120,105 @@ Notary v2 requires the signature envelope to support the following signed attrib
...
-```json
 {
-    "timestamp": "<Base64(TimeStampToken)>",
-    "x5c": ["<Base64(DER(leafCert))>", "<Base64(DER(intermediateCACert))>", "<Base64(DER(rootCert))>"]
+  'timestamp': << TimeStampToken >>,
+  / x5chain / 33: [
+    << DER(leafCert) >>,
+    << DER(intermediate CACert) >>,
+    << DER(rootCert) >>
+  ]
```
briankr-ms, line 164: There may be value in including x5chain in the protected headers to ensure certs cannot be substituted in an attack. If not (or in addition to) that, x5t/34 should be a protected header.

signature-specification.md
```diff
@@ -120,125 +120,105 @@ Notary v2 requires the signature envelope to support the following signed attrib
...
+- **`timestamp`** (*byte string*): This OPTIONAL property is used to store time stamp token.
   Only [RFC3161]([rfc3161](https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2)) compliant TimeStampToken are supported.
-- **`x5c`** (*array of strings*): This REQUIRED property contains the list of X.509 certificate or certificate chain([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the JWS.
-  The certificate containing the public key corresponding to the key used to digitally sign the JWS MUST be the first certificate.
+- **`x5chain`** (*array of byte strings*): This REQUIRED property (label: `33` by [IANA](https://www.iana.org/assignments/cose/cose.xhtml#header-parameters)) contains the list of X.509 certificate or certificate chain ([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the COSE.
+  The certificate containing the public key corresponding to the key used to digitally sign the COSE MUST be the first certificate.
+  Optionally, this header can be presented in the protected header.
```
briankr-ms, line 174: If x5chain is not protected, should require that x5t/34 is. There may also be value in simply requiring that x5chain is protected. This guards against a potential cert substitution attack, at the expense of requiring a larger signature envelope size because the header can't be stripped in environments where the value is known externally.

signature-specification.md
```diff
@@ -120,125 +120,105 @@ Notary v2 requires the signature envelope to support the following signed attrib
...
+      'timestamp': << TimeStampToken >>,
+      / x5chain / 33: [
+        << DER(leafCert) >>,
+        << DER(intermediate CACert) >>,
+        << DER(rootCert) >>
+      ]
     },
-    "signature": "Base64Url( sign( ASCII( <Base64Url(ProtectedHeader)>.<Base64Url(JWSPayload)> )))"  
-}
+    / payload / << descriptor >>,
```
briankr-ms, line 216: If detached payloads are not supported, I suggest documenting that explicitly in this spec.

roywill:
>   - **`org.cncf.notary.x509certs.fingerprint.sha256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprint of signing certificate and certificate chain used for signature generation.

If we adopt\moving to a Ledger type system, why would we keep the certificate chain anywhere?  Is this a hold over from threat?

---
Refers to: signature-specification.md:25 in 644c8cd. [](commit_id = 644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2, deletion_comment = False)

roywill:
>             "mediaType": "application/jose+json",

Why Jose?

---
Refers to: signature-specification.md:33 in 644c8cd. [](commit_id = 644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2, deletion_comment = False)

roywill:
> The Notary v2 signature artifact's `org.cncf.notary.x509certs.fingerprint.sha256` annotations key MUST contain the list of SHA-256 fingerprints of certificate and certificate chain used for signing.

Why do we need this?

---
Refers to: signature-specification.md:59 in 644c8cd. [](commit_id = 644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2, deletion_comment = False)

roywill:
> - Signed attributes `v`: The signature metadata that is integrity protected - e.g. signature expiration time, signing time, etc.

Why do we keep the expiration time in its own field.  Is this part of the support for extending life based on Secure TimeStamp data?

---
Refers to: signature-specification.md:67 in 644c8cd. [](commit_id = 644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2, deletion_comment = False)

roywill:
> - Unsigned attributes `u`: This OPTIONAL property represents signature metadata that is not integrity protected - e.g. timestamp, certificates, etc.

In the case of our other scenarios, the key thing that we store in unprotected headers is another signed message.  How much of this is true here?

---
Refers to: signature-specification.md:68 in 644c8cd. [](commit_id = 644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2, deletion_comment = False)

---
Edit from @SteveLasker 

Updated to reflect:

- Unsigned attributes `u`: This OPTIONAL property represents signature metadata that is not integrity protected by this envelope- e.g. timestamp, certificates, etc.
  We anticipate all unsigned attributes contain signed content.

roywill:
>       "io.wabbit-networks.buildId": "123"  // user defined signed attribute.

It would be nice to show something under io.cncf.notary

---
Refers to: signature-specification.md:91 in 644c8cd. [](commit_id = 644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2, deletion_comment = False)

--- 
@SteveLasker  Commenting in an edit:

We could switch to a CNCF project. Within the notary project, we use two companies as representative users
- Wabbit Networks is a small ISV, shipping their software
- Acme Rockets is a small consumer, that consumes public software, such as the `net-monitor:v1` image from Wabbit Networks. 
This is just to be consistent with those users.

roywill:
>    "mediaType": "sbom/example",

Media type is usually text\plain, multipart\related, image\jpg .

Did we really define a type sbom/example?

---
Refers to: signature-specification.md:98 in 644c8cd. [](commit_id = 644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2, deletion_comment = False)

---
Inlined edit from @SteveLasker 

To avoid referencing specific projects, we use `[artifactType]/example`

roywill:
>   Its value should be numeric representing the number of seconds (not milliseconds) since Epoch.

Is it the time it was signed or the time the payload was created?  Thus in a disaster scenario when things come back up is it the time the "operation happened" or the time when we requested a signature?

---
Refers to: signature-specification.md:109 in 644c8cd. [](commit_id = 644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2, deletion_comment = False)

roywill:
>   Its value should be numeric representing the number of seconds since Epoch.

Isn't this also expected to be > than Creation Time?  Or is it truly free form and thus can be earlier than or equal to creation time?

---
Refers to: signature-specification.md:111 in 644c8cd. [](commit_id = 644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2, deletion_comment = False)

roywill:
> - **Certificates**: The ordered collection of X.509 certificates with a signing certificate as the first entry.

Why are all the intermediaries here and does it include the root?
What happens if we have to support multiple signatures?

---
Refers to: signature-specification.md:116 in 644c8cd. [](commit_id = 644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2, deletion_comment = False)

roywill:
> - **Timestamp**:  The time stamp token generated for a given signature.

This as it stands requires a CMS parser to exist to understand it.  The IETF has work on its place to come up with another representation and as such is this something we can postpone?  

---
Refers to: signature-specification.md:117 in 644c8cd. [](commit_id = 644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2, deletion_comment = False)

roywill:
> - Signed attributes `v`: The signature metadata that is integrity protected - e.g. signature expiration time, signing time, etc.

The signature was valid at the time pen went to paper.  Really what you are alluding to is when support stops or after which it is impossible to check revocation.  Which is it?

---
In reply to: [1067417710](https://github.com/notaryproject/notaryproject/pull/139#issuecomment-1067417710) [](http://example.com/codeflow?ancestors=1067417710,1067417710)

---
Refers to: signature-specification.md:67 in 644c8cd. [](commit_id = 644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2, deletion_comment = False)

briankr-ms:
> > * **Timestamp**:  The time stamp token generated for a given signature.
> 
> This as it stands requires a CMS parser to exist to understand it. The IETF has work on its place to come up with another representation and as such is this something we can postpone?
> 
> Refers to: signature-specification.md:117 in [644c8cd](https://github.com/notaryproject/notaryproject/commit/644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2). [](commit_id = [644c8cd](https://github.com/notaryproject/notaryproject/commit/644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2), deletion_comment = False)

An RFC3161 countersignature is specifically useful as independent evidence that a signature was created (or at least existed) within the validity period of an x.509 certificate, for the purpose of determining whether it can be considered valid after the signing certificate's expiration.

I think this is distinct from other time-related claims (i.e. "signingtime" or "expiry") that are headers protected by the same key that signed the payload, and which are potentially meaningful in a non-x.509 trust model where key and signature validity are not associated with a certificate. So while there is some appearance of redundancy here, ultimately they serve different purposes.

Given that x.509 already implies the ASN.1 overhead (if not CMS), and that RFC3161 countersignatures are already commonly used to establish validity after a certificate's expiration, I think it's reasonable to use the existing standard for this purpose, and that this is preferable to omitting it until a CBOR variant is standardized/widely available. Without it, you run into scenarios where an artifact is signed shortly before a certificate's expiration and is only "valid" for a brief period. If and when the trust model moves away from x.509, this field is likely no longer needed or useful.

dlorenc:
> If we adopt\moving to a Ledger type system, why would we keep the certificate chain anywhere? Is this a hold over from threat?

Is there any documentation on the Ledger system?

kaywilliams:
> Is there any documentation on the Ledger system?

The reference is to an early proposal in the IETF. 

[https://datatracker.ietf.org/doc/draft-birkholz-scitt-architecture/](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fdatatracker.ietf.org%2Fdoc%2Fdraft-birkholz-scitt-architecture%2F&data=04%7C01%7Ckayw%40microsoft.com%7Cb4a5861e8f004178495008da06b83947%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637829685549522253%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=f1TerknB0Jo%2FEm%2FK3lQRivb9xbdCSyNMVPG%2B6SP41Us%3D&reserved=0)
[https://datatracker.ietf.org/doc/draft-birkholz-scitt-receipts/](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fdatatracker.ietf.org%2Fdoc%2Fdraft-birkholz-scitt-receipts%2F&data=04%7C01%7Ckayw%40microsoft.com%7Cb4a5861e8f004178495008da06b83947%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637829685549572227%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=Ts9Y3glmJZKdiO141BRit58tUu1ZYSRhUqbRrrMJQFE%3D&reserved=0)
https://github.com/ietf-scitt

Discussion is happening on the ietf-scitt mailing list:
[SCITT Info Page (ietf.org)](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.ietf.org%2Fmailman%2Flistinfo%2Fscitt&data=04%7C01%7Ckayw%40microsoft.com%7Cb4a5861e8f004178495008da06b83947%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637829685549522253%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=nL2GfNrpY6mqBrqRVqe5jcKgYkJJ2rRZybVN%2FRyoXW8%3D&reserved=0)

signature-specification.md
shizhMSFT: @priteshbandi Should we move `Certificates` attribute from _Unsigned Attributes_ to _Signed Attributes_?

shizhMSFT:
@priteshbandi @SteveLasker Could you also take a look at @roywill 's comment since his comment is mainly for the `signature specification` itself and is not specific to this PR?

shizhMSFT:
>             "mediaType": "application/jose+json",

updated to `cose`.

---
In reply to: [1067416112](https://github.com/notaryproject/notaryproject/pull/139#issuecomment-1067416112) [](http://example.com/codeflow?ancestors=1067416112,1067416112)

---
Refers to: signature-specification.md:33 in 644c8cd. [](commit_id = 644c8cdcc702be5f18f2e5c0ee8ae9d5486cf0c2, deletion_comment = False)

letmaik:
> Highlight: Unlike JWS, `alg` is not a required header in COSE. Therefore, we can remove it completely, making the signature envelope more secure.

Keeping `alg` in the header is useful when relying on key discovery systems that don't bind the alg to the key itself. An example is DID where `alg` can be included in the JWK object. Since different DID methods have more or less guarantees on the integrity of the resolved DID document (DID documents are unsigned), e.g. when using ledgers vs. fetching a URL, it makes sense to reduce reliance on additional metadata as much as possible.

shizhMSFT:
> > Highlight: Unlike JWS, `alg` is not a required header in COSE. Therefore, we can remove it completely, making the signature envelope more secure.
> 
> Keeping `alg` in the header is useful when relying on key discovery systems that don't bind the alg to the key itself. An example is DID where `alg` can be included in the JWK object. Since different DID methods have more or less guarantees on the integrity of the resolved DID document (DID documents are unsigned), e.g. when using ledgers vs. fetching a URL, it makes sense to reduce reliance on additional metadata as much as possible.

The `alg` header is OPTIONAL and many COSE implementations still set values to it. Unlike JWS, the verification won't fail if `alg` does not present.

SteveLasker:
Proposing we limit the scope of this effort to `COSE_SIGN1`, deferring `COSE_SIGN` to a future date.

signature-specification.md
```diff
@@ -120,125 +120,105 @@ Notary v2 requires the signature envelope to support the following signed attrib
...
 
-- **`subject`**(*descriptor*): A REQUIRED top-level node consisting of the manifest that needs to be integrity protected.
-  Please refer [Payload](#payload) section for more details.
-- **`iat`**(*number*): The REQUIRED property Issued-at(`iat`) identifies the time at which the signature was issued.
-- **`exp`**(*number*): This OPTIONAL property contains the expiration time on or after which the signature must not be considered valid.
+- **`crit`** (*array of integers or strings*): This REQUIRED property (label: `2`) lists the headers that implementation MUST understand and process.
+  The array MUST contain `3` (`cty`), and `signingtime`. If `expiry` is presented, the array MUST also contain `expiry`.
+- **`cty`** (*string*): The REQUIRED property content-type (label: `3`) is used to declare the media type of the secured content (the payload).
+- **`signingtime`** (*datetime*): The REQUIRED property identifies the time at which the signature was generated.
+- **`expiry`** (*datetime*): This OPTIONAL property contains the expiration time on or after which the signature must not be considered valid.
```
letmaik, line 153: Can you add **`iss`** (*string*) to this list as optional property indicating the issuer of the signature? If it follows a well-known scheme like DID (starting with `did:`) then it may be used for key discovery, together with `kid` which is also missing here.

shizhMSFT: Currently, Notary v2 only supports certs. Other key discovery systems like `kid` or `iss` or others are not supported, and are not covered in this PR. This PR focuses on changing what we have for JWS to COSE and nothing else. BTW, `iss` makes it more like a CWT.

@letmaik Could you follow up #95 or open a new issue in case we need to support other key discovery systems so that the community can review?

SteveLasker:
@shizhMSFT, you’re correct this PR is focused on COSE and x509. And, we’re looking to use DiD for where to discover the public key. Knowing we’ll add other key and identity types, is there anything we would add now to make that support less of a breaking change?

SteveLasker:
From today's Notary call, we agreed to keep both the current JWT, adding COSE. 

We’re hoping to land the COSE format for the initial release, but we’ve said notary will support multiple signature formats over time, for instance, different governments (China has been the common example, but also support for newer fomats over time). If we add multiples now, we validate the spec supports multiples. If we land COSE for v1, we can pull JWT out, knowing the spec supports multiple formats.

SteveLasker:
To move the spec forward, aleviating Shiwei to focus on the go-cose work, suggesting we merge this into [cose-envelope](https://github.com/notaryproject/notaryproject/tree/cose-envelope) branch.
Please LGTM as the PR target was changed.

SteveLasker (approved):
LGTM to merge into the cose-envelope branch, and we'll continue to iterate the feedback there.

gokarnm:
LGTM to merge into the cose-envelope branch, to unblock work. @priteshbandi  and me have yet to review this PR completely.

gokarnm (approved):

signature-specification.md
SteveLasker: @briankr-ms , @shizhMSFT what's the call here? I can make changes if we can conclude on a decision.

SteveLasker:
> > The Notary v2 signature artifact's `org.cncf.notary.x509certs.fingerprint.sha256` annotations key MUST contain the list of SHA-256 fingerprints of certificate and certificate chain used for signing.
> 
> Why do we need this?
> 

@shizhMSFT, @roywill, @briankr-ms what's the final call here?
Do we need an issue to track this?

SteveLasker:
> Signed attributes v: The signature metadata that is integrity protected - e.g. signature expiration time, signing time, etc.

> Why do we keep the expiration time in its own field. Is this part of the support for extending life based on Secure TimeStamp data?

Need a call here: @roywill, @shizhMSFT, @gokarnm

SteveLasker:
> Its value should be numeric representing the number of seconds (not milliseconds) since Epoch.

> Is it the time it was signed or the time the payload was created? Thus in a disaster scenario when things come back up is it the time the "operation happened" or the time when we requested a signature?

@shizhMSFT, @gokarnm, @priteshbandi, what would you like to make this?

SteveLasker:
> Its value should be numeric representing the number of seconds since Epoch.

> Isn't this also expected to be > than Creation Time? Or is it truly free form and thus can be earlier than or equal to creation time?

@shizhMSFT, @gokarnm, @priteshbandi, same question here ^

SteveLasker:
@roywill 
> Certificates: The ordered collection of X.509 certificates with a signing certificate as the first entry.
> Why are all the intermediaries here and does it include the root?
> What happens if we have to support multiple signatures?

Notary doesn't support multiple signatures in the same payload.  You can create multiple detached signatures, each with their own `subject` and each with their own claim. 

However, this does call out "signing certificate", which we've since said is not required. Should this be updated to say root certificate as the first entry?

signature-specification.md
```diff
@@ -17,7 +17,7 @@ The signature manifest has an artifact type that specifies it's a Notary V2 sign
 
 - **`artifactType`** (*string*): This REQUIRED property references the Notary version of the signature: `application/vnd.cncf.notary.v2.signature`.
 - **`blobs`** (*array of objects*): This REQUIRED property contains collection of only one [artifact descriptor](https://github.com/oras-project/artifacts-spec/blob/main/descriptor.md) referencing signature envelope.
-  - **`mediaType`** (*string*): This REQUIRED property contains media type of signature envelope blob. The supported value is `application/jose+json`
+  - **`mediaType`** (*string*): This REQUIRED property contains media type of signature envelope blob. The supported value is `application/cose`.
```
gokarnm, line 20: Is the intent to disregard any optional parameter e.g. `application/cose; cose-type="cose-sign1"` ? Notation should probably include `code-type` param to be more specific.

signature-specification.md
```diff
@@ -120,125 +120,105 @@ Notary v2 requires the signature envelope to support the following signed attrib
...
-         "key1": "value1",
-         "key2": "value2",
-         ...
-      }
-   },
-   "iat": 1234567891000,
-   "exp": 1234567891011
+  / crit / 2: [
+    / cty / 3,
+    'signingtime',
```
gokarnm, line 138: Is there a reason to use headers of type string instead of int? Do we plan to followup and include these in IANA?

https://www.iana.org/assignments/cose/cose.xhtml#header-parameters

signature-specification.md
gokarnm: @briankr-ms could you provide more details for cert substitution attack. Notation will use x5chain in conjunction with trust store configured by the user. As part of verification, the cert chain is checked to terminate in one of the certs in trust store. @shizhMSFT  are we using x5t/thumbprint in any other context, other than signature manifest attributes, for matching signature to policy for evaluation?

gokarnm: @shizhMSFT The certs in chain are already signed by the next cert in the chain, and is expected to be authenticated against the trust store. Open to discussing if moving it to signed attributes protects us against certain attacks.

signature-specification.md
```diff
@@ -120,125 +120,105 @@ Notary v2 requires the signature envelope to support the following signed attrib
...
 Unless explicitly specified as OPTIONAL, all fields are required.
-Also, there shouldn’t be any additional fields other than ones specified in JWSPayload, ProtectedHeader, and UnprotectedHeader.
 
-**JWSPayload a.k.a. Claims**:
-Notary v2 is using one private claim (`notary`) and two public claims (`iat` and `exp`).
-An example of the claim is described below
+**Payload**: COSE signs the payload as defined in the [Payload](#payload) section. Detached payloads are not supported.
 
-```json
+**ProtectedHeaders**: Notary v2 supports the following protected headers. Other header fields can be included but will be ignored.
```
gokarnm, line 132: I'd explicitly mention that `alg` if included will not be considered by Notation during verification.

gokarnm:
> > > The Notary v2 signature artifact's `org.cncf.notary.x509certs.fingerprint.sha256` annotations key MUST contain the list of SHA-256 fingerprints of certificate and certificate chain used for signing.
> > 
> > 
> > Why do we need this?
> 
> @shizhMSFT, @roywill, @briankr-ms what's the final call here? Do we need an issue to track this?

Multiple signatures can be associated with and OCI artifact, which are can be queried through the ORAS references API. This field is used to filter down signatures that will be considered for signature verification against the customer configured Trust store.

More context in #131

gokarnm:
> > Signed attributes v: The signature metadata that is integrity protected - e.g. signature expiration time, signing time, etc.
> 
> > Why do we keep the expiration time in its own field. Is this part of the support for extending life based on Secure TimeStamp data?
> 
> Need a call here: @roywill, @shizhMSFT, @gokarnm

The signature can be generated by ephemeral keys, the optional signature expiry field allows the signatures to be valid beyond key expiry without relying on TSA.

Can be renamed to signature-expiry. Clarify that signature-expiry can be shorter or larger than key expiry.

signature-specification.md
gokarnm: TODO: Capture the intent for x5chain.

PR #140, opened 2022-03-09
CWT Signature Envelope

shizhMSFT:
This PR shows a CWT alternative to COSE (#139) without the use of custom COSE headers except `timestamp`.


SteveLasker:
Thanks @shizhMSFT for posting this as an option. I think we've agreed we want to CWT.
ping @henkbirkholz for some details, but suggest we close this one and focus on #139

PR #147, opened 2022-04-12
Updates to signature specification

gokarnm:
* Moves signed attributes out of payload
* Defines new signed attributes and marks them as critical/informational
* Introduces signature scheme to support different trust models

cc: [roywill](https://github.com/roywill) for review.


signature-specification.md
```diff
@@ -84,20 +84,42 @@ Notary v2 requires Payload to be the content **descriptor** of the subject manif
...
+- **Client identifier**: The version of a client (e.g. Notation) that produced the signature. This is an OPTIONAL claim. It uses the following format `{client}/{version}` e.g. “notation/1.0.0”. This claim in intended to be used for diagnostic and troubleshooting purposes.
 
 ### Unsigned Attributes
 
-- **Certificates**: The ordered collection of X.509 certificates with a signing certificate as the first entry.
+These attributes are considered unsigned with respect to the signing key that generates the signature. These attributes are typically signed by a third party (e.g. CA, TSA).
+
+- **Certificate Chain**: This property contains the list of X.509 certificate or certificate chain. This is a REQUIRED attribute. The certificate chain is authenticated using against a trust store as part of signature validation.
+- **TSA counter signature** : The time stamp token generated for a given signature. Only [RFC3161](ietf-rfc3161) compliant TimeStampToken are supported. This is an OPTIONAL attribute.
```
roywill, line 122: I wonder if you should call out the work by the IETF to formulate a new standard here and though this may be the future it is not standardized at this time.

signature-specification.md
```diff
@@ -148,6 +170,12 @@ The client implementation can use the aforementioned `mediaType` to parse the si
 [COSE_Sign1_Tagged](./signature-envelope-cose.md) signature envelope versioning is achieved by the `cty` field in ProtectedHeaders.  
 [JWS JSON serialization](./signature-envelope-jwt.md) signature envelope versioning is achieved by the `cty` field in ProtectedHeaders.
 
+## Appendix
+
+### Signing time
+
+The signing time denotes the time at which the signature was generated. A X509 certificate has a defined lifetime during which it can be used to generate signatures. Signatures generated after the certificate expires are considered invalid. A trusted timestamp allows a verifier to determine if the signature was generated when the certificate was valid. It also allows a verifier to determine if a signature be treated as valid when a certificate is revoked, if the certificate was revoked after the signature was generated. In the absence of a trusted timestamp, signatures are considered invalid after certificate expires, and all signatures are considered revoked when a certificate is revoked.
```
roywill: This means that the signing time must be greater than the expiration date of the certificate or do we support greater than or equal to symantics?

signature-specification.md
gokarnm: Were you referring to the [COSE x.509](https://datatracker.ietf.org/doc/html/draft-ietf-cose-x509-08) or some standard related to TSA signature? I intended this spec to be signature format agnostic, we'll cover JWS/COSE specific mapping of these attributes in their specific doc - e.g. [Notary v2 COSE](https://github.com/notaryproject/notaryproject/blob/cose-envelope/signature-envelope-cose.md).

signature-specification.md
gokarnm: Updated this section to be more explicit. Let me know if more details are required.

```suggestion
The signing time denotes the time at which the signature was generated. A X509 certificate has a defined [validity](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.5) during which it can be used to generate signatures. The signing time must be greater than or equal to certificate's `notBefore` attribute, and signing time must be less than or equal to certificate's `notAfter` attribute. Signatures generated after the certificate expires are considered invalid. A trusted timestamp allows a verifier to determine if the signature was generated when the certificate was valid. It also allows a verifier to determine if a signature be treated as valid when a certificate is revoked, if the certificate was revoked after the signature was generated. In the absence of a trusted timestamp, signatures are considered invalid after certificate expires, and all signatures are considered revoked when a certificate is revoked.
```

signature-specification.md
```diff
@@ -35,22 +34,23 @@ Each Notary signature artifact refers to a signature envelope blob.
...
 
 The Signature Envelope is a standard data structure for creating a signed message.
 A signature envelope consists of the following components:
 
-- Payload `m`: The data that is integrity protected - e.g. descriptor of the artifact being signed.
+- Payload `p`: The data that is integrity protected - e.g. descriptor of the artifact being signed.
 - Signed attributes `v`: The signature metadata that is integrity protected - e.g. signature expiration time, creation time, etc.
-- Unsigned attributes `u`: This OPTIONAL property represents signature metadata that is not integrity protected by this envelope- e.g. timestamp, certificates, etc.
-  We anticipate all unsigned attributes contain signed content.
+- Unsigned attributes `u`: These attributes are considered unsigned with respect to the signing key that generates the signature. These attributes are typically signed by a third party (e.g. CA, TSA).
```
shizhMSFT: > These attributes are typically signed by a third party (e.g. CA, TSA).

Third parties won't sign those attributes. Most likely, we will put counter signatures in those attributes.

signature-specification.md
```diff
@@ -35,22 +34,23 @@ Each Notary signature artifact refers to a signature envelope blob.
...
-The Notary v2 signature artifact's `io.cncf.notary.x509certs.fingerprint.sha256` annotations key MUST contain the list of SHA-256 fingerprints of certificate and certificate chain used for signing.
+The Notary v2 signature artifact's `io.cncf.notary.x509.fingerprint.sha256` annotations key MUST contain the list of SHA-256 fingerprints of certificate and certificate chain used for signing.
 
 ## Signature Envelope
 
 The Signature Envelope is a standard data structure for creating a signed message.
 A signature envelope consists of the following components:
 
-- Payload `m`: The data that is integrity protected - e.g. descriptor of the artifact being signed.
+- Payload `p`: The data that is integrity protected - e.g. descriptor of the artifact being signed.
```
shizhMSFT: Although `p` refers to payload, it is more common to use `m` for message in cryptography in terms of signature schemes.

signature-specification.md
```diff
@@ -35,22 +34,23 @@ Each Notary signature artifact refers to a signature envelope blob.
...
-A signature envelope is `e = {m, v, u, s}` where `s` is signature.
+A signature envelope is `e = {p, v, u, s}` where `s` is signature.
 
-Notary v2 supports multiple envelope formats, including:
+This specification defines the set of signed and unsigned attributes that make up a valid The Notary v2 signature. This specification aims to be be agnostic of signature envelope format (e.g. COSE, JWS), details of encoding the envelope in a specific signature envelope format are covered in in separate specs.
+
+Notary v2 supports the following envelope formats:
 
 - [COSE Sign1](./signature-envelope-cose.md)
 - [JWT](./signature-envelope-jwt.md)
```
shizhMSFT: nit: It is actually `JWS`.

signature-specification.md
shizhMSFT: BTW, `p` is usually stand for prime and `\rho` for policy.

signature-specification.md
```diff
@@ -84,20 +84,42 @@ Notary v2 requires Payload to be the content **descriptor** of the subject manif
...
+
+`notary.signer.x509` - This signature scheme defines the trust model that uses traditional X509 based PKI. Users owns signing keys and use CA issued certificates to represent identity.
+
+`notary.signingservice.x509` - This signature scheme defines the trust model for a signing service that uses X509 based PKI. In this model, a trusted signing service manages the keys of behalf of the user, and generates signatures. The end user does not have direct access to signing keys.
```
shizhMSFT, line 93: Could we replace the term `trust model` with `security model`, which is a more appropriate academic term?

signature-specification.md
shizhMSFT: > It also allows a verifier to determine if a signature be treated as valid when a certificate is revoked, if the certificate was revoked after the signature was generated.

If the certificate is revoked, the signature should also be revoked even if it is generated before the revocation.

Here's a typical scenario: Eve hacks Alice's computer and get Alice's private key. Eve uses Alice's key to generate a signature. After a few hours, Alice detects the intrusion and revoked her certificate. Apparently, Alice wants the signatures generated by Eve to be revoked, even with the cost of revoking Alice's genuine signatures.

signature-specification.md
shizhMSFT: From the `Standard attributes` section, it seems that the signing time is trusted if it is from a signing service, not from a signer.

Then I am confused. It seems `notary.signer.x509` is a subset of `notary.signingservice.x509` as a signer can be a signing service.

For example, Alice is a signer. The signing time of her signatures are not trusted. However, Alice can set up a signing service and pretends to be a signing service. Then the signing time of her signatures are trusted. It does not make sense.

Readers may argue that signer and signing service are distinguished by the pinned trust by verifier-side policies. This is actually a paradox. If we trust the signing time using verifier-side policies based on the signer identity, why should we bother the signer type (i.e. signer or signing service)?

signature-specification.md
gokarnm: I was trying to convey that these attributes are typically not "unsigned", they are signed by CA (for cert chain) and by TSA for TSA counter signature. I can change the wording. The previous version had the wording *"We anticipate all unsigned attributes contain signed content"*

```suggestion
- Unsigned attributes `u`: These attributes are not signed by the signing key that generates the signature. We anticipate unsigned attributes contain content that may be signed by an different party e.g. Certificate chain signed by a CA, or TSA countersignature signed by the TSA.
```

signature-specification.md
gokarnm: Changed based on your feedback, readers not familiar with that convention were confused by it.

```suggestion
- Payload/Message `m`: The data that is integrity protected - e.g. descriptor of the artifact being signed.
```

signature-specification.md
gokarnm: Agreed, I made that feedback on the last version by Steve and it got missed, I'll address in next update.

signature-specification.md
gokarnm: We can change `trust model` to `security model`, I preferred `trust model` as each model deals with how we establish trust (mainly authenticity), `security model` seemed much broader, maybe including other things like signing algorithm which are addressed by other part of the spec. 

You are right, Alice a signer can generate a signature that includes `trusted signing time`, but it's not a valid attribute for `notary.signer.x509` scheme. Alice may try to impersonate a signing service by generating the signature with `notary.signingservice.x509` schema attribute, but the signature will not validate. The trust store distinguishes between trusted roots for CAs (used for `notary.signer.x509`) and Signing service (used for `notary.signingservice.x509`), these roots are stored under different location. I'm plan to propose `trust_store\x509\ca_roots` and `trust_store\x509\signing_service_roots` in trust store, trust policy PR.

signature-specification.md
gokarnm: This is usually the signer's choice. Alice may request the CA to revoke the certificate outright (all signatures are considered invalid), or from a particular date (e.g. X.509 cert [InvalidityDate](https://datatracker.ietf.org/doc/html/rfc5280#section-5.3.2)). This allows the signer to dictate which behavior they want, `InvalidityDate` may not be widely supported in verification environments, I've seen few examples which honor it.

gokarnm:
@roywill @shizhMSFT made minor changes to address your feedback. Let me know if more changes are required.

SteveLasker (approved):
LGTM to merge and continue iteration

signature-specification.md
sajayantony: Having some samples values and/or a scenario description in using these might be helpful as per discussion with @shizhMSFT.

PR #148, opened 2022-04-20
COSE envelope signature spec updates

SteveLasker:
Tracking the merger of the [cose-envelope](https://github.com/notaryproject/notaryproject/tree/cose-envelope) branch into main.


signature-envelope-cose.md
```diff
@@ -0,0 +1,134 @@
+# COSE Sign1 Envelope Serialization
+
+In COSE ([rfc8152](https://datatracker.ietf.org/doc/html/rfc8152)), data is stored as either payload or headers (protected and unprotected).
+Notary v2 supports [COSE_Sign1_Tagged](https://datatracker.ietf.org/doc/html/rfc8152#section-4.2) as a signature envelope with some additional constraints on the header fields.
+
+A JWS based signature will be persisted with a blob `mediaType` of `"application/cose"`
```
iamsamirzon: Typo: I believe this should be COSE instead of JWS on this line

signature-envelope-jws.md
```diff
@@ -0,0 +1,142 @@
...
+      }
+   },
+   "iat": 1234567891000,
+   "exp": 1234567891011
+}
+```
+
+The payload contains the subject manifest and other attributes that have to be integrity protected.
+
+- **`subject`**(*descriptor*): A REQUIRED top-level node consisting of the manifest that needs to be integrity protected.
```
letmaik: Inconsistent with above, which says `"notary"`.

signature-envelope-jws.md
```diff
@@ -0,0 +1,142 @@
...
+        "io.cncf.notary.x509certs.fingerprint.sha256": "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\" \"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
+    }
+}
+```
+
+Unless explicitly specified as OPTIONAL, all fields are required.
+Also, there shouldn’t be any additional fields other than ones specified in JWSPayload, ProtectedHeader, and UnprotectedHeader.
+
+**JWSPayload a.k.a. Claims**:
+Notary v2 is using one private claim (`notary`) and two public claims (`iat` and `exp`).
```
letmaik: How is "public" and "private" defined? Isn't this a notion of JWTs?

signature-envelope-jws.md
letmaik: What is a "node"? Probably this should be "property"?

signature-envelope-jws.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+The payload contains the subject manifest and other attributes that have to be integrity protected.
+
+- **`subject`**(*descriptor*): A REQUIRED top-level node consisting of the manifest that needs to be integrity protected.
+  Please refer [Payload](#payload) section for more details.
+- **`iat`**(*number*): The REQUIRED property Issued-at(`iat`) identifies the time at which the signature was issued.
+- **`exp`**(*number*): This OPTIONAL property contains the expiration time on or after which the signature must not be considered valid.
+
+To leverage JWS claims validation functionality already provided by libraries, we have defined `iat`, `exp` as top-level nodes.
```
letmaik: Again, "nodes" -> "properties".

letmaik: Which libraries? JWS libraries or JWT libraries?

signature-envelope-jws.md
```diff
@@ -0,0 +1,142 @@
...
+}
+```
+
+- **`alg`**(*string*): This REQUIRED property defines which algorithm was used to generate the signature.
+  JWS needs an algorithm(`alg`) to be present in the header, so we have added it as a protected header.
+- **`cty`**(*string*): The REQUIRED property content-type(cty) is used to declare the media type of the secured content(the payload).
+  This will be used to version different variations of JWS signature.
+  The supported value is `application/vnd.cncf.notary.v2.jws.v1`.
+- **`crit`**(*array of strings*): This REQUIRED property lists the headers that implementation MUST understand and process.
+  The value MUST be `["cty"]`.
```
letmaik: This makes no sense. `"cty"` is a standard JOSE header parameter and hence mustn't be included in `"crit"`. The spec says:
> Producers MUST NOT include Header Parameter names defined by this specification [...] Recipients MAY consider the JWS to be invalid if the critical list contains any Header Parameter names defined by this specification.

signature-envelope-jws.md
```diff
@@ -0,0 +1,142 @@
...
+- **`iat`**(*number*): The REQUIRED property Issued-at(`iat`) identifies the time at which the signature was issued.
+- **`exp`**(*number*): This OPTIONAL property contains the expiration time on or after which the signature must not be considered valid.
+
+To leverage JWS claims validation functionality already provided by libraries, we have defined `iat`, `exp` as top-level nodes.
+
+**ProtectedHeaders**: Notary v2 supports only three protected headers: `alg`, `cty`, and `crit`.
+
+```jsonc
+{
+    "alg": "RS256",
```
letmaik: Example `"alg"` doesn't follow own restrictions declared later on. Should be `"PS256"`.

signature-envelope-jws.md
```diff
@@ -0,0 +1,142 @@
...
+    "x5c": ["<Base64(DER(leafCert))>", "<Base64(DER(intermediateCACert))>", "<Base64(DER(rootCert))>"]
+}
+```
+
+- **`timestamp`** (*string*): This OPTIONAL property is used to store time stamp token.
+  Only [RFC3161]([rfc3161](https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2)) compliant TimeStampToken are supported.
+- **`x5c`** (*array of strings*): This REQUIRED property contains the list of X.509 certificate or certificate chain([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the JWS.
+  The certificate containing the public key corresponding to the key used to digitally sign the JWS MUST be the first certificate.
+
+- **`timestamp`** (*string*): This OPTIONAL property is used to store time stamp token.
```
letmaik: Duplicated.

signature-envelope-jws.md
```diff
@@ -0,0 +1,142 @@
...
+  Only [RFC3161]([rfc3161](https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2)) compliant TimeStampToken are supported.
+- **`x5c`** (*array of strings*): This REQUIRED property contains the list of X.509 certificate or certificate chain([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the JWS.
```
letmaik: Duplicated.

signature-envelope-jws.md
```diff
@@ -0,0 +1,142 @@
...
+```jsonc
+{
+    "timestamp": "<Base64(TimeStampToken)>",
+    "x5c": ["<Base64(DER(leafCert))>", "<Base64(DER(intermediateCACert))>", "<Base64(DER(rootCert))>"]
+}
+```
+
+- **`timestamp`** (*string*): This OPTIONAL property is used to store time stamp token.
+  Only [RFC3161]([rfc3161](https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2)) compliant TimeStampToken are supported.
+- **`x5c`** (*array of strings*): This REQUIRED property contains the list of X.509 certificate or certificate chain([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the JWS.
```
letmaik: @briankr-ms Just a ping, in case you had further comments to the earlier discussion on putting x5c in protected vs unprotected. See https://github.com/notaryproject/notaryproject/pull/139#discussion_r824276018.

signature-envelope-jws.md
```diff
@@ -0,0 +1,142 @@
...
+            "size": 32654
+        }
+    ],
+    "subject": {
+        "mediaType": "application/vnd.oci.image.manifest.v1+json",
+        "digest": "sha256:73c803930ea3ba1e54bc25c2bdc53edd0284c62ed651fe7b00369da519a3c333",
+        "size": 16724
+    },
+    "annotations": {
+        "io.cncf.notary.x509certs.fingerprint.sha256": "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\" \"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
```
letmaik: Inconsistent with `signature-specification.md`: `io.cncf.notary.x509.fingerprint.sha256`

signature-specification.md
```diff
@@ -1,256 +1,125 @@
...
+  - **`mediaType`** (*string*): This REQUIRED property contains media type of signature envelope blob. The currently supported values are  `application/cose` and  `application/jose+json`.
 - **`subject`** (*descriptor*): A REQUIRED artifact descriptor referencing the signed manifest, including, but not limited to image manifest, image index, oras-artifact manifest.
 - **`annotations`** (*string-string map*): This REQUIRED property contains metadata for the artifact manifest.
   It is being used to store information about the signature.
   Keys using the `io.cncf.notary` namespace are reserved for use in Notary and MUST NOT be used by other specifications.
-  - **`io.cncf.notary.x509certs.fingerprint.sha256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprint of signing certificate and certificate chain used for signature generation.
+  - **`io.cncf.notary.x509.fingerprint.sha256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprint of signing certificate and certificate chain used for signature generation.
```
letmaik: Not clear how the fingerprints are computed exactly.

signature-specification.md
```diff
@@ -1,256 +1,125 @@
...
-    "payload": "<Base64Url(JWSPayload)>",
-    "protected": "<Base64Url(ProtectedHeaders)>",
-    "header": {
-        "timestamp": "<Base64(TimeStampToken)>",
-        "x5c": ["<Base64(DER(leafCert))>", "<Base64(DER(intermediateCACert))>", "<Base64(DER(rootCert))>"]
-    },
-    "signature": "Base64Url( sign( ASCII( <Base64Url(ProtectedHeader)>.<Base64Url(JWSPayload)> )))"  
-}
-```
+- Any claims MUST NOT be stored/appended in the payload itself, as the payload is only parsed and processed once the signature has been verified (signature is valid, and from a trusted key) and trust is established.
```
letmaik: Wouldn't `"exp"` also fall in this category? In `signature-envelope-jws.md` it's part of the payload, right?

signature-specification.md
```diff
@@ -1,256 +1,125 @@
...
+- Specific claims can be either required or optional.
+- Claims that if present, MUST be processed by a verifier MUST be marked as critical.
```
letmaik: I don't understand this requirement. It seems that this excludes certain envelope formats, but see also my other comment in -jws.md about how `"crit": ["cty"]` is going against the JWS spec.

signature-specification.md
```diff
@@ -1,256 +1,125 @@
...
-  | ECDSA on secp256r1 with SHA-256 | ES256             |
-  | ECDSA on secp384r1 with SHA-384 | ES384             |
-  | ECDSA on secp521r1 with SHA-512 | ES512             |
-1. Signing certificate MUST be a valid codesigning certificate.
-1. Only JWS JSON flattened format is supported.
-   See 'Signature Envelope' section.
+#### Standard attributes
 
-## Signature Algorithm Requirements
+- **Notary Signature scheme** (critical): The trust model used by Notary currently supports the following values for this claim `notary.signer.x509` and `notary.signingservice.x509`. Both signature schemes use the same set of claims defined in this document. One difference is that `notary.signer.x509` uses a TSA signature to determine authenticated signing time (timestamp), and `notary.signingservice.x509` uses the trusted signing time claim instead, with optional support for TSA signatures. Signature verification MUST consider the signature to be invalid if any other value is used.
```
letmaik: I couldn't find this claim in the JWS md for example. Shouldn't it be there?

signature-envelope-cose.md
```diff
@@ -0,0 +1,134 @@
...
+   This is the value of the signature property used in the signature envelope.
+
+**Signature Envelope**: The final signature envelope is a `COSE_Sign1_Tagged` object, consisting of Payload, ProtectedHeaders, UnprotectedHeaders, and Signature.
+
+```
+18(
+  [
+    / protected / << {
+      / crit / 2: [
+        / cty / 3,
```
letmaik: COSE spec says "Integer labels in the range of 0 to 7 SHOULD be omitted.". This is less strict than JWS which forbids it, but would probably do the same for consistency.

signature-envelope-cose.md
```diff
@@ -0,0 +1,134 @@
...
+  'expiry': 1234567891011
+}
+```
+
+Note: The above example is represented using the [extended CBOR diagnostic notation](https://datatracker.ietf.org/doc/html/rfc8152#appendix-C).
+
+- **`crit`** (*array of integers or strings*): This REQUIRED property (label: `2`) lists the headers that implementation MUST understand and process.
+  The array MUST contain `3` (`cty`), and `signingtime`. If `expiry` is presented, the array MUST also contain `expiry`.
+- **`cty`** (*string*): The REQUIRED property content-type (label: `3`) is used to declare the media type of the secured content (the payload).
+- **`signingtime`** (*datetime*): The REQUIRED property identifies the time at which the signature was generated.
```
letmaik: Should be more specific on the format. `*datetime*` isn't defined anywhere.

signature-envelope-jws.md
```diff
@@ -0,0 +1,142 @@
...
+   "iat": 1234567891000,
+   "exp": 1234567891011
+}
+```
+
+The payload contains the subject manifest and other attributes that have to be integrity protected.
+
+- **`subject`**(*descriptor*): A REQUIRED top-level node consisting of the manifest that needs to be integrity protected.
+  Please refer [Payload](#payload) section for more details.
+- **`iat`**(*number*): The REQUIRED property Issued-at(`iat`) identifies the time at which the signature was issued.
```
letmaik: Needs more details on the format, i.e. UNIX timestamp in some unit. Same for `exp`.

signature-envelope-cose.md
```diff
@@ -0,0 +1,134 @@
...
+    ```
+    Sig_structure = [
+        / context / 'Signature1',
+        / body_protected / << ProtectedHeaders >>,
+        / external_aad / h'',
+        / payload / << Payload >>,
+    ]
+    ```
+3. Encode `Sig_structure` into a CBOR object as a byte string named `ToBeSigned`.
+4. Compute the signature on the `ToBeSigned` constructed in the previous step by using the signature algorithm defined in the corresponding header element: `alg`.
```
letmaik: `alg` is missing in the protected headers above.

signature-envelope-cose.md
```diff
@@ -0,0 +1,134 @@
...
+    << DER(leafCert) >>,
+    << DER(intermediate CACert) >>,
+    << DER(rootCert) >>
+  ]
+}
+```
+
+Note: `<<` and `>>` are used to notate the byte string resulting from encoding the data item.
+
+- **`timestamp`** (*byte string*): This OPTIONAL property is used to store time stamp token.
```
letmaik: Should be more specific on format, i.e. DER encoding.

signature-envelope-cose.md
letmaik: Any reason this can't use the JWT name `"iat"`? Same for expiry `"exp"`.

SteveLasker: @letmaik, which RFC should we reference here?

SteveLasker: Happy to change to whatever folks prefer

signature-envelope-cose.md
SteveLasker: Can you please propose what you'd like?

signature-envelope-cose.md
BryceDFisher: RFC 7519 specifies this field as number of _seconds_ since epoch, if it's to be aligned with formats used in JWT.

signature-envelope-jws.md
BryceDFisher: RFC 7519 specifies these as _seconds_ since epoch

signature-envelope-cose.md
BryceDFisher: Is this the same as the difference between iat and nbf?  The signature could be generated at one time but not be considered valid until its issued?

SteveLasker:
With the [go-cose v1.0.0-rc.1](https://github.com/veraison/go-cose/releases/tag/v1.0.0-rc.1) release, with a [security review completed](https://github.com/veraison/go-cose/blob/main/reports/NCC_Microsoft-go-cose-Report_2022-05-26_v1.0.pdf), we'll be prioritizing this work. 

There are a few updates that we'll incorporate that include multiple content types that support registry descriptors, hashes of files that may not be store in a registry and direct wrapping of content.

signature-envelope-cose.md
```diff
@@ -0,0 +1,197 @@
+# COSE Sign1 Signature Envelope
+
+This specification implements the [Notary v2 Signature specification](signature-specification.md) using 
+CBOR Object Signing and Encryption (COSE). COSE ([RFC8152](https://datatracker.ietf.org/doc/html/rfc8152)) is a CBOR based envelope format for digital signatures over any type of payload (e.g. CBOR, JSON, binary).
+Notary v2 specifically supports [COSE_Sign1_Tagged](https://datatracker.ietf.org/doc/html/rfc8152#section-4.2) as a signature envelope.
+
+## Storage
+
+A COSE signature envelope will be stored in an OCI registry as a blob, and referenced in the signature manifest as a blob with `mediaType` of `"application/cose"`.
```
thomas-fossati: a more precise media type to use in this context would be `application/cose; cose-type="cose-sign1"` (see [Section 2 of STD 96](https://www.rfc-editor.org/rfc/rfc9052#section-2)).

signature-envelope-cose.md
```diff
@@ -0,0 +1,197 @@
...
+    << DER(leafCert) >>,
+    << DER(intermediate CACert) >>,
+    << DER(rootCert) >>
+  ],
+  'io.cncf.notary.timestampSignature': << TimeStampToken >>,
+  'io.cncf.notary.signingAgent': 'notation/1.0.0'
+}
+```
+
+Note: `<<` and `>>` are used to notate the byte string resulting from encoding the data item.
```
thomas-fossati:
```suggestion
Note: `<<` and `>>` are used to notate the CBOR byte string resulting from encoding the data item.
```

signature-envelope-cose.md
```diff
@@ -0,0 +1,197 @@
...
+
+- **[`x5chain`](https://datatracker.ietf.org/doc/html/draft-ietf-cose-x509-08#section-2)** (*array of bstr*): This REQUIRED parameter (label `33` by [IANA](https://www.iana.org/assignments/cose/cose.xhtml#header-parameters)) contains the ordered list of X.509 certificate or certificate chain ([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the COSE. The certificate chain is represented as an array of certificate value byte string, each certificate in the array is DER encoded. The certificate containing the public key corresponding to the key used to digitally sign the COSE MUST be the first certificate, followed by the intermediate and root certificates in the correct order. Refer [*Certificate Chain* unsigned attribute](signature-specification.md#unsigned-attributes) for more details. Optionally, this header can be presented in the protected header.
```
thomas-fossati:
```suggestion
- **[`x5chain`](https://datatracker.ietf.org/doc/html/draft-ietf-cose-x509-08#section-2)** (*array of bstr*): This REQUIRED parameter (label `33` by [IANA](https://www.iana.org/assignments/cose/cose.xhtml#header-parameters)) contains the ordered list of X.509 certificate or certificate chain ([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the COSE. The certificate chain is represented as an array of certificate, each certificate in the array is DER encoded and then wrapped in a CBOR byte string. The certificate containing the public key corresponding to the key used to digitally sign the COSE MUST be the first certificate, followed by the intermediate and root certificates in the correct order. Refer [*Certificate Chain* unsigned attribute](signature-specification.md#unsigned-attributes) for more details. Optionally, this header can be presented in the protected header.
```

signature-specification.md
```diff
@@ -43,6 +43,7 @@ Signature Manifest Example
 - **`blobs`** (*array of objects*): This REQUIRED property contains collection of only one [artifact descriptor][artifact-descriptor] referencing signature envelope.
   - **`mediaType`** (*string*): This REQUIRED property contains media type of signature envelope blob. Following values are supported
     - `application/jose+json`
+    - `application/cose`
```
thomas-fossati: similar to the suggestion I made above, I think this should be paired by the right `cose-type` parameter.

signature-specification.md
```diff
@@ -165,6 +168,127 @@ These attributes are considered unsigned with respect to the signing key that ge
...
+        "timestamp": "<Base64(TimeStampToken)>",
+        "x5c": ["<Base64(DER(leafCert))>", "<Base64(DER(intermediateCACert))>", "<Base64(DER(rootCert))>"]
+    },
+    "signature": "Base64Url( sign( ASCII( <Base64Url(ProtectedHeader)>.<Base64Url(JWSPayload)> )))"  
+}
+```
+
+**Implementation Constraints**: Notary v2 implementation MUST enforce the following constraints on signature generation and verification:
+
+1. `alg` header value MUST NOT be `none` or any symmetric-key algorithm such as `HMAC`.
```
thomas-fossati: re: prohibition on HMAC.  This may be redundant since I don't think one can use symmetric key algorithms with COSE Sign/Sign1.  Those should be only available for Mac or Mac0 COSEs.

thomas-fossati (commented):
A really nice work. I have left a few flyby comments for your consideration.

signature-envelope-cose.md
letmaik: I think this is redundant since the COSE envelopes are tagged anyway. In my mind this would make client code harder to write because now you need to define how media type parameters are encoded in this context. This is not HTTP (which has its own definition for encoding those).

thomas-fossati: > I think this is redundant since the COSE envelopes are tagged anyway.

Yeah it's redundant.  Whether this hint is worthwhile depends on how processing is implemented.  With the extra information, the COSE payload can be dispatched directly to the right API, whilst without, the implementation needs to peek into the COSE tag to dispatch.

> In my mind this would make client code harder to write because now you need to define how media type parameters are encoded in this context. This is not HTTP (which has its own definition for encoding those).

Not sure I understand what the encoding problem is?  The media parameters that'd be used here are all UTF-8 strings (which both JSON and I-JSON support off-the-shelf) with a-priori known values.

( I have the feeling I am missing something though :-) )

letmaik: In practice, even outside the context of Email or HTTP headers, people use the same syntax as first defined in https://datatracker.ietf.org/doc/html/rfc2045#section-5.1 to encode parameters, especially for protocols/formats where you want to use a single string to encode the media type. 
But it has to be made explicit with a reference that this is what's happening here.
From an implementation point of view, for example in Python, clients would have to do the following to match the media type (note the `parse_header` function name):
```py
import cgi
mediatype, params = cgi.parse_header('application/cose; cose-type="cose-sign1"')
if mediatype == "application/cose" and params["cose-type"] == "cose-sign1":
  # process
```
This is necessary since whitespace becomes a factor now. Previously, a simple string comparison would have been enough: `mediatype == "application/cose"`.
Of course you could prescribe that the media type string must always use a specific fixed string that can be matched by string comparison. There is some chance that people will mess it up though.

I think, given that the ORAS artifact type already says that it is a (single) signature (`"artifactType": "application/vnd.cncf.notary.signature"`), I don't see why it has to be repeated here using the parameter.

dtzar: Agree with letmaik - going to close this comment out.

signature-specification.md
dtzar: [same as above](https://github.com/notaryproject/notaryproject/pull/148/files#r975829734), redundant/unnecessary

signature-envelope-jws.md
dtzar: @letmaik There are no changes to signature-envelope-jws.md at this point, so marking all the ones for this file as resolved.

signature-envelope-cose.md
dtzar: This is present in the spec

signature-envelope-cose.md
dtzar: Its value is the number of seconds from `1970-01-01T00:00Z` in UTC time, commonly known as UNIX timestamp

signature-envelope-jws.md
dtzar: jws not in scope for this PR

signature-envelope-cose.md
dtzar: Its value is the number of seconds from `1970-01-01T00:00Z` in UTC time, commonly known as UNIX timestamp

dtzar: We're consistently using io.cncf.notary.expiry on both and io.cncf.notary.signingTime

SteveLasker:
With the COSE support now in notation, just a reminder to incorporate the cose specs

dtzar (approved):

signature-envelope-cose.md
```diff
@@ -0,0 +1,197 @@
...
+    "annotations": {
+        "io.cncf.notary.x509chain.thumbprint#S256": 
+        "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\",\"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
+    }
+}
+```
+
+## COSE Payload
+
+The COSE envelope contains a [Notary v2 Payload](./signature-specification.md#payload).
```
gokarnm: Would be good to add an example of Notary V2 payload so reader does not have to jump to generic signature spec to see one.

```suggestion
The COSE envelope contains a [Notary v2 Payload](./signature-specification.md#payload).

Example of Notary v2 payload

```jsonc
{
  "targetArtifact": {
    "mediaType": "application/vnd.oci.image.manifest.v1+json",
    "digest": "sha256:73c803930ea3ba1e54bc25c2bdc53edd0284c62ed651fe7b00369da519a3c333",
    "size": 16724,
    "annotations": {
        "io.wabbit-networks.buildId": "123"  // user defined metadata
    }
  }
}
```

```

File: signature-envelope-cose.md [hunk 26]
```diff
@@ -0,0 +1,197 @@
...
+  - Label `1`: `alg`
+  - Label `2`: `crit`
+  - Label `3`: `content type`
+- Additional parameters with collision resistant names.
+  - `io.cncf.notary.signingScheme`
+  - `io.cncf.notary.signingTime`
+  - `io.cncf.notary.authenticSigningTime`
+  - `io.cncf.notary.expiry`
+
+Example with Signing Scheme `notary.x509`
```
gokarnm:
```suggestion
COSE examples in this document use [CBOR diagnostic notation](https://www.rfc-editor.org/rfc/rfc7049#section-6)([examples](https://datatracker.ietf.org/doc/html/rfc8152#appendix-C)).

Example with Signing Scheme `notary.x509`
```

signature-envelope-cose.md
```diff
@@ -0,0 +1,197 @@
...
+## COSE Payload
+
+The COSE envelope contains a [Notary v2 Payload](./signature-specification.md#payload).
+
+## Protected Header
+
+The COSE envelope for Notary v2 uses the following header parameters:
+
+- [Common parameters](https://www.iana.org/assignments/cose/cose.xhtml#header-parameters)
+  - Label `1`: `alg`
```
gokarnm: @shizhMSFT , is `alg` a required header for COSE? We included it in JWS envelope as the JWS spec requires it. If not, can it be ommited and use signature algorithm from signing key/cert as source of truth?

signature-envelope-cose.md
```diff
@@ -0,0 +1,197 @@
...
+- **[`content type`](https://datatracker.ietf.org/doc/html/rfc8152#section-3.1)** (*tstr*): The REQUIRED parameter content type (label `3`) is used to declare the media type of the secured content (the payload). The supported value is `application/vnd.cncf.notary.payload.v1+json`.
+- **`io.cncf.notary.signingScheme`** (*tstr*, critical): This REQUIRED header specifies the [Notary v2 Signing Scheme](./signing-scheme.md) used by the signature. Supported values are `notary.x509` and `notary.x509.signingAuthority`.
+- **`io.cncf.notary.signingTime`** (*date/time*): This header specifies the time at which the signature was generated. This is an untrusted date/time, and therefore not used in trust decisions. Its value is an Epoch-Based Date/Time defined in [RFC 8949](https://datatracker.ietf.org/doc/html/rfc8949#section-3.4.2). This claim is REQUIRED and only valid when signing scheme is `notary.x509`.
```
gokarnm: @shizhMSFT  Is there a reason the RFC 3339 text based date time representation supported by COSE (tag 0) is not used in favor of epoch based representation (tag 1)? More concise representation may be one reason (~ 8 vs 20 bytes), are there other stronger reasons?

signature-envelope-cose.md
```diff
@@ -0,0 +1,197 @@
...
+  'io.cncf.notary.signingAgent': 'notation/1.0.0'
+}
+```
+
+Note: `<<` and `>>` are used to notate the CBOR byte string resulting from encoding the data item.
+
+- **[`x5chain`](https://datatracker.ietf.org/doc/html/draft-ietf-cose-x509-08#section-2)** (*array of bstr*): This REQUIRED parameter (label `33` by [IANA](https://www.iana.org/assignments/cose/cose.xhtml#header-parameters)) contains the ordered list of X.509 certificate or certificate chain ([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the COSE. The certificate chain is represented as an array of certificate, each certificate in the array is DER encoded and then wrapped in a CBOR byte string. The certificate containing the public key corresponding to the key used to digitally sign the COSE MUST be the first certificate, followed by the intermediate and root certificates in the correct order. Refer [*Certificate Chain* unsigned attribute](signature-specification.md#unsigned-attributes) for more details. Optionally, this header can be presented in the protected header.
+  - **TODO** update signature specification to allow chains in protected header
```
gokarnm: What does the implementation use now?

signature-envelope-cose.md
```diff
@@ -0,0 +1,197 @@
...
+In COSE, signature is calculated by constructing the `Sig_structure` for `COSE_Sign1`.
+The process is described below:
+
+1. Encode the protected header into a CBOR object as a byte string named `body_protected`.
+2. Construct the `Sig_structure` for `COSE_Sign1`.
+    ```yaml
+    Sig_structure = [
+        / context / 'Signature1',
+        / body_protected / << ProtectedHeaders >>,
+        / external_aad / h'',
```
gokarnm: What is `external_aad` , it's not defined here.

signature-envelope-cose.md
```diff
@@ -0,0 +1,197 @@
...
+## Signature
+
+In COSE, signature is calculated by constructing the `Sig_structure` for `COSE_Sign1`.
+The process is described below:
+
+1. Encode the protected header into a CBOR object as a byte string named `body_protected`.
+2. Construct the `Sig_structure` for `COSE_Sign1`.
+    ```yaml
+    Sig_structure = [
+        / context / 'Signature1',
```
gokarnm: It'd be good to add detail about `context Signature1`, that the identifying context for COSE_Sign1 is `"Signature1"`.

signature-envelope-cose.md
```diff
@@ -0,0 +1,197 @@
...
+        / body_protected / << ProtectedHeaders >>,
+        / external_aad / h'',
+        / payload / << Payload >>,
+    ]
+    ```
+3. Encode `Sig_structure` into a CBOR object as a byte string named `ToBeSigned`.
+4. Compute the signature on the `ToBeSigned` constructed in the previous step by using the signature algorithm defined in the corresponding header element: `alg`.
```
gokarnm:
```suggestion
4. Compute the signature on the `ToBeSigned` constructed in the previous step by using the signature algorithm of the signing key, which MUST match the corresponding protected header element `alg`.
```

signature-specification.md
```diff
@@ -165,6 +168,127 @@ These attributes are considered unsigned with respect to the signing key that ge
 - **Timestamp signature** : An OPTIONAL counter signature which provides [authentic timestamp](#signing-time)e.g. Time Stamp Authority (TSA) generated timestamp signature. Only [RFC3161](ietf-rfc3161) compliant TimeStampToken are currently supported.
 - **Signing Agent**: An OPTIONAL claim that provides the identifier of the software (e.g. Notation) that produced the signature on behalf of the user. It is an opaque string set by the software that produces the signature. It's intended primarily for diagnostic and troubleshooting purposes, this attribute is unsigned, the verifier MUST NOT validate formatting, or fail validation based on the content of this claim. The suggested format is one or more tokens of the form `{id}/{version}` containing identifier and version of the software, seperated by spaces. E.g. “notation/1.0.0”, “notation/1.0.0 com.example.nv2plugin/0.8”.
 
+### Supported Signature Envelopes
```
gokarnm: This whole section for JWS is redundant, we moved to to ./signature-envelope-jws.md , this is probably because COSE spec was branched before we split up JWS into it's own spec.

signature-specification.md
```diff
@@ -193,6 +317,16 @@ The signing certificate's public key algorithm and size MUST be used to determin
...
+The **signing certificate** MUST meet the following minimum requirements:
+
+- The keyUsage extension MUST be present and MUST be marked critical.
+  The bit positions for digitalSignature MUST be set ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3)).
+- The extKeyUsage extension MUST be present and its value MUST be id-kp-codeSigning ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12)).
+- If the basicConstraints extension is present, the cA field MUST be set false ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9)).
+- The certificate MUST abide by the following key length restrictions:
+- For RSA public key, the key length MUST be 2048 bits or higher.
+  - For ECDSA public key, the key length MUST be 256 bits or higher.
+
```
gokarnm: Same here, this section is redundant. It's covered by and updated version of the content from line 330.

gokarnm (changes requested):
Overall looks good, except for few specific details about COSE envelope and outdated content from merge, which should be resolved.

signature-envelope-cose.md
shizhMSFT: According to the COSE spec, `alg` header MUST be present if there is no externally supplied data.

shizhMSFT: In the notary scenario, there is no externally supplied data. Therefore, `alg` is a required header.

signature-envelope-cose.md
shizhMSFT: COSE is based on CBOR where CBOR standards for **Concise Binary** Object Representation. Concise by default is the principle.

Oppositely, JWS is based on JSON, which is text-based not binary-based. That's why RFC3339 is preferable for JWS.

signature-envelope-cose.md
gokarnm: Could you clarify that a bit more, what does externally supplied data mean in this context? Isn't signing cert's signature alg from certificate chain enough in this case?

signature-envelope-cose.md
shizhMSFT: Currently, it is the same as JWS that chains are from unprotected header.

See https://github.com/notaryproject/notation-core-go/blob/v0.2.0-beta.1/signature/cose/envelope.go#L218

signature-envelope-cose.md
shizhMSFT: The `external_aad` is the COSE external data.

signature-envelope-cose.md
shizhMSFT: The context `Signature1` is fixed for `COSE_Sign1` and is defined by the COSE spec. No one can change it.

gokarnm: Ya, I meant just documenting what `context Signature1` is in this section.

signature-envelope-cose.md
yizha1: @gokarnm  There is already a sentence after the example, see below, so I don't think we need another note before the example.

" Note: The above examples are represented using the [extended CBOR diagnostic notation](https://datatracker.ietf.org/doc/html/rfc8152#appendix-C)."

signature-envelope-cose.md
yizha1: @gokarnm See below my proposal for the 2nd bullet, what do you think?
from 
"Construct the `Sig_structure` for `COSE_Sign1`." 
to 
"Construct the `Sig_structure` for `COSE_Sign1`. The context string `Signature1` is used to identify the context of signatures using the `COSE_Sign1` structure."

signature-envelope-cose.md
yizha1: @gokarnm  Did shiwei answer your questions?

specs/signature-specification.md
```diff
@@ -193,6 +317,16 @@ The signing certificate's public key algorithm and size MUST be used to determin
...
 
+The **signing certificate** MUST meet the following minimum requirements:
+
+- The keyUsage extension MUST be present and MUST be marked critical.
+  The bit positions for digitalSignature MUST be set ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3)).
+- The extKeyUsage extension MUST be present and its value MUST be id-kp-codeSigning ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12)).
+- If the basicConstraints extension is present, the cA field MUST be set false ([RFC-5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9)).
+- The certificate MUST abide by the following key length restrictions:
+- For RSA public key, the key length MUST be 2048 bits or higher.
+  - For ECDSA public key, the key length MUST be 256 bits or higher.
```
priteshbandi: Already covered at https://github.com/notaryproject/notaryproject/blob/main/specs/signature-specification.md#leaf-certificates.

specs/signature-specification.md
```diff
@@ -165,6 +168,127 @@ These attributes are considered unsigned with respect to the signing key that ge
...
+  | RSASSA-PSS with SHA-256         | PS256             |
+  | RSASSA-PSS with SHA-384         | PS384             |
+  | RSASSA-PSS with SHA-512         | PS512             |
+  | ECDSA on secp256r1 with SHA-256 | ES256             |
+  | ECDSA on secp384r1 with SHA-384 | ES384             |
+  | ECDSA on secp521r1 with SHA-512 | ES512             |
+1. Signing certificate MUST be a valid codesigning certificate.
+1. Only JWS JSON flattened format is supported.
+   See 'Signature Envelope' section.
+
```
priteshbandi: These were moved to [signature-envelope-jws.md](https://github.com/notaryproject/notaryproject/blob/main/specs/signature-envelope-jws.md) as part of https://github.com/notaryproject/notaryproject/commit/9cfd6a91b1dfad48886af17edf7cc8235556f550, perhaps you need to rebase changes.

specs/signature-specification.md
```diff
@@ -241,7 +375,9 @@ The client implementation can use the aforementioned `mediaType` to parse the si
 **Q: How will Notary v2 support multiple payload formats?**
 
 **A:** The Signature envelope MUST have a versioning mechanism to support multiple payload formats.
-For [JWS JSON serialization](./signature-envelope-jwt.md) signature envelope, versioning is achieved by the `cty` field in ProtectedHeaders.
+
+- For [JWS JSON serialization](./signature-envelope-jwt.md) signature envelope, versioning is achieved by the `cty` field in ProtectedHeaders.
+- For [COSE_Sign1_Tagged](./signature-envelope-cose.md) signature envelope, versioning is achieved by the `content type` (label: `3`) field in ProtectedHeaders.
```
priteshbandi, line 248:
```suggestion
- For [COSE_Sign1](./signature-envelope-cose.md) signature envelope, versioning is achieved by the `content type` (label: `3`) field in ProtectedHeaders.  
```

specs/signature-specification.md
```diff
@@ -79,6 +80,7 @@ This specification defines the set of signed and unsigned attributes that make u
 Notary v2 supports the following envelope formats:
 
 - [JWS](./signature-envelope-jws.md)
+- [COSE Sign1](./signature-envelope-cose.md)
```
priteshbandi, line 83:
```suggestion
- [COSE_Sign1](./signature-envelope-cose.md)
```

signature-envelope-cose.md
gokarnm: Some content describing this field and if/how Notary v2 uses it, and any example will help.

signature-envelope-cose.md
gokarnm: Yes that helps, maybe I missed it while reviewing last version. Resolving.

signature-envelope-cose.md
gokarnm: @yizha1 how about?

"Construct the Sig_structure for `COSE_Sign1`.  The `context` field is used to identify the context of signature,   string `Signature1` is used to identity a `COSE_Sign1` structure."

signature-envelope-cose.md
yizha1: The externally supplied data is explained in COSE spec, see https://www.rfc-editor.org/rfc/rfc9052.html#Extern_AAD, and Notary v2 doesn't use it. The value of `alg` is explained here https://www.rfc-editor.org/rfc/rfc9052.html#name-common-cose-header-paramete, which was copied below for your convenience:
"alg: This header parameter is used to indicate the algorithm used for the security processing. This header parameter MUST be authenticated where the ability to do so exists. This support is provided by AEAD algorithms or construction (e.g., COSE_Sign and COSE_Mac0). This authentication can be done either by placing the header parameter in the protected-header-parameters bucket or as part of the externally supplied data ([Section 4.3](https://www.rfc-editor.org/rfc/rfc9052.html#Extern_AAD)). The value is taken from the "COSE Algorithms" registry (see [[COSE.Algorithms](https://www.rfc-editor.org/rfc/rfc9052.html#COSE.Algorithms)])."

signature-envelope-cose.md
yizha1: This field is explained here https://www.rfc-editor.org/rfc/rfc9052.html#Extern_AAD, since Notary v2 doesn't use it, so it's empty.

specs/signature-specification.md
yizha1: This should not be changed. Check this https://datatracker.ietf.org/doc/html/rfc8949#section-3.4

signature-envelope-cose.md
```diff
@@ -0,0 +1,197 @@
...
+Example with Signing Scheme `notary.x509`
+
+```yaml
+{
+  / alg / 1: / PS384 / -38,
+  / crit / 2: [
+    'io.cncf.notary.signingScheme',
+    'io.cncf.notary.expiry'
+  ],
+  / cty / 3: 'application/vnd.cncf.notary.payload.v1+json',
```
OR13: this header struct looks good to me.

signature-envelope-cose.md
```diff
@@ -0,0 +1,197 @@
...
+
+- `io.cncf.notary.timestampSignature`
+- Label `33`: `x5chain`
+- `io.cncf.notary.signingAgent`
+
+```yaml
+{
+  / x5chain / 33: [
+    << DER(leafCert) >>,
+    << DER(intermediate CACert) >>,
```
OR13: I am still a bit nervous about exactly what this means, but the approach seems reasonable... I would just want to see a complete example.

signature-envelope-cose.md
```diff
@@ -0,0 +1,197 @@
...
+1. `alg` parameter values for various signature algorithms is a subset of values supported by [COSE](https://www.iana.org/assignments/cose/cose.xhtml#algorithms).
+
+**Mapping of Notary v2 approved algorithms to COSE `alg` header parameter values**
+
+  | Signature Algorithm             | `alg` Label       |
+  | ------------------------------- | ----------------- |
+  | RSASSA-PSS with SHA-256         | -37               |
+  | RSASSA-PSS with SHA-384         | -38               |
+  | RSASSA-PSS with SHA-512         | -39               |
+  | ECDSA on secp256r1 with SHA-256 | -7                |
```
OR13: I assume lack of EdDSA is intentional here.

specs/signature-specification.md
```diff
@@ -43,6 +43,7 @@ Signature Manifest Example
 - **`blobs`** (*array of objects*): This REQUIRED property contains collection of only one [artifact descriptor][artifact-descriptor] referencing signature envelope.
   - **`mediaType`** (*string*): This REQUIRED property contains media type of signature envelope blob. Following values are supported
     - `application/jose+json`
```
OR13: This seems like maybe not correct? are you saying that compact representation is NOT allowed? thats how i read them.

specs/signature-specification.md
```diff
@@ -165,6 +168,127 @@ These attributes are considered unsigned with respect to the signing key that ge
...
+Unless explicitly specified as OPTIONAL, all fields are required.
+Also, there shouldn’t be any additional fields other than ones specified in JWSPayload, ProtectedHeader, and UnprotectedHeader.
+
+**JWSPayload a.k.a. Claims**:
+Notary v2 is using one private claim (`notary`) and two public claims (`iat` and `exp`).
+An example of the claim is described below
+
+```json
+{
+   "subject": {
```
OR13: good call, since `sub` is a string... maybe you want to make it clear that this is a "private claim", that is not part of the original RFC.

yizha1:
@gokarnm  All your comments are resolved now. Could you give a final check and approve it if there are no further issues?

gokarnm (approved):
LGTM! @priteshbandi could you review and approve as well.

SteveLasker:
+1 to merging, with incremental improvements.

specs/signature-envelope-cose.md
```diff
@@ -0,0 +1,212 @@
...
+  'io.cncf.notary.signingAgent': 'notation/1.0.0'
+}
+```
+
+Note: `<<` and `>>` are used to notate the CBOR byte string resulting from encoding the data item.
+
+- **[`x5chain`](https://datatracker.ietf.org/doc/html/draft-ietf-cose-x509-08#section-2)** (*array of bstr*): This REQUIRED parameter (label `33` by [IANA](https://www.iana.org/assignments/cose/cose.xhtml#header-parameters)) contains the ordered list of X.509 certificate or certificate chain ([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the COSE. The certificate chain is represented as an array of certificate, each certificate in the array is DER encoded and then wrapped in a CBOR byte string. The certificate containing the public key corresponding to the key used to digitally sign the COSE MUST be the first certificate, followed by the intermediate and root certificates in the correct order. Refer [*Certificate Chain* unsigned attribute](signature-specification.md#unsigned-attributes) for more details. Optionally, this header can be presented in the protected header.
+  - **TODO** update signature specification to allow chains in protected header
```
priteshbandi: What's the motivation behind this? can we remove it from spec and open it as issue for discussion ?

priteshbandi (approved):
LGTM with one comment

specs/signature-envelope-cose.md
SteveLasker: @letmaik, @shizhMSFT can you comment?

shizhMSFT: The motivation is that the signing key of a signer may have multiple organizations associated. How to determine which organization that the signer represents at the signing time is an issue.

Here's an example, the signer has two cert chains `Company A -> Signer` and `Company B -> Signer`.
```mermaid
graph TD;

a["CA (Company A)"] --> s["Signer Cert"]
b["CA (Company B)"] --> s
```
Consider the following scenario:
1. Signer signs the artifact and associates cert chain `Company A -> Signer` to the resulted signature. In other words, the signer signs the artifact **representing the company A**.
2. Signer publishes the artifact and the signature.
3. Adversary sees the signature and replaces the cert chain in the unprotected header with a cert chain `Company B -> Signer`.
4. Now, from the view of verifiers, the signature is signed by the Signer, **representing the company B**.

This kind of attack can be mitigated by never reusing the signing key but it is better if we can sign the signing identity in the signature itself.

@yizha1 Let's remove it from spec and open it as issue for discussion.

SteveLasker: @ianjmcm, can you provide some 👀 ?

shizhMSFT (approved):
LGTM

SteveLasker:
Thanks @shizhMSFT, @qmuntal, @letmaik, @yogeshbdeshpande, @thomas-fossati and everyone else that completed this baseline, including all the updates to  https://github.com/veraison/go-cose

Great work!

PR #149, opened 2022-04-21
Simplify trust store and trust policy.

gokarnm:
* Changes to trust store
	* Moved to directory based trust store
* Changes to trust policy
	* Removed support to specify multiple trust stores
	* Renamed `trustAnchor` to `trustedIdentity`, can support other trusted identities in future
	* Simplified verification policy config
	* Rename `scope` to `registryScopes` to be more explicit
	* Simplified signature verification with values `strict`, `permissive`, `audit`, `skip` , no need to a separate `skipSignatureVerification`

cc: [roywill](https://github.com/roywill) for review.


trust-store-trust-policy-specification.md
```diff
@@ -1,154 +1,88 @@
...
+
+- The Notary v2 trust store consists of multiple named collections of certificates, called named stores.
+- The trust store is a directory location, under which each sub directory is considered a named store, that contains zero or more certificates. The name of this sub directory is used to reference the specfic store in trust policy.
+- Certificates in a trust store are typically root certificates. Intermediate certificates can be used, but not recommended as intermediate certificates can be rotated more often and can break signature verification.
+
+Notary v2 uses following directory structure to represent the trust store. The example shows named stores `acme-rockets` and `wabbit-networks`, which are subseqently references in the trust policy. Without this reference, presence of a named store and certificates in it does not confer trust automatically to the named store. The trust store is configured ahead of verification time, by an out of band mechanism that is beyond the scope of this document. Different entities and organizations have their own processes and policies to configure and distribute trust stores.
+
+```text
+~/.notation/trust_store
```
shizhMSFT: The notation CLI currently follows the [XDG](https://wiki.archlinux.org/title/XDG_Base_Directory) directories. Should the root directory be `$XDG_CONFIG_HOME/notation/trust_store`?

trust-store-trust-policy-specification.md
```diff
@@ -1,154 +1,88 @@
...
+    /x509
+        /ca_roots
+            /acme-rockets
+                cert1.pem
```
shizhMSFT, line 37: Is the extension restricted to `.pem`? Is `.crt` accepted?

trust-store-trust-policy-specification.md
shizhMSFT: It seems that it is more common in Linux to have hyphens than underscores. For instances, `trust-store` and `ca-roots`.

trust-store-trust-policy-specification.md
```diff
@@ -1,154 +1,88 @@
...
+                cert3.pem
+        /tsa_roots
+            /publicly-trusted-tsa
+                tsa-cert1.pem
 ```
 
-### Trust Store Properties
+The Trust store currently supports two kinds of identities, additional identities may be supported in future :
+
+- **Certificates**: The `x509/ca` trust store contains named stores that contain Certificate Authority (CA) root and intermediate certificates.
```
shizhMSFT: `x509/ca` is inconsistent with `/x509/ca_roots`.

shizhMSFT: Intermediate certificates should not be in the `roots` folders since they are not root CA certs.

trust-store-trust-policy-specification.md
```diff
@@ -1,154 +1,88 @@
...
-                "registry.wabbit-networks.io/software/net-logger" ],
-            "trustStores": [ "trust-store-name-1", "trust-store-name-2" ],
-            "trustAnchors": [
-                "subject: C=US, ST=WA, L=Seattle, O=acme-rockets.io"
+            "name": "policy-for-wabbit-networks-images",
+            "registryScopes": [
+                "registry.acme-rockets.io/software/net-monitor"
+                "registry.acme-rockets.io/software/net-logger" ],
+            "signatureVerification" : "strict",
+            "trustStore": "ca/wabbit-networks",
```
shizhMSFT: I'm confused here since `ca` does not match `ca_roots`. Should we make everything consistent?

trust-store-trust-policy-specification.md
```diff
@@ -1,154 +1,88 @@
...
-                "signingIdentityRevocation": "enforceWithFailOpen | enforceWithFailClose | warn | skip",
-                "timestampRevocation": "enforceWithFailOpen | enforceWithFailClose | warn | skip"
-            }
         },
         {
-            "name": "skip-signature-verification",
-            "scopes": [ "wabbit-networks.io/software/unsigned/productA" ],
-            "skipSignatureVerification": true,
+            "name": "policy-for-unsigned-image",
+            "registryScopes": [ "registry.wabbit-networks.io/software/unsigned/productA" ],
```
shizhMSFT: nit: repository name does not allow capital letters.

trust-store-trust-policy-specification.md
```diff
@@ -159,38 +93,34 @@ The trust policy is represented as JSON data structure as shown below:
...
+**Trusted timestamp** : Guarantees that the signature was generated when the certificate was valid. It also allows a verifier to determine if a signature be treated as valid when a certificate is revoked, if the certificate was revoked after the signature was generated. In the absence of a trusted timestamp, signatures are considered invalid after certificate expires, and all signatures are considered revoked when a certificate is revoked.
+**Expiry** : This is an optional feature that gurantees that an artifact is considered fresh, if the signing identity indicates an optional artifact expiry time.
+**Revocation check** : Guarantees that the signing identity is still trusted at signature verification time. Events such as key or system compromise can make a signing identity that was previously trusted, to be subsequrntly untrusted. This guarantee typically requires a verification time call to an external system, which may not be consistently reliable. The `permissive` verification level only audits for revocation check and does not enforce it. If a particular revocation mechnanism provides is reliable, use `strict` verification level instead.
```
shizhMSFT: We probably need empty lines between those lines for better rendering.

trust-store-trust-policy-specification.md
gokarnm: Should it be `$XDG_CONFIG_HOME/.notation/trust_store` instead?

trust-store-trust-policy-specification.md
gokarnm: Addressed, what about `.cer`?

trust-store-trust-policy-specification.md
gokarnm: Fixed in latest version.

trust-store-trust-policy-specification.md
gokarnm: Fixed the typo in latest version.

gokarnm:
* Final version for review
  *  Updated signature verification process based in changes to trust policy
  * Added additional policy examples
  * Clarified usage of optional `expiry` feature
  * Changed trusted identities configuration structure
  * Added a DRAFT notice to revocation section (CRL/OCSP), we should revisit this after RC1.
  
cc: @shizhMSFT @rgnote @roywill for review

trust-store-trust-policy-specification.md
```diff
@@ -1,154 +1,143 @@
 # Trust Store and Trust Policy Specification
 
-This document describes how Notary v2 signatures are evaluated for trust.
+Notary v2 currently supports X.509 based PKI and identities, and uses a trust store and trust policy to determine if a signed artifact is considered authentic.
+
 The document consists of the following sections:
 
-- **[Trust Store](#trust-store)**: Defines set of signing identity that user trusts.
-- **[Trust Policy](#trust-policy)**: Defines how the artifact is evaluated for trust.
+- **[Trust Store](#trust-store)**: Contains a set of trusted identities through which trust is derived for the rest of the system. For X509 PKI, the trust store contains a set of root certificates.
```
roywill: This is a question.  Is an intermediary Certificate allowed in this collection and thus not always in the true sense a root certificate.

roywill: Notice below you changed the verbiage to say typically.

trust-store-trust-policy-specification.md
```diff
@@ -1,154 +1,143 @@
...
 
 ## Trust Store
 
-Users who consume and execute the signed artifact from a registry need a mechanism to specify the trusted producers.
-This is where Trust Store is used.
+Contains a set of trusted identities through which trust is derived for the rest of the system. For X509 PKI, the trust store typically contains root certificates.
+
+- The Notary v2 trust store consists of multiple named collections of certificates, called named stores.
+- Following certificate formats are supported - Files with extension .pem, .crt and .cer, the files are expected to contain certificate(s) in DER (binary) format or PEM format (base-64 encoded DER).
+- The trust store is a directory location, under which each sub directory is considered a named store, that contains zero or more certificates. The name of this sub directory is used to reference the specfic store in trust policy.
```
roywill: <sp> specfic => specific

trust-store-trust-policy-specification.md
```diff
@@ -158,41 +147,44 @@ The trust policy is represented as JSON data structure as shown below:
...
+
+#### Signature Verification details
+
+Notary v2 defines the following signature verification levels, to provide different levels of enforcement for different scenarios. Signature verrification is a multi step process that checks for the following - integrity (artifact is unaltered, signature is not currupted), authenticity (the signature is really from the identity that claims to have signed it), trusted timestamping (the signature was generated when the key/certificate were unexpired), expiry (an optional check if the artifact specifies an expiry date), revocation check (is the signing identity still trusted at present time). Based on the signature verification level, each of these checks is `enforced` or `logged`. `Enforced` causes signature verification to fail, and `logged` causes failures to be logged and the next check is evaluated. NOTE: Implementations may change the ordering of these checks, but all checks MUST to be performed for signature verification process to be considered complete.
```
roywill: <sp> currupted => corrupted

roywill: Two comments.  Normally expiry check is cheap and done first, and the MUST in the last sentence is odd.  Normally at the first failure you stop, but do you want to continue the rest of the checks to perform a full audit?  That though means potentially megacycles of CPU.

trust-store-trust-policy-specification.md
```diff
@@ -158,41 +147,44 @@ The trust policy is represented as JSON data structure as shown below:
...
+|skip      |Use to exclude verification for unsigned images|skipped|skipped|skipped|skipped|skipped|
+
+**Integrity** : Guarantees that the artifact wasn't altered after it was signed, or the signature isn't corrupted. All signature verification levels always enforce integrity.
+
+**Authenticity** : Guarantees that the artifact was signed by an identity trusted by the verifier. Its definition does not include revocation, which is when a trusted identity is subsequently untrusted because of a compromise.
+
+**Trusted timestamp** : Guarantees that the signature was generated when the certificate was valid. It also allows a verifier to determine if a signature be treated as valid when a certificate is revoked, if the certificate was revoked after the signature was generated. In the absence of a trusted timestamp, signatures are considered invalid after certificate expires, and all signatures are considered revoked when a certificate is revoked. **NOTE**: `notation` RC1 will generate trusted timestamp using a TSA when signature is generated, but will not support verification of TSA countersignatures.
```
roywill: <grammar> TSA when signature => TSA when the signature

trust-store-trust-policy-specification.md
```diff
@@ -158,41 +147,44 @@ The trust policy is represented as JSON data structure as shown below:
...
+
+**Expiry** : This is an optional feature that guarantees that artifact is within “best by use” date indicated in the signature. Notary v2 allows users to include an optional expiry time when they generate a signature. The expiry time is not set by default and requires explicit configuration by users at the time of signature generation. The artifact is considered expired after the expiry time has elapsed, users performing verification can either configure their trust policies to fail the verification or even accept the signature with expiry date in the past using policy. This is an advanced feature that allows implementing controls for user defined semantics like deprecation for older artifacts, or block older artifacts in a production environment. Users should only include an expiry time in the signed artifact after considering the behavior they expect for consumers of the artifact after it expires. Users can choose to consume an artifact even after the expiry time based on their specific needs.
```
roywill: expiry time has elapsed is unclear.  Do you mean when the current time is greater than or equal to the expiry time or just greater?

trust-store-trust-policy-specification.md
```diff
@@ -212,7 +207,7 @@ The trust policy is represented as JSON data structure as shown below:
   1. *Global*: If there exists a trust policy with global scope then use that policy for signature evaluation.
      Otherwise, fail the signature verification.
 
-### Trust Anchors Constraints
+### Trusted Identities Constraints
 
 A distinguished name (usually just shortened to "DN") uniquely identifies an entry and in the case of the certificate's subject, DN uniquely identifies the requestor/holder of the certificate.
 The DN is comprised of zero or more comma-separated components called relative distinguished names, or RDNs.
```
roywill, line 215: in the line below do you mean to have '"' at the end of org1?

trust-store-trust-policy-specification.md
```diff
@@ -221,16 +216,16 @@ The RDN consists of an attribute type name followed by an equal sign and the str
...
 - If the subject DN of the signing certificate is used in the trust anchor, then it MUST meet the following requirements:
-  - The value of `trustAnchors` MUST begin with `subject:` followed by comma-separated one or more RDNs.
-    For example, `subject: C=${country}, ST=${state}, L=${locallity}, O={organization}, OU=${organization-unit}, CN=${common-name}`.
+  - The value of `trustedIdentities` MUST begin with `x509.subject:` followed by comma-separated one or more RDNs.
+    For example, `x509.subject: C=${country}, ST=${state}, L=${locallity}, O={organization}, OU=${organization-unit}, CN=${common-name}`.
   - Trust anchor MUST contain country (CN), state Or province (ST), and organization (O) RDNs.
     All other RDNs are optional.
-    The minimal possible trust anchor is `subject: C=${country}, ST=${state}, O={organization}`,
+    The minimal possible trust anchor is `x509.subject: C=${country}, ST=${state}, O={organization}`,
   - Trust anchor MUST NOT have overlapping values.
```
roywill: What is the expected error behavior when this happens.  Does the first one win or does functionality stop, or does the attempt to update this policy fail?

roywill (changes requested):

trust-store-trust-policy-specification.md
gokarnm: Yes, intermediate certs are allowed. Updated.
```suggestion
- **[Trust Store](#trust-store)**: Contains a set of trusted identities through which trust is derived for the rest of the system. For X.509 PKI, the trust store typically contains a set of root certificates.
```

trust-store-trust-policy-specification.md
gokarnm:
```suggestion
- The trust store is a directory location, under which each sub directory is considered a named store, that contains zero or more certificates. The name of this sub directory is used to reference the specific store in trust policy.
```

trust-store-trust-policy-specification.md
gokarnm: That's good feedback, the last sentence isn't clear enough. Reworded and addressed typos.

```suggestion
Notary v2 defines the following signature verification levels, to provide different levels of enforcement for different scenarios. Signature verification is a multi step process performs the following validation checks - integrity (artifact is unaltered, signature is not corrupted), authenticity (the signature is really from the identity that claims to have signed it), trusted timestamping (the signature was generated when the key/certificate were unexpired), expiry (an optional check if the artifact specifies an expiry time), revocation check (is the signing identity still trusted at the present time). Based on the signature verification level, each of these validation checks is `enforced` or `logged`. If a validation check is `enforced`, a failure is treated as critical failure, and causes the overall signature verification to fail.  If a validation check is `logged`, a failure causes details to be logged, and the next validation check is evaluated till all checks succeed or a critical failure is encountered. NOTE: Implementations may change the ordering of these checks based on efficiency, but all validation checks MUST to be performed till the first critical failure is encountered, or all checks succeed, for signature verification process to be considered complete.
```

trust-store-trust-policy-specification.md
gokarnm:
```suggestion
**Trusted timestamp** : Guarantees that the signature was generated when the certificate was valid. It also allows a verifier to determine if a signature be treated as valid when a certificate is revoked, if the certificate was revoked after the signature was generated. In the absence of a trusted timestamp, signatures are considered invalid after certificate expires, and all signatures are considered revoked when a certificate is revoked. **NOTE**: `notation` RC1 will generate trusted timestamp using a TSA when the signature is generated, but will not support verification of TSA countersignatures.
```

trust-store-trust-policy-specification.md
gokarnm: Updated wording to be precise - current time is greater than or equal to expiry time
```suggestion
**Expiry** : This is an optional feature that guarantees that artifact is within “best by use” date indicated in the signature. Notary v2 allows users to include an optional expiry time when they generate a signature. The expiry time is not set by default and requires explicit configuration by users at the time of signature generation. The artifact is considered expired when the current time is greater than or equal to expiry time, users performing verification can either configure their trust policies to fail the verification or even accept the artifact with expiry date in the past using policy. This is an advanced feature that allows implementing controls for user defined semantics like deprecation for older artifacts, or block older artifacts in a production environment. Users should only include an expiry time in the signed artifact after considering the behavior they expect for consumers of the artifact after it expires. Users can choose to consume an artifact even after the expiry time based on their specific needs.
```

trust-store-trust-policy-specification.md
gokarnm: Good point, I'll clarify that signature verification fails with "invalid policy".

trust-store-trust-policy-specification.md
gokarnm: Fixed in latest revision.

trust-store-trust-policy-specification.md
```diff
@@ -1,154 +1,143 @@
 # Trust Store and Trust Policy Specification
 
-This document describes how Notary v2 signatures are evaluated for trust.
+Notary v2 currently supports X.509 based PKI and identities, and uses a trust store and trust policy to determine if a signed artifact is considered authentic.
+
 The document consists of the following sections:
 
-- **[Trust Store](#trust-store)**: Defines set of signing identity that user trusts.
-- **[Trust Policy](#trust-policy)**: Defines how the artifact is evaluated for trust.
+- **[Trust Store](#trust-store)**: Contains a set of trusted identities through which trust is derived for the rest of the system. For X.509 PKI, the trust store typically contains a set of root certificates.
```
SteveLasker, line 7: nit, "typically" are there times when it wouldn't be root certs?

trust-store-trust-policy-specification.md
```diff
@@ -1,154 +1,143 @@
...
+- **[Trust Policy](#trust-policy)**: A policy language which indicates which identities are trusted to produce artifacts. Both trust store and trust policy need to be configured by users/administrators before artifact signature can be evaluated.
+- **[Signature Verification](#signature-verification)**: Describes how signatures are evaluated using the policy, to determine if a signed artifact is authentic. This section is meant for implementors of Notary v2 standards.
+
+Other types of identities and trust models may be supported in future, which may introduce other constructs/policy elements to support signature evaluation.
+
+All examples use the actors defined in Notary v2 [scenario](https://github.com/notaryproject/notaryproject/blob/main/scenarios.md#scenario-0-build-publish-consume-enforce-policy-deploy)
+
```
SteveLasker, line 16: A suggestion to put a header that says: ## Scenario

trust-store-trust-policy-specification.md
```diff
@@ -1,154 +1,143 @@
...
+- Wabbit Networks company builds, signs and distributes their `net-monitor` software though public registries.
+- ACME Rockets consumes the `net-monitor` software from a public registry importing the  artifacts and reference artifacts (signatures, SBoMs) into their private registry. The private registry also contains additional artifacts that ACME Rockets themselves sign.
```
SteveLasker: nit the <double space> artifacts

trust-store-trust-policy-specification.md
```diff
@@ -1,154 +1,143 @@
...
 ## Trust Store
 
-Users who consume and execute the signed artifact from a registry need a mechanism to specify the trusted producers.
-This is where Trust Store is used.
+Contains a set of trusted identities through which trust is derived for the rest of the system. For X.509 PKI, the trust store typically contains root certificates.
+
+- The Notary v2 trust store consists of multiple named collections of certificates, called named stores.
+- Following certificate formats are supported - Files with extension .pem, .crt and .cer, the files are expected to contain certificate(s) in DER (binary) format or PEM format (base-64 encoded DER).
+- The trust store is a directory location, under which each sub directory is considered a named store, that contains zero or more certificates. The name of this sub directory is used to reference the specific store in trust policy.
+- Symlinks are not supported for the named store directories or certificate files. Implementation MUST validate that the named store directory or certificate files are not symlinks, and fail it it does not meet this condition.
```
SteveLasker: ...fail it it 
should be:
...fail **if** it

trust-store-trust-policy-specification.md
```diff
@@ -1,154 +1,143 @@
...
         {
-            "name": "global-trust-policy",
-            "scopes": [ "*" ],
-            "trustStores": [ "trust-store-name-1", "trust-store-name-2" ],
-            "expiryValidations": {
-                "signatureExpiry": "enforce | warn",
-                "signingIdentityExpiry": "enforce | warn",
-                "timestampExpiry": "enforce | warn"
+            # Policy for with custom verification policy
+            "name": "use-expired-image",
```
SteveLasker, line 115: Should we require * to be added to support all?
Or, should we assume a default scenario where a an entire registry is trusted, unless scoped?

I realize that someone may want to limit docker hub content to library, however most secured environments use private registries, where the registry is the trust boundary, not necessarily a sub-namespace.

gokarnm: The `registryScopes` with value `*` makes it explicit to user that any registry is trusted, it's better to be explicit in security related policies, unlike other policies where we may favor conciseness.  We used the same for `identities` where it's now a required property but you can specify a wildcard `*`.

```
"name": "myPolicy",
            "registryScopes": [ "*" ],
            "signatureVerification" : "strict",
            "trustedIdentities": {
                "trustStore": "ca:acme-rockets",
                "identities": [ "*"]
            }
```

gokarnm:
Adding @rgnote as a reviewer, as he has been implementing this spec.

rgnote:
LGTM

SteveLasker (approved):
LGTM

NiazFK:
LGTM

roywill (approved):

SteveLasker:
Great work @gokarnm

PR #151, opened 2022-04-26
Plugin extensibility JSON schema cleanup

qmuntal:
This PR contains several changes for the plugin extensibility JSON schema:

- Use camelCase instead of kebab-case. Existing notary schemas already use camelCase, so better stick to it instead of diverging.
- Remove `plugin` prefix from the plugin metadata schema. It seems clear that all props refer to the plugin itself, so the prefix doesn't add much value.
- Quote json placeholder values and add missing commas.

@SteveLasker @gokarnm


SteveLasker (approved):
LGTM

gokarnm (approved):

gokarnm:
LGTM!

PR #153, opened 2022-04-29
Update signature-envelope-cose.md

briankr-ms:
The `crit` header should be optional, not required. The spec already defines `cty` and `signingtime` as required, so a compliant verifier should reject any signature that does not include them, irrespective of whether or not the envelope also declares them as critical.

An appropriate use of `crit` would be if a signer intends to include a new header not defined in the spec, to instruct verifiers that they must understand and process that header in order for the signature to be successfully validated. Another scenario is if a signer intends to mandate processing of an otherwise optional header (e.g., `expiry`).

In other words, `crit` can be used by a signer to add requirements to the verification flow, but it is unnecessary and redundant if the verifier already has the same requirements.


SteveLasker (approved):
LGTM

SteveLasker (approved):
LGTM

PR #155, opened 2022-05-04
Added describe-key command and pass-through plugin config

gokarnm:
* Changed plugin path from `~/.notation/plugins/` to `~/notation/plugins/` to be consistent with other paths that use $XDG_CONFIG_HOME
* Updated `Signature Generator` workflow to include `describe-key` command, which is required to find the signing algorithm (to set `alg` header) before payload to sign is generated.
* Added spec for pass-through plugin config using CLI and `config.json`
* Minor formatting updates

cc: @qmuntal , @rgnote, @shizhMSFT  for review

### DIFF
--- a/specs/plugin-extensibility.md
+++ b/specs/plugin-extensibility.md
@@ -73,7 +73,8 @@ List all valid plugins.
           "plugin": "com.example.nv2plugin",
           // Optional configuration as required by the plugin
           "pluginConfig" : {   
-            "key1" : "value1" 
+            "key1" : "value1",
+            "key 2" : "value 2"
           }
         }
     ]  
@@ -88,10 +89,13 @@ Plugins may require additional configuration to work correctly, that could be se
 * To use this feature, plugin authors MUST define and document the set of plugin configuration keys-values, which is set by users when they associate a plugin with signing key.
 * Plugin authors SHOULD NOT use plugin configuration to store sensitive configuration in plaintext, such as authentication keys used by the plugin etc.
 
-In the previous example, plugin config can be set using command line arguments as part of registring a key, or while signing using the key. Plugin config values provided by `notation sign` override values already set in `config.json` E.g.
+For the previous example, plugin config can be set using command line arguments as part of registering a key with the `notation key add` command. E.g.
 
-* `notation key add --name "mysigningkey" --id "keyid" --plugin "com.example.nv2plugin" --pluginConfig.key1 "value1"`
-* `notation sign $IMAGE --key "mysigningkey" --pluginConfig.key1 "value2"`
+* `notation key add --name "mysigningkey" --id "keyid" --plugin "com.example.nv2plugin" --pluginConfig key1=value1,"key 2"="value 2"`
+
+Plugin config can be also set/overriden during signing with the `notation sign` command. Following example overrides value for `key 2` already set in `config.json` by previous command.
+
+* `notation sign $IMAGE --key "mysigningkey" --pluginConfig "key 2"=newValue2`
 
 ### Plugin contract
 
--- a/specs/plugin-extensibility.md
+++ b/specs/plugin-extensibility.md
@@ -185,10 +185,10 @@ This interface targets plugins that integrate with providers of basic cryptograp
         2. Generate the payload to be signed for [JWS](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#supported-signature-envelopes) envelope format.
            1. Create the JWS protected headers collection and set `alg` to value corresponding to `describe-key.response.keySpec` as per [signature algorithm selection](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#algorithm-selection).
            2. Create the `JWSPayload` with appropriate private (`subject`) and public (`iat,exp`) claims.
-           3. The payload to sign is then created as - `ASCII(BASE64URL(UTF8(ProtectedHeaders)) ‘.’ BASE64URL(JWSPayload))`
+           3. The *payload to sign* is then created as - `ASCII(BASE64URL(UTF8(ProtectedHeaders)) ‘.’ BASE64URL(JWSPayload))`
         3. Execute the plugin with `generate-signature` command.
            1. Set `request.keyId` and the optional `request.pluginConfig` to corresponding values associated with signing key `keyName` in `config.json`.
-           2. Set `request.payload` as the payload to sign.
+           2. Set `request.payload` as base64 encoded *payload to sign* (the JWS *payload to sign* is double encoded, this is a shortcoming of using plugin contract with JSON encoding).
            3. Set `keySpec` to value returned by `describe-key` command in `response.keySpec`, and `hashAlgorithm` to hash algorithm corresponding to the key spec, as per [signature algorithm selection](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#algorithm-selection). The algorithm specified in `hashAlgorithm` MUST be used by the plugin to hash the payload (`request.payload`) as part of signature generation.
         4. Validate the generated signature, return an error if any of the checks fails.
            1. Check if `response.signingAlgorithm` is one of [supported signing algorithms](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#algorithm-selection).
@@ -265,11 +265,8 @@ This command is used to generate the raw signature for a given payload.
   // hash the payload using this hash algorithm
   "hashAlgorithm" : "SHA_256" | "SHA_384" | "SHA_512",
 
-  // The signature envelope type associated with the payload.
-  "signatureEnvelopeType" : "application/vnd.cncf.notary.v2.jws.v1",
-  
-  // Payload to sign, it's encoded based on the signature envelope type
-  "payload" : "<payload to be signed>",
+  // Payload to sign, this is base64 encoded
+  "payload" : "<base64 encoded payload to be signed>",
 
 }
 ```
@@ -282,9 +279,7 @@ This command is used to generate the raw signature for a given payload.
 
 *hashAlgorithm* : Required field that specifies Hash algorithm corresponding to the signature algorithm determined by `keySpec` for the key.
 
-*signatureEnvelopeType* - Required field that specifies signature envelope type associated with the payload. This is used by the plugin to appropriately decode the payload in case a binary payload is passed.
-
-*payload* : Required field that contains payload to be signed. For JWS envelope type, the payload to sign is sent as is, without any additional encoding.
+*payload* : Required field that contains base64 encoded payload to be signed. For JWS, the *payload to sign* is base64 encoded (a second time) before sending to the plugin.
 
 *Response*
 
--- a/specs/plugin-extensibility.md
+++ b/specs/plugin-extensibility.md
@@ -177,15 +177,15 @@ This interface targets plugins that integrate with providers of basic cryptograp
 4. Determine if the registered key uses a plugin
 5. Execute the plugin with `get-plugin-metadata` command
     1. If plugin supports capability `SIGNATURE_GENERATOR`
-        1. Execute the plugin with `describe-key` command, set `request.keyName`, `request.keyId` and the optional `request.pluginConfig` to corresponding values associated with signing key `keyName` in `config.json`.
+        1. Execute the plugin with `describe-key` command, set `request.keyId` and the optional `request.pluginConfig` to corresponding values associated with signing key `keyName` in `config.json`.
         2. Generate the payload to be signed for [JWS](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#supported-signature-envelopes) envelope format.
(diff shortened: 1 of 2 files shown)


gokarnm:
Proposed changes to directory structure in [#167](https://github.com/notaryproject/notation/issues/167#issuecomment-1116978641)

gokarnm:
@qmuntal , the updated spec contains details for pass-through plugin config whose implementation can be deferred.

qmuntal (approved):

qmuntal:
@gokarnm the `generate-signature` request payload is supposed to be digested by the plugin using the hashing algorithm associated to the signing algorithm, but I see nowhere in the spec where is this mapping defined.

Would it make sense to add it, or even just pass the payload digested so the plugin doesn't have to do so?

gokarnm:
> @gokarnm the `generate-signature` request payload is supposed to be digested by the plugin using the hashing algorithm associated to the signing algorithm, but I see nowhere in the spec where is this mapping defined.

The mapping is available in this section of [signature spec](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#algorithm-selection). I think this points to a gap in the interface, if the plugin is expected to generate the hash, it would need to store this mapping, and potentially call `describe-key` again internally to get the key spec. This can be addressed by passing `keySpec`(informational) and `hashAlgorithm` in `generate-signature` request.  

> Would it make sense to add it, or even just pass the payload digested so the plugin doesn't have to do so?

In [draft spec](https://hackmd.io/jhYbA4aNRFW7nmp7gPTIBA?view#Signature-Generator), hash was calculated by notation and passed in the request. I changed it to current form to make it easier to add FIPS support.
* Hash algorithms are in the scope of FIPS 140-2
* Go lang crypto is not FIPS validated
* Some KMS and remote providers are FIPS validated, and provide a sign operation that accepts payload in the `sign` operation, which is hashed and signed by it.
* As plugin is an executable, plugin authors have the option to implement plugin in a different language that has better FIPS support (say using BouncyCastle for Java and .NET)

So we moved the burden of hashing to the plugin. There is an option to support plugins that would not prefer hashing, and do not expect the provided hash to be from a FIPS validated module. We can pass both the `payload` and `payloadHash` in the `generate-signature` request (an additional step after generating payload to sign). The alternate requires discovering if the plugin prefers hashing which makes `get-plugin-metadata` more complicated, something I'd avoid.

qmuntal:
> The mapping is available in this section of [signature spec](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#algorithm-selection). I think this points to a gap in the interface, if the plugin is expected to generate the hash, it would need to store this mapping, and potentially call describe-key again internally to get the key spec. This can be addressed by passing keySpec(informational) and hashAlgorithm in generate-signature request.

Thanks for pointing out the mapping. I don't have a strong option whether we should add `keySpec` and `hashAlgorithm` to `generate-signature`. On one side, it complicates the schema and opens the door to inconsistencies. One the other side, plugin authors may avoid fetching the key spec a second time and don't have to code the `keySpec` => `hashAlgorithm` mapping.

> In [draft spec](https://hackmd.io/jhYbA4aNRFW7nmp7gPTIBA?view#Signature-Generator), hash was calculated by notation and passed in the request. I changed it to current form to make it easier to add FIPS support.

That's a very good reason to delegate hashing to plugins. The burden is not that big, so it pay-off.

gokarnm:
@qmuntal , I've updated the `generate-signature` request to indicate hash algorithm to use.

> Thanks for pointing out the mapping. I don't have a strong option whether we should add `keySpec` and `hashAlgorithm` to `generate-signature`. On one side, it complicates the schema and opens the door to inconsistencies. One the other side, plugin authors may avoid fetching the key spec a second time and don't have to code the `keySpec` => `hashAlgorithm` mapping.

I think the potential for inconsistencies is addressed as notation will perform basic validation on the signature generated by the plugin, that plugin really does use the same signing and hash alg notation expects.

specs/plugin-extensibility.md
```diff
@@ -155,53 +171,122 @@ This interface targets plugins that integrate with providers of basic cryptograp
...
+
+  // The key spec for the given key id
+  "keySpec" : "<key type and size>",
+
+  // Hash algorithm associated with the key spec, plugin must 
+  // hash the payload using this hash algorithm
+  "hashAlgorithm" : "SHA_256" | "SHA_384" | "SHA_512",
+
+  // The signature envelope type associated with the payload.
+  "signatureEnvelopeType" : "application/vnd.cncf.notary.v2.jws.v1",
```
qmuntal: This used to be only necessary for the generate envelope request. Why would a plugin which is not envelope-aware want to know the envelope type?

qmuntal:
Thinking out loud: is it really necessary to pass `keyName` in each request?

`keyName` is intended to be a friendly name only relevant to `notation`. The external key identifier is `keyId`, and the plugin should be good only having it. I'm afraid plugins will start using `keyName` as another way to identify the keys, which will end up hurting users, because `keyName` will then have to be unique among `notation` local configuration and the plugin's kms.

qmuntal:
Thinking out loud: is it really necessary to pass `keyName` in each request?

`keyName` is intended to be a friendly name only relevant to `notation`. The external key identifier is `keyId`, and the plugin should be good only having it. I'm afraid plugins will start using `keyName` as another way to identify the keys, which will end up hurting users, because `keyName` will then have to be unique among `notation` local configuration and the plugin's kms.

gokarnm:
> Thinking out loud: is it really necessary to pass `keyName` in each request?
> 
> `keyName` is intended to be a friendly name only relevant to `notation`. The external key identifier is `keyId`, and the plugin should be good only having it. I'm afraid plugins will start using `keyName` as another way to identify the keys, which will end up hurting users, because `keyName` will then have to be unique among `notation` local configuration and the plugin's kms.

That's a good observation, `keyName` should not have any relevance to plugin, and can be omitted from request. The only reason would be to include it in error messages and such for correlation, but the caller (notation) can do the same if required. I'll remove it in a separate PR.

gokarnm:
Pushed an update to remove `keyName` from plugin requests. Was a small change, so included it here so that the second reviewer also covers it.

specs/plugin-extensibility.md
```diff
@@ -56,33 +56,49 @@ To be considered a valid plugin a candidate must pass each of these "plugin cand
...
+
+Plugins may require additional configuration to work correctly, that could be set out of band, or provided by notation when it invokes a plugin. To support plugin configuration through notation, the key configuration provides an optional `pluginConfig` map of of key value pairs, that is passed as-is by notation to the plugin.
+
+* To use this feature, plugin authors MUST define and document the set of plugin configuration keys-values, which is set by users when they associate a plugin with signing key.
+* Plugin authors SHOULD NOT use plugin configuration to store sensitive configuration in plaintext, such as authentication keys used by the plugin etc.
+
+In the previous example, plugin config can be set using command line arguments as part of registring a key, or while signing using the key. Plugin config values provided by `notation sign` override values already set in `config.json` E.g.
+
+* `notation key add --name "mysigningkey" --id "keyid" --plugin "com.example.nv2plugin" --pluginConfig.key1 "value1"`
+* `notation sign $IMAGE --key "mysigningkey" --pluginConfig.key1 "value2"`
```
rgnote: Suggest we follow this pattern ```--pluginConfig Key1=Value1,Key2=Value2```

specs/plugin-extensibility.md
```diff
@@ -155,53 +171,119 @@ This interface targets plugins that integrate with providers of basic cryptograp
...
-        2. Execute the plugin with `generate-signature` command, set `request.keyDefinition` to key definition corresponding to `keyName`, `request.payload` to the generated payload.
-        3. Validate the generated signature, return an error if any of the checks fails.
+        1. Execute the plugin with `describe-key` command, set `request.keyId` and the optional `request.pluginConfig` to corresponding values associated with signing key `keyName` in `config.json`.
+        2. Generate the payload to be signed for [JWS](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#supported-signature-envelopes) envelope format.
+           1. Create the JWS protected headers collection and set `alg` to value corresponding to `describe-key.response.keySpec` as per [signature algorithm selection](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#algorithm-selection).
+           2. Create the `JWSPayload` with appropriate private (`subject`) and public (`iat,exp`) claims.
```
rgnote, line 187: How does notation set iat, exp values? Wouldn't these be dependent on the plugin itself?

specs/plugin-extensibility.md
gokarnm: Sounds good, will update in next rev

specs/plugin-extensibility.md
gokarnm: `iat` (signature creation time) and `exp` (optional expiry time based on user input) will be set by notation, these are not influenced by the key itself when using the `signature-generator` interface.

rgnote:
LGTM

rgnote (approved):

gokarnm:
@qmuntal this PR is ready for review, if you want to review the delta from your last approval.

specs/plugin-extensibility.md
gokarnm: Generally this would not be required. Currently the encoding of the `payload` field depends on envelope type being targeted. JWS payload is already base64 encoded and signature is generated over a base64 encoded payload, whereas a binary payload would have to be encoded when sending request and decoded by plugin before it generates the raw signature. This is a shortcoming of using JSON as the request encoding, when we switch to something binary friendly like COSE, `signatureEnvelopeType` will not be required in this request.

qmuntal (approved):
LGTM!

specs/plugin-extensibility.md
```diff
@@ -32,13 +32,13 @@ Plugin publisher will provide instructions to download and install the plugin. P
 
 To enumerate all available plugins the following paths are scanned:
 * Unix-like OSes:
-  * `$HOME/.notation/plugins`
+  * `$HOME/notation/plugins`
```
shizhMSFT, line 35: We still need the `.` for Unix-like systems.

shizhMSFT: There will be another doc to formalize all the paths.

specs/plugin-extensibility.md
```diff
@@ -155,53 +175,119 @@ This interface targets plugins that integrate with providers of basic cryptograp
...
+3. Append any user provided metadata and Notary metadata as descriptor annotations.
+4. Determine if the registered key uses a plugin
+5. Execute the plugin with `get-plugin-metadata` command
     1. If plugin supports capability `SIGNATURE_GENERATOR`
-        1. Generate the payload to be signed. For JWS this includes [JWS](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#supported-signature-envelopes) payload with `subject` claim as descriptor, additional JWS claims, and protected headers.
-        2. Execute the plugin with `generate-signature` command, set `request.keyDefinition` to key definition corresponding to `keyName`, `request.payload` to the generated payload.
-        3. Validate the generated signature, return an error if any of the checks fails.
+        1. Execute the plugin with `describe-key` command, set `request.keyId` and the optional `request.pluginConfig` to corresponding values associated with signing key `keyName` in `config.json`.
+        2. Generate the payload to be signed for [JWS](https://github.com/notaryproject/notaryproject/blob/main/signature-specification.md#supported-signature-envelopes) envelope format.
```
shizhMSFT, line 185: How about other envelope format other than JWS?

specs/plugin-extensibility.md
qmuntal: Wouldn't it be simpler if `payload` is always base64-encoded? Making the encoding depend on the envelope type, means that plugins implementing the `generate-signature` workflows will have to be updated if we add a new envelope type, instead of being generic enough to blindly sign whatever payload.

The only advantage I see on having the encoding depending on the envelope format is to avoid double-encoding payloads, for example in JWS. On the other hand, base64 encoding is normally highly optimized, my computer can encode bytes at 500MB/s, so no-one will notice that we are double-encoding some payloads.

specs/plugin-extensibility.md
gokarnm: Plan to update this section after we update the signature spec to support COSE, and separate out JWS and COSE spec as their own docs. This should probably just reference a section in the JWS/COSE spec instead of calling out the details here. Lots of doc refactoring to do, after the specs settle down :)

specs/plugin-extensibility.md
gokarnm: Agree that including `signatureEnvelopeType` makes the plugin implementing `generate-signature` interface to be updated to support newer payload types, which is not one of the goals of this interface. Consistent payload encoding will in be inline with request contract we would have when we switch to a binary friendly encoding in future.

gokarnm: @qmuntal , latest rev includes this change.

qmuntal (approved):
LGTM!

gokarnm:
@rgnote added clarification for multi-valued RDNs.

NiazFK (approved):
LGTM

PR #156, opened 2022-05-04
Address review comments on #148 for JWS

letmaik:
This addresses the comments I made in #148 on the JWS spec.

@shizhMSFT Please review.

PR #158, opened 2022-05-16
Signature spec updates

gokarnm:
- Updates Notary v2 signature envelope and JWS envelope in `main`, based subset of updates in `cose-envelope` branch.
- JWS envelope does not rely on JWT defined claims.
- Payload is standardized so we can use the same payload across different signature envelope formats.
- Incorporated some feedback from #156 by @letmaik, JWT related updates are discarded.
- I'll address the merge conflict at the end after other review feedback is addressed.

cc: @shizhMSFT @roywill @letmaik @priteshbandi @rgnote for review


signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+  }
+}
+```
+
+## Protected Headers
+
+The JWS envelope for Notary v2 uses following headers
+
+- Registered headers `alg`, `cty`, and `crit`
+- Public headers `io.cncf.notary.signingTime`, `io.cncf.notary.expiry`, `io.cncf.notary.clientID`
```
priteshbandi: Are registering(or plans to register) them in IANA? https://datatracker.ietf.org/doc/html/rfc7515#section-4.2

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+}
+```
+
+- **[`alg`](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.1)**(*string*): This REQUIRED header defines which algorithm was used to generate the signature.JWS specification defines `alg` as a required header, that MUST be present and MUST be understood and processed by verifier. This [section](#supported-alg-header-values) lists the Notary v2 allowed subset of `alg` values supported by JWS.
+- **[`cty`](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.10)**(*string*): The REQUIRED header content-type is used to declare the media type of the secured content (the payload). The supported value is `application/vnd.cncf.notary.payload.v1+json`.
+- **`io.cncf.notary.signingTime`**(*string*): This REQUIRED header specifies the time at which the signature was generated. Its value is a RFC 3339 formatted date time.
+- **`io.cncf.notary.expiry`**(*string*): This OPTIONAL header provides a “best by use” time for the artifact, as defined by the signer. Its value is a RFC 3339 formatted date time.
+- **`io.cncf.notary.clientID`**(*string*): This OPTIONAL header provides The version of a client (e.g. Notation) that produced the signature. E.g. “notation/1.0.0”.
```
priteshbandi: nit:
s/The version/the identifier of client

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Implementation Constraints
+
+### Supported `alg` header values
+
+Notary v2 implementation MUST enforce the following constraints on signature generation and verification:
+
+1. `alg` header value MUST NOT be `none` or any symmetric-key algorithm such as `HMAC`.
+2. `alg` header value MUST be same as that of signature algorithm identified using signing certificate's public key algorithm and size.
+3. `alg` header values for various signature algorithms is a subset of values supported by [JWS][jws-alg-values]:
```
priteshbandi: nit: please use 1. for all numbered points

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+
+Since Notary v2 restricts one signature per signature envelope, the compliant signature envelope MUST be in flattened JWS JSON format.
+
+```jsonc
+{
+    "payload": "<Base64Url(JWSPayload)>",
+    "protected": "<Base64Url(ProtectedHeaders)>",
+    "header": {
+        "timestamp": "<Base64(TimeStampToken)>",
+        "x5c": ["<Base64(DER(leafCert))>", "<Base64(DER(intermediateCACert))>", "<Base64(DER(rootCert))>"]
```
jondonas, line 127: Is cert ordering in the chain suggested or required?

signature-specification.md
```diff
@@ -1,243 +1,118 @@
...
 
 The Signature Envelope is a standard data structure for creating a signed message.
 A signature envelope consists of the following components:
 
-- Payload `m`: The data that is integrity protected - e.g. descriptor of the artifact being signed.
-- Signed attributes `v`: The signature metadata that is integrity protected - e.g. signature expiration time, signing time, etc.
-- Unsigned attributes `u`: This OPTIONAL property represents signature metadata that is not integrity protected - e.g. timestamp, certificates, etc.
+- Payload/Message `m`: The data that is integrity protected - e.g. descriptor of the artifact being signed.
+- Signed attributes `v`: The signature metadata that is integrity protected - e.g. signature expiration time, creation time, etc.
+- Unsigned attributes `u`: Unsigned attributes `u`: These attributes are not signed by the signing key that generates the signature. We anticipate unsigned attributes contain content that may be signed by an different party e.g. Certificate chain signed by a CA, or TSA countersignature signed by the TSA.
```
priteshbandi: nit:  Unsigned attributes `u` is redundant.

signature-specification.md
```diff
@@ -1,243 +1,118 @@
...
-
-- **Creation time**: The time at which the signature was generated.
-  Its value should be numeric representing the number of seconds (not milliseconds) since Epoch.
-- **Expiration time**: The time after which signature shouldn't be considered valid.
-  Its value should be numeric representing the number of seconds since Epoch.
-  This is an OPTIONAL attribute.
+Signed attributes/claims are additional metadata apart from the payload, which are required to support the signature verification process.
 
-### Unsigned Attributes
+- Any claims MUST NOT be stored/appended in the payload itself, as the payload is only parsed and processed once the signature has been verified (signature is valid, and from a trusted key) and trust is established.
```
priteshbandi: Instead of telling what not to do, can we start with what should go in signed attributes?

signature-specification.md
```diff
@@ -1,243 +1,118 @@
...
-  | RSASSA-PSS with SHA-256         | PS256             |
-  | RSASSA-PSS with SHA-384         | PS384             |
-  | RSASSA-PSS with SHA-512         | PS512             |
-  | ECDSA on secp256r1 with SHA-256 | ES256             |
-  | ECDSA on secp384r1 with SHA-384 | ES384             |
-  | ECDSA on secp521r1 with SHA-512 | ES512             |
-1. Signing certificate MUST be a valid codesigning certificate.
-1. Only JWS JSON flattened format is supported.
-   See 'Signature Envelope' section.
+- **Certificate Chain**: This property contains the list of X.509 certificate or certificate chain. This is a REQUIRED attribute. The certificate chain is authenticated using against a trust store as part of signature validation.
```
priteshbandi: Should certificate chain be in order leaf->root?

signature-specification.md
```diff
@@ -1,243 +1,118 @@
...
-- **Timestamp**:  The time stamp token generated for a given signature.
-  Only [RFC3161](https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2) compliant *TimeStampToken* are supported.
-  This is an OPTIONAL attribute.
+Notary v2 requires the signature envelope to support the following signed attributes/claims.
 
-### Supported Signature Envelopes
+#### Standard attributes
 
-#### JWS JSON Serialization
+- **Signing time**: The time at which the signature was generated. Though this claim is signed by the signing key, it’s considered unauthenticated as a signer can modify local time and manipulate this claim. More details [here](#signing-time).
```
priteshbandi: REQUIRED ?

signature-specification.md
```diff
@@ -1,243 +1,118 @@
...
-  | RSASSA-PSS with SHA-384         | PS384             |
-  | RSASSA-PSS with SHA-512         | PS512             |
-  | ECDSA on secp256r1 with SHA-256 | ES256             |
-  | ECDSA on secp384r1 with SHA-384 | ES384             |
-  | ECDSA on secp521r1 with SHA-512 | ES512             |
-1. Signing certificate MUST be a valid codesigning certificate.
-1. Only JWS JSON flattened format is supported.
-   See 'Signature Envelope' section.
+- **Certificate Chain**: This property contains the list of X.509 certificate or certificate chain. This is a REQUIRED attribute. The certificate chain is authenticated using against a trust store as part of signature validation.
+- **TSA counter signature** : The time stamp token generated for a given signature. Only [RFC3161](ietf-rfc3161) compliant TimeStampToken are supported. This is an OPTIONAL attribute.
```
gokarnm: Mention that TSA is a trusted time stamp.

signature-specification.md
```diff
@@ -1,243 +1,118 @@
...
-  This is an OPTIONAL attribute.
+Notary v2 requires the signature envelope to support the following signed attributes/claims.
 
-### Supported Signature Envelopes
+#### Standard attributes
 
-#### JWS JSON Serialization
+- **Signing time**: The time at which the signature was generated. Though this claim is signed by the signing key, it’s considered unauthenticated as a signer can modify local time and manipulate this claim. More details [here](#signing-time).
+- **Expiry** (critical): An OPTIONAL claim that provides a “best by use” time for the artifact, as defined by the signer. More details [here](#expiry).
+- **Content type** (critical): A REQUIRED clain that indicates the content type of the payload. Notary currently a payload that contains the OCI descriptor of a subject manifest as the payload, the supported value is `application/vnd.cncf.notary.payload.v1+json`, other payload types MAY be supported in future.
```
gokarnm: typo claim

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+        "io.cncf.notary.x509.fingerprint.sha256": "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\" \"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
+    }
+}
+```
+
+## JWS Payload
+
+The payload in the JWS envelope is set to BASE64URL encoded value of [Notary v2 Payload](./signature-specification.md#payload).
+
+Example of payload before BASE64URL encoding.
```
roywill: This example does not flow from the previous text.  You were talking about BASE64URL and then give an example that is before encoding but it then implies an after example.

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+    "io.cncf.notary.signingTime": "2022-04-06 07:01:20.52Z",
+    "io.cncf.notary.expiry": "2022-10-06 07:01:20.52Z",
+    "io.cncf.notary.clientID": "notation/1.0.0",
+    "crit":["io.cncf.notary.expiry"]
+}
+```
+
+- **[`alg`](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.1)**(*string*): This REQUIRED header defines which algorithm was used to generate the signature.JWS specification defines `alg` as a required header, that MUST be present and MUST be understood and processed by verifier. This [section](#supported-alg-header-values) lists the Notary v2 allowed subset of `alg` values supported by JWS.
+- **[`cty`](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.10)**(*string*): The REQUIRED header content-type is used to declare the media type of the secured content (the payload). The supported value is `application/vnd.cncf.notary.payload.v1+json`.
+- **`io.cncf.notary.signingTime`**(*string*): This REQUIRED header specifies the time at which the signature was generated. Its value is a RFC 3339 formatted date time.
```
roywill: Since time-frac is optional, do we recommend its specification?

signature-envelope-jws.md
gokarnm: It's in order from leaf to root, will clarify in description.

signature-envelope-jws.md
gokarnm: No, https://datatracker.ietf.org/doc/html/rfc7515#section-4.2 defines Public Headers as follow

```
either be registered in the IANA "JSON Web Signature and
   Encryption Header Parameters" registry established by [Section 9.1](https://datatracker.ietf.org/doc/html/rfc7515#section-9.1) or
   be a Public Name (a value that contains a Collision-Resistant Name)
```

We are using a collision resistant name.

signature-specification.md
gokarnm: Will reword.

signature-envelope-jws.md
gokarnm: Agreed, will address in next rev.

signature-envelope-jws.md
gokarnm: This is a good call out, we can mention that in signature spec, the optional [`time-secfrac` ](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6) MUST NOT be used.

https://datatracker.ietf.org/doc/html/rfc3339#section-5.3 
```
Rarely used options should be made mandatory or omitted
 for the sake of interoperability whenever possible. 
The format defined below includes only one rarely used option:
 fractions of a second. It is expected that this will be used only by applications which 
require strict ordering of date/time stamps or
which have an unusual precision requirement.
```

gokarnm:
@shizhMSFT would like to get your feedback before finalizing this spec.

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+    ],
+    "subject": {
+        "mediaType": "application/vnd.oci.image.manifest.v1+json",
+        "digest": "sha256:73c803930ea3ba1e54bc25c2bdc53edd0284c62ed651fe7b00369da519a3c333",
+        "size": 16724
+    },
+    "annotations": {
+        "io.cncf.notary.x509.fingerprint.sha256": "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\" \"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
+    }
+}
```
shizhMSFT:
```suggestion
{
    "artifactType": "application/vnd.cncf.notary.v2.signature",
    "blobs": [
        {
            "mediaType": "application/jose+json",
            "digest": "sha256:9834876dcfb05cb167a5c24953eba58c4ac89b1adf57f28f2f9d09af107ee8f0",
            "size": 32654
        }
    ],
    "subject": {
        "mediaType": "application/vnd.oci.image.manifest.v1+json",
        "digest": "sha256:73c803930ea3ba1e54bc25c2bdc53edd0284c62ed651fe7b00369da519a3c333",
        "size": 16724
    },
    "annotations": {
        "io.cncf.notary.x509.fingerprint.sha256": "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\",\"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
    }
}
```

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+
+## JWS Payload
+
+The payload in the JWS envelope is set to BASE64URL encoded value of [Notary v2 Payload](./signature-specification.md#payload).
+
+Example of payload before BASE64URL encoding.
+
+```jsonc
+{
+  "subject": {
```
shizhMSFT: This object has only one field `subject`. Instead of `application/vnd.cncf.notary.payload.v1+json`, should we sign `application/vnd.cncf.oras.artifact.descriptor.v1+json`?

shizhMSFT: By extending the above idea, I have a more generic thought. How about allowing all the media types for the payload? If the payload media type is recognizable like `application/vnd.cncf.oras.artifact.descriptor.v1+json`, we use the descriptor directly. If it is not, then we can generate a descriptor without annotations from the payload where
- Media Type: `<ProtectedHeader>.cty`
- Size: Length of the JWS Payload
- Digest: Hash of the JWS Payload

signature-envelope-jws.md
shizhMSFT: It seems `io.cncf.notary.signingTime` is a required header. Should it be in the `crit`?

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+### Create the *Payload to Sign*
+
+1. Compute the Base64Url value of ProtectedHeaders.
+2. Compute the Base64Url value of JWSPayload.
+3. Build message to be signed by concatenating the values generated in step 1 and step 2 using '.'
+`ASCII(BASE64URL(UTF8(ProtectedHeaders)) ‘.’ BASE64URL(JWSPayload))`
+
+### Generate the signature
+
+1. Compute the signature on the message constructed in the previous step by using the signature algorithm defined in the corresponding header element: `alg`.
```
shizhMSFT: We probably need to call out that the source of truth of the `signature algorithm` is the signing key, and the header element `alg` MUST match the `signature algorithm`.

shizhMSFT: We probably need to call out that `message` is the `Payload to Sign`.

signature-envelope-jws.md
shizhMSFT, line 127: What if we have multiple intermediate certs and CA certs? Are they allowed?

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+3. `alg` header values for various signature algorithms is a subset of values supported by [JWS][jws-alg-values]:
+
+  | Signature Algorithm             | `alg` Param Value |
+  | ------------------------------- | ----------------- |
+  | RSASSA-PSS with SHA-256         | PS256             |
+  | RSASSA-PSS with SHA-384         | PS384             |
+  | RSASSA-PSS with SHA-512         | PS512             |
+  | ECDSA on secp256r1 with SHA-256 | ES256             |
+  | ECDSA on secp384r1 with SHA-384 | ES384             |
+  | ECDSA on secp521r1 with SHA-512 | ES512             |
```
shizhMSFT: We probably need a caption or descriptive words for this table.

signature-specification.md
```diff
@@ -1,243 +1,118 @@
...
+  - **`mediaType`** (*string*): This REQUIRED property contains media type of signature envelope blob. Following values are supported
+    - `application/jose+json`
 - **`subject`** (*descriptor*): A REQUIRED artifact descriptor referencing the signed manifest, including, but not limited to image manifest, image index, oras-artifact manifest.
 - **`annotations`** (*string-string map*): This REQUIRED property contains metadata for the artifact manifest.
   It is being used to store information about the signature.
   Keys using the `io.cncf.notary` namespace are reserved for use in Notary and MUST NOT be used by other specifications.
-  - **`io.cncf.notary.x509certs.fingerprint.sha256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprint of signing certificate and certificate chain used for signature generation.
+  - **`io.cncf.notary.x509.fingerprint.sha256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprint of signing certificate and certificate chain used for signature generation.
```
shizhMSFT: How about `io.cncf.notary.x5t#S256` in [RFC7515 4.1.8](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.8)?

`x5t#S256` looks more common and shorter and it also defines the encoding (i.e. a base64url-encoded SHA-256 thumbprint). However, it seems only defines the cert for the signing key and is not a list.

signature-specification.md
```diff
@@ -1,243 +1,118 @@
...
+The client should be able to discover all the signatures belonging to an artifact (such as image manifest) by using [ORAS Manifest Referrers API][oras-artifacts-referrers].
 ORAS Manifest Referrers API returns a paginated list of all artifacts belonging to a target artifact (such as container images, SBoMs).
 The implementation can filter Notary signature artifacts by either using ORAS Manifest Referrers API or using custom logic on the client.
 Each Notary signature artifact refers to a signature envelope blob.
 
 ### Signature Filtering
 
 An OCI artifact can have multiple signatures, Notary v2 uses annotations of the signature artifact to filter relevant signatures based on the applicable trust policy.
-The Notary v2 signature artifact's `io.cncf.notary.x509certs.fingerprint.sha256` annotations key MUST contain the list of SHA-256 fingerprints of certificate and certificate chain used for signing.
+The Notary v2 signature artifact's `io.cncf.notary.x509.fingerprint.sha256` annotations key MUST contain the list of SHA-256 fingerprints of certificate and certificate chain used for signing.
```
shizhMSFT: Is the order of the certs in `io.cncf.notary.x509.fingerprint.sha256` the same as the order of the certs in `x5c`?

signature-specification.md
shizhMSFT: This sentence is ambiguous for me. The intention seems to be "the claim set MUST NOT be tampered once verified" but it is hard for me to get from it. Request for rewording.

signature-specification.md
```diff
@@ -1,243 +1,118 @@
...
-  Its value should be numeric representing the number of seconds (not milliseconds) since Epoch.
-- **Expiration time**: The time after which signature shouldn't be considered valid.
-  Its value should be numeric representing the number of seconds since Epoch.
-  This is an OPTIONAL attribute.
+Signed attributes/claims are additional metadata apart from the payload, which are required to support the signature verification process.
 
-### Unsigned Attributes
+- Any claims MUST NOT be stored/appended in the payload itself, as the payload is only parsed and processed once the signature has been verified (signature is valid, and from a trusted key) and trust is established.
+- Specific claims can be either required or optional.
+- Claims that if present, MUST be processed by a verifier MUST be marked as critical.
```
shizhMSFT: This sentence is ambiguous for me. Adding a pair of brackets enables me understand it faster:
```
Claims (that if present, MUST be processed by a verifier) MUST be marked as critical.
```
Request for rewording.

signature-specification.md
```diff
@@ -1,243 +1,118 @@
...
+Notary v2 requires the signature envelope to support the following signed attributes/claims.
 
-### Supported Signature Envelopes
+#### Standard attributes
 
-#### JWS JSON Serialization
+- **Signing time**: The time at which the signature was generated. Though this claim is signed by the signing key, it’s considered unauthenticated as a signer can modify local time and manipulate this claim. More details [here](#signing-time).
+- **Expiry** (critical): An OPTIONAL claim that provides a “best by use” time for the artifact, as defined by the signer. More details [here](#expiry).
+- **Content type** (critical): A REQUIRED clain that indicates the content type of the payload. Notary currently a payload that contains the OCI descriptor of a subject manifest as the payload, the supported value is `application/vnd.cncf.notary.payload.v1+json`, other payload types MAY be supported in future.
+- **Client identifier**: The version of a client (e.g. Notation) that produced the signature. This is an OPTIONAL claim. It uses the following format `{client}/{version}` e.g. “notation/1.0.0”. This claim in intended to be used for diagnostic and troubleshooting purposes.
```
shizhMSFT: Curious if `v` is allowed in `{version}` like `notation/v1.0.0`?

signature-specification.md
```diff
@@ -1,243 +1,118 @@
...
 
 ### Payload
 
-Notary v2 requires Payload to be the content **descriptor** of the subject manifest that is being signed.
+Notary v2 payload is a JSON document with media type `application/vnd.cncf.notary.payload.v1+json` and has following properties.
 
-1. Descriptor MUST contain `mediaType`, `digest`, `size` fields.
-1. Descriptor MAY contain `annotations` and if present it MUST follow the [annotation rules](https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules). Notary v2 uses annotations for storing both Notary specific and user defined signed attributes. The prefix `io.cncf.notary` in annotation keys is reserved for use in Notary v2 and MUST NOT be used outside this specification.
-1. Descriptor MAY contain `artifactType` field for artifact manifests, or the `config.mediaType` for `oci.image` based manifests.
+- `subject` : Required property whose value is the **OCI descriptor** of the subject manifest that is being signed, with media type of `application/vnd.cncf.oras.artifact.descriptor.v1+json`.
```
rgnote: > application/vnd.cncf.oras.artifact.descriptor.v1+json

The examples below have mediaType of the artifact that needs to be signed. Can you clarify (for my understanding), why we are saying the mediaType should be `application/vnd.cncf.oras.artifact.descriptor.v1+json` here?

signature-specification.md
rgnote: Are there any guidelines on how a signer should encode the TSA signature into an unsigned attribute?

signature-specification.md
rgnote: Should this be a signed attribute? Since it is for debugging purposes only, can it be unsigned?

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
+# JWS Signature Envelope
+
+JWS ([RFC7515](https://datatracker.ietf.org/doc/html/rfc7515)) is a JSON based envelope format for digital signatures over any type of payload (e.g. JSON, binary). Notary v2 uses JWS as a supported signature format. It specifically uses the *JWS JSON Serialization* representation, which supports unsigned attributes. JWS Compact serialization, and JWT (also based on Compact serialization), do not support unsigned attributes.
+
+## Storage
+
+A JWS based signature will be persisted with a blob `mediaType` of `"application/jose+json"`
```
rgnote: nit: signature -> signature envelope

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+
+Example
```
rgnote: nit: Signature Manifest Example:

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+```jsonc
+{
+    "timestamp": "<Base64(TimeStampToken)>",
+    "x5c": ["<Base64(DER(leafCert))>", "<Base64(DER(intermediateCACert))>", "<Base64(DER(rootCert))>"]
+}
+```
+
+- **`timestamp`** (*string*): This OPTIONAL property is used to store time stamp token.
+  Only [RFC3161]([rfc3161](https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2)) compliant TimeStampToken are supported.
+- **[`x5c`](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.6)** (*array of strings*): This REQUIRED property contains the list of X.509 certificate or certificate chain([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the JWS. The certificate or certificate chain is represented as a JSON array of certificate value strings. Each string in the array is a base64-encoded DER  certificate value. The certificate containing the public key corresponding to the key used to digitally sign the JWS MUST be the first certificate.
```
rgnote: > The certificate containing the public key corresponding to the key used to digitally sign the JWS MUST be the first certificate.

Let's call out that the order should be from `leaf -> root` with `root` being optional.

signature-envelope-jws.md
rgnote: Why do we need the root in the envelope? Isn't it redundant?

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
+# JWS Signature Envelope
+
+JWS ([RFC7515](https://datatracker.ietf.org/doc/html/rfc7515)) is a JSON based envelope format for digital signatures over any type of payload (e.g. JSON, binary). Notary v2 uses JWS as a supported signature format. It specifically uses the *JWS JSON Serialization* representation, which supports unsigned attributes. JWS Compact serialization, and JWT (also based on Compact serialization), do not support unsigned attributes.
```
rgnote: > JWS Compact serialization, and JWT (also based on Compact serialization), do not support unsigned attributes.

nit: Not relevant here. Can be moved to an FAQ.

signature-specification.md
letmaik: I think the chain has to be ordered, but the actual ordering is up to the serialization format.

signature-specification.md
letmaik: How about
> Claims that are present and MUST be processed by a verifier MUST be marked as critical.

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+Notary v2 supports only two unprotected headers: `timestamp` and `x5c`.
+
+```jsonc
+{
+    "timestamp": "<Base64(TimeStampToken)>",
+    "x5c": ["<Base64(DER(leafCert))>", "<Base64(DER(intermediateCACert))>", "<Base64(DER(rootCert))>"]
+}
+```
+
+- **`timestamp`** (*string*): This OPTIONAL property is used to store time stamp token.
```
letmaik: Is `"timestamp"` going to be registered as JWS header parameter? If not, shouldn't it be a collision resistant name like the protected header parameters?

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+  Only [RFC3161]([rfc3161](https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2)) compliant TimeStampToken are supported.
+- **[`x5c`](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.6)** (*array of strings*): This REQUIRED property contains the list of X.509 certificate or certificate chain([RFC5280](https://datatracker.ietf.org/doc/html/rfc5280)) corresponding to the key used to digitally sign the JWS. The certificate or certificate chain is represented as a JSON array of certificate value strings. Each string in the array is a base64-encoded DER  certificate value. The certificate containing the public key corresponding to the key used to digitally sign the JWS MUST be the first certificate.
+
+## Signature
+
+In JWS signature is calculated by combining JWSPayload and protected headers.
+The process is described below:
+
+### Create the *Payload to Sign*
```
letmaik: Can this be renamed to something not containing "Payload"? It's confusing given there's already the JWS payload. The JWS spec calls it "JWS Signing Input".

signature-specification.md
letmaik: I'm assuming ASN.1 DER encoded, then base64 encoded. Definitely has to be spelled out.

letmaik: Nevermind, this is a serialization detail that has to go into the JWS doc.

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+
+```jsonc
+{
+    "timestamp": "<Base64(TimeStampToken)>",
+    "x5c": ["<Base64(DER(leafCert))>", "<Base64(DER(intermediateCACert))>", "<Base64(DER(rootCert))>"]
+}
+```
+
+- **`timestamp`** (*string*): This OPTIONAL property is used to store time stamp token.
+  Only [RFC3161]([rfc3161](https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2)) compliant TimeStampToken are supported.
```
letmaik: I'm assuming ASN.1 DER encoded, then base64 encoded. Definitely has to be spelled out.
Also, RFC3161 signs an opaque datum, typically a hash of something. So this parameter definition also has to define what is actually being sent to the TSA for signing.

signature-envelope-jws.md
```diff
@@ -0,0 +1,158 @@
...
+    "cty": "application/vnd.cncf.notary.v2.jws.v1",
+    "io.cncf.notary.signingTime": "2022-04-06 07:01:20.52Z",
+    "io.cncf.notary.expiry": "2022-10-06 07:01:20.52Z",
+    "io.cncf.notary.clientID": "notation/1.0.0",
+    "crit":["io.cncf.notary.expiry"]
+}
+```
+
+- **[`alg`](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.1)**(*string*): This REQUIRED header defines which algorithm was used to generate the signature.JWS specification defines `alg` as a required header, that MUST be present and MUST be understood and processed by verifier. This [section](#supported-alg-header-values) lists the Notary v2 allowed subset of `alg` values supported by JWS.
+- **[`cty`](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.10)**(*string*): The REQUIRED header content-type is used to declare the media type of the secured content (the payload). The supported value is `application/vnd.cncf.notary.payload.v1+json`.
```
qmuntal, line 72: In the previous example, the content type is set to `application/vnd.cncf.notary.v2.jws.v1` instead.

qmuntal, line 72: In the previous example, the content type is set to `application/vnd.cncf.notary.v2.jws.v1` instead.

qmuntal (changes requested):

qmuntal (changes requested):

signature-specification.md
gokarnm: Yes, this detail should go in the specific envelope format doc.

signature-specification.md
gokarnm: I've updated this as ordered list from signing cert to root, which I think is a fair assumption for any cert chain (provided by a CA etc), irrespective of signature's serialization format.

signature-envelope-jws.md
gokarnm: I've used a wrapped payload, so that we can add in additional "attestation" and other payload related metadata additional claims within the wrapped payload.

gokarnm: > By extending the above idea, I have a more generic thought. How about allowing all the media types for the payload?

Isn't this assuming that all artifacts are being stored in a registry and therefore have a descriptor?

signature-envelope-jws.md
gokarnm: `io.cncf.notary.signingTime` is untrusted timestamp, and therefore not used in trust decisions, and omitted from `crit`.

signature-envelope-jws.md
gokarnm: Added a link to JWS public header section.

signature-envelope-jws.md
gokarnm: Updated that fractional seconds SHOULD NOT be used.

signature-envelope-jws.md
gokarnm: That's a good catch, will update to be collision resistant name.

signature-specification.md
gokarnm: > Curious if `v` is allowed in `{version}` like `notation/v1.0.0`?

Yes, I'll update this section that its an opaque string, and only make a suggestion about the format to be used (formatting is not validated).

signature-envelope-jws.md
gokarnm: Updated language to make this clearer.

signature-envelope-jws.md
gokarnm: Updated to "JWS Signing Input" as per suggestion.

signature-specification.md
gokarnm: Renamed to `io.cncf.notary.x509chain.thumbprint#S256`, let me know if you have another suggestion.

signature-specification.md
gokarnm: Added a line to clarify that the ordering is the same.

signature-specification.md
gokarnm: Updated to

>Claims that MUST be processed by a verifier MUST be marked as critical. Some claims may be optional and critical, i.e. they MUST be processed by a verifier only if they are present.

signature-specification.md
gokarnm: > Should this be a signed attribute? Since it is for debugging purposes only, can it be unsigned?

Changed this to an unsigned attribute, as this attribute is informational, and cannot be a signed attribute when a `envelope-generator` plugin is used.

signature-specification.md
gokarnm: I actually removed the media type, we support both [OCI descriptor](https://github.com/opencontainers/image-spec/blob/main/descriptor.md) and ORAS [artifact descriptors](https://github.com/oras-project/artifacts-spec/blob/main/descriptor.md) in the payload.

signature-envelope-jws.md
gokarnm: Addressed, with root being required.

signature-envelope-jws.md
gokarnm: The root cert isn't generally required for verification. The intent is for signature producers to always include full chain, which will be validated by signing tool (notation), so that consumers can set trusted root as root cert (our recommendation). We want to avoid partial chains in the envelope as tooling cannot determine if the partial chain was only one level below the root. We can relax this requirement if there is a pressing reason in future.

gokarnm: If you are referring to cross signed certificates, we currently don't support them, as they are used less frequently and are more difficult to validate using standard libraries.

signature-specification.md
gokarnm: I've reworded to 

>Any metadata that is used to verify the payload itself, and establish trust MUST be stored separately from the payload itself, as signed attributes or claims. Such metadata MUST NOT be stored/appended in the payload, as the payload is only parsed and processed once the signature has been verified and trust is established.

signature-envelope-jws.md
gokarnm: That's a very good point. I'll add a TODO here as this requires some more work.

gokarnm:
- Addressed review feedback
- Updated EC 512 to EC 521 to address feedback in notation-go [PR 46](https://github.com/notaryproject/notation-go/pull/46#discussion_r881537420)

qmuntal:
@gokarnm this PR changes some pieces of the signature spec that affect the wording of the plugin spec. The plugin spec should also be updated either in this PR or in a follow up. Examples:

- keySpec `EC_512` is mentioned several times: 
https://github.com/notaryproject/notaryproject/blob/cf6113ce44c25f89de67d62feebba67e2be7505b/specs/plugin-extensibility.md?plain=1#L278

-  payload and envelope types:
https://github.com/notaryproject/notaryproject/blob/cf6113ce44c25f89de67d62feebba67e2be7505b/specs/plugin-extensibility.md?plain=1#L323

gokarnm:
> @gokarnm this PR changes some pieces of the signature spec that affect the wording of the plugin spec. The plugin spec should also be updated either in this PR or in a follow up. Examples:
> 
>     * keySpec `EC_512` is mentioned several times:
>       https://github.com/notaryproject/notaryproject/blob/cf6113ce44c25f89de67d62feebba67e2be7505b/specs/plugin-extensibility.md?plain=1#L278
> 
>     * payload and envelope types:
>       https://github.com/notaryproject/notaryproject/blob/cf6113ce44c25f89de67d62feebba67e2be7505b/specs/plugin-extensibility.md?plain=1#L323

@qmuntal, totally agree, I want to merge this PR to unblock implementation, and address plugin spec updates as separate one. I was aware of the keyspec updates, thanks for pointing the payload type inconsistency.

qmuntal (approved):

gokarnm:
To be addressed in a subsequent PR
- [TSA request and response details](https://github.com/notaryproject/notaryproject/pull/158#discussion_r877977861)
- [Updates to plugin spec](https://github.com/notaryproject/notaryproject/pull/158#issuecomment-1138237156)

gokarnm:
@shizhMSFT @roywill @letmaik @qmuntal @rgnote @priteshbandi @jondonas thanks for all the great feedback! I've addressed most of it. I've resolved ones which are straightforward, and kept more contentious conversations open if you have more feedback. Would appreciate your review as this is blocking implementation.

priteshbandi (changes requested):
LGTM

rgnote:
LGTM

rgnote (approved):

signature-envelope-jws.md
```diff
@@ -0,0 +1,165 @@
...
+- Registered headers - `alg`, `cty`, and `crit`
+- [Public headers](https://datatracker.ietf.org/doc/html/rfc7515#section-4.2) with collision resistant names - `io.cncf.notary.signingTime`, `io.cncf.notary.expiry`
+
+Example
+
+```jsonc
+{
+    "alg": "PS384",
+    "cty": "application/vnd.cncf.notary.payload.v1+json",
+    "io.cncf.notary.signingTime": "2022-04-06 07:01:20Z",
```
letmaik: Not RFC 3339, missing the `T`.

signature-envelope-jws.md
```diff
@@ -0,0 +1,165 @@
...
+  If present, the value MUST be `["io.cncf.notary.expiry"]`.
+
+## Unprotected Headers
+
+Notary v2 supports following unprotected headers: `timestamp`, `x5c` and `io.cncf.notary.signingAgent`
+
+```jsonc
+{
+    "x5c": ["<Base64(DER(leafCert))>", "<Base64(DER(intermediateCACert))>", "<Base64(DER(rootCert))>"],
+    "io.cncf.notary.timestamptoken": "<Base64(TimeStampToken)>",
```
letmaik:
```suggestion
    "io.cncf.notary.timestamp": "<Base64(DER(TimeStampToken))>",
```

signature-envelope-jws.md
```diff
@@ -0,0 +1,165 @@
...
+The final signature envelope comprises of Payload, ProtectedHeaders, UnprotectedHeaders, and Signature, no additional top level fields are supported.
+
+Since Notary v2 restricts one signature per signature envelope, the compliant signature envelope MUST be in flattened JWS JSON format.
+
+```jsonc
+{
+    "payload": "<Base64Url(JWSPayload)>",
+    "protected": "<Base64Url(ProtectedHeaders)>",
+    "header": {
+        "io.cncf.notary.timestamp": "<Base64(TimeStampToken)>",
```
letmaik, line 126:
```suggestion
        "io.cncf.notary.timestamp": "<Base64(DER(TimeStampToken))>",
```

signature-specification.md
```diff
@@ -41,203 +32,111 @@ The signature manifest has an artifact type that specifies it's a Notary V2 sign
...
 
 ### Payload
 
-Notary v2 requires Payload to be the content **descriptor** of the subject manifest that is being signed.
+Notary v2 payload is a JSON document with media type `application/vnd.cncf.notary.payload.v1+json` and has following properties.
 
-1. Descriptor MUST contain `mediaType`, `digest`, `size` fields.
-1. Descriptor MAY contain `annotations` and if present it MUST follow the [annotation rules](https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules). Notary v2 uses annotations for storing both Notary specific and user defined signed attributes. The prefix `io.cncf.notary` in annotation keys is reserved for use in Notary v2 and MUST NOT be used outside this specification.
-1. Descriptor MAY contain `artifactType` field for artifact manifests, or the `config.mediaType` for `oci.image` based manifests.
+- `targetArtifact` : Required property whose value is the descriptor of the target artifact manifest that is being signed. Both [OCI descriptor][oci-descriptor] and [ORAS artifact descriptors][artifact-descriptor] are supported.
```
letmaik, line 86: How do you know which descriptor type is used? If the descriptor was signed directly, then `cty` of the envelope would indicate it, but now with the additional wrapping it seems this information is gone. Maybe I'm missing something though.

letmaik (changes requested):

signature-specification.md
gokarnm: [Line 87-89](https://github.com/notaryproject/notaryproject/pull/158/files#diff-06dd727cd694419f847f4f7b27b4121e910873b6db2ba32b967d36ae1999fb5fR87-R89) provide the requirements for the descriptor. Notary only depends on attributes common to both OCI and ORAS descriptor - `digest`, `size`, `mediaType` and `annotations`.

letmaik: Got it.

signature-envelope-jws.md
gokarnm: Got some more feedback on this, plan to rename to `io.cncf.notary.timestampSignature` to indicate it's a countersignature.

gokarnm: Isn't the value of RFC3161 TimeStampToken already DER encoded, do we need to specify it additionally here?

gokarnm: I guess it's confusing because we mention DER encoding for the cert chain (x5c), which could originally be in PEM or DER format.

signature-envelope-jws.md
gokarnm: RFC 3339 allows a space separator [[1](https://ijmacd.github.io/rfc3339-iso8601/)]. I'll update the example though to commonly used representation.

https://datatracker.ietf.org/doc/html/rfc3339#section-5.6

> NOTE: ISO 8601 defines date and time separated by "T".
      Applications using this syntax may choose, for the sake of
      readability, to specify a full-date and full-time separated by
      (say) a space character.

signature-envelope-jws.md
letmaik: TimeStampToken is defined in terms of ASN.1, and section 3 specifies encodings for multiple transports like email, files, HTTP. All of them use DER encoding, and then some of them base64 encode after that. So yes, it's very likely DER already, but given that we define a new context I think we need to specify the ASN.1 encoding as well.

signature-envelope-jws.md
```diff
@@ -0,0 +1,166 @@
...
+
+In JWS signature is calculated by combining JWSPayload and protected headers.
+The process is described below:
+
+### Create the *JWS Signing Input*
+
+1. Compute the Base64Url value of ProtectedHeaders, this is the value of `protected` property in the signature envelope.
+1. Compute the Base64Url value of JWSPayload, this is the value of `payload` property in the signature envelope.
+1. Build *JWS Signing Input* to be signed by concatenating the values generated in step 1 and step 2 using '.'
+`ASCII(BASE64URL(UTF8(ProtectedHeaders)) ‘.’ BASE64URL(JWSPayload))`
```
shizhMSFT, line 105: The `Base64Url` here includes paddings or not? It seems `Base64Url` is never defined in this documentation.

signature-envelope-jws.md
shizhMSFT: Time stamp token is a [CMS signed-data](https://datatracker.ietf.org/doc/html/rfc3852#section-5) and is BER encoded. If we require DER here, the client needs to convert the BER encoded timestamp token to DER encoded. Is it intended?

signature-specification.md
```diff
@@ -93,26 +117,26 @@ Notary v2 payload is a JSON document with media type `application/vnd.cncf.notar
...
+- **Timestamp signature** : An OPTIONAL counter signature which provides [trusted timestamp](#signing-time)e.g. Time Stamp Authority (TSA) generated timestamp signature. Only [RFC3161](ietf-rfc3161) compliant TimeStampToken are currently supported.
+- **Signing Agent**: An OPTIONAL claim that provides the identifier of the software (e.g. Notation) that produced the signature on behalf of the user. It is an opaque string set by the software that produces the signature. It's intended primarily for diagnostic and troubleshooting purposes, this attribute is unsigned, the verifier MUST NOT validate formatting, or fail validation based on the content of this claim. The suggested format is one or more tokens of the form `{id}/{version}` containing identifier and version of the software, seperated by spaces. E.g. “notation/1.0.0”, “notation/1.0.0 com.example.nv2plugin/0.8”.
```
shizhMSFT, line 139: Should this part be signed?

signature-envelope-jws.md
letmaik: You may be right, I was looking at RFC 3161 which only says DER, not BER. Let's postpone any timestamp-related changes to a follow-up PR.

signature-envelope-jws.md
gokarnm: Good point, clarified that JWS uses base64url without padding as defined in *Base64url Encoding* in [RFC 7515 section 2](https://datatracker.ietf.org/doc/html/rfc7515#section-2).

signature-specification.md
gokarnm: *Signing Agent* could be signed, which makes it tamper proof. But it does not play a part in signature verification, only diagnostics, so it's not required to be signed.
 Notation's default signing and *Signature Generator* plugins can support this as a signed attribute, but *Envelope generator* plugin can't. I went with unsigned for simplicity, but we can change it such that it can be a signed attribute where possible if there is a strong case for it.

signature-envelope-jws.md
gokarnm: [*Time-Stamp Protocol via HTTP*](https://datatracker.ietf.org/doc/html/rfc3161#section-3.4) in RFC 3161 specifies that the TSA response is ASN.1 DER-encoded. Validated with a [freetsa.org](https://www.freetsa.org) example [here](https://lapo.it/asn1js/#MIIVcjADAgEAMIIVaQYJKoZIhvcNAQcCoIIVWjCCFVYCAQMxDzANBglghkgBZQMEAgMFADCCAaQGCyqGSIb3DQEJEAEEoIIBkwSCAY8wggGLAgEBBgQqAwQBMFEwDQYJYIZIAWUDBAIDBQAEQAy_TK7zgEe7qaJOYhqWFITl0qkhdqhZ5-sn3zQ90065jVOKbF9Noc4wLsJQuCHMAB5GzJenBJiClxhaTffplgICBAHVxjwYDzIwMjIwNTI3MjMwNzU2WgEB_6CCARGkggENMIIBCTERMA8GA1UEChMIRnJlZSBUU0ExDDAKBgNVBAsTA1RTQTF2MHQGA1UEDRNtVGhpcyBjZXJ0aWZpY2F0ZSBkaWdpdGFsbHkgc2lnbnMgZG9jdW1lbnRzIGFuZCB0aW1lIHN0YW1wIHJlcXVlc3RzIG1hZGUgdXNpbmcgdGhlIGZyZWV0c2Eub3JnIG9ubGluZSBzZXJ2aWNlczEYMBYGA1UEAxMPd3d3LmZyZWV0c2Eub3JnMSIwIAYJKoZIhvcNAQkBFhNidXNpbGV6YXNAZ21haWwuY29tMRIwEAYDVQQHEwlXdWVyemJ1cmcxCzAJBgNVBAYTAkRFMQ8wDQYDVQQIEwZCYXllcm6gghAIMIIIATCCBemgAwIBAgIJAMHphhYNqOmCMA0GCSqGSIb3DQEBDQUAMIGVMREwDwYDVQQKEwhGcmVlIFRTQTEQMA4GA1UECxMHUm9vdCBDQTEYMBYGA1UEAxMPd3d3LmZyZWV0c2Eub3JnMSIwIAYJKoZIhvcNAQkBFhNidXNpbGV6YXNAZ21haWwuY29tMRIwEAYDVQQHEwlXdWVyemJ1cmcxDzANBgNVBAgTBkJheWVybjELMAkGA1UEBhMCREUwHhcNMTYwMzEzMDE1NzM5WhcNMjYwMzExMDE1NzM5WjCCAQkxETAPBgNVBAoTCEZyZWUgVFNBMQwwCgYDVQQLEwNUU0ExdjB0BgNVBA0TbVRoaXMgY2VydGlmaWNhdGUgZGlnaXRhbGx5IHNpZ25zIGRvY3VtZW50cyBhbmQgdGltZSBzdGFtcCByZXF1ZXN0cyBtYWRlIHVzaW5nIHRoZSBmcmVldHNhLm9yZyBvbmxpbmUgc2VydmljZXMxGDAWBgNVBAMTD3d3dy5mcmVldHNhLm9yZzEiMCAGCSqGSIb3DQEJARYTYnVzaWxlemFzQGdtYWlsLmNvbTESMBAGA1UEBxMJV3VlcnpidXJnMQswCQYDVQQGEwJERTEPMA0GA1UECBMGQmF5ZXJuMIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEAtZEEjE5IbzTp3Ahif8I3UWIjaYS4LLEwvv9RfPw4-EvOXGWodNqyYhrgvOfjNWPg7ek0_V-IIxWfB4SICCJ0YMHtiCYXBvQoEzQ1nfu4G9E1P8F5YQrxqMjIZdwA6iOzqJvmvQO6hansgn1gVlkF4i1qWE7ROArhUCgM7jl-mKAS84BGQAeGJEO8B3y5X0Ia8xcS2Wg8223_uvPIululZq5SPUWdYXc0bU2EDieIa3wBxbiQ14ouJ7uo3S-aKBLhV9YvkhxlliVIBp3Nt9Bt4YHeDpVw1m-HIgzii2KKtVkG8-4MIQ9wUej0hYr4uaktCeRq8tnLpb_PrRaM32BEkaSwZgOxFMr3Ax8GXn7u-lPFdfNJDAWdLjLdx2rE1MTHEGg7l_0b5ZG8YQVRhtiPmgORswe2-R7ZVNqjb5rNah4Uqi5K3xdGS1TbGNu2_-MAgCRlRzcENs5Od7rl3m_g8_nW5_--tGHnlOkvsJUfiq5hpBLM6bIQdGNci-MnrhoPa0pkbrD4RjvGO_hFUwQ10Z6AJRHsn2bDSWlS2L7LabCqTUxB9gUV_n3LuJMZzdpZumrqS-POrnGOb8tszX25_FC7FbEvNmWwqjByicLm3UsRHOSLotnv21prmlBgaTNPs09vx64zDws0IIqsgN8yZv3ZBGWHa6LLiY2VBTFbbnsCAwEAAaOCAdswggHXMAkGA1UdEwQCMAAwHQYDVR0OBBYEFG52C3tOT5zhYMptLOknoqKUs3c3MB8GA1UdIwQYMBaAFPpVDYw0ZlFDTPfns6dsla965qSXMAsGA1UdDwQEAwIGwDAWBgNVHSUBAf8EDDAKBggrBgEFBQcDCDBjBggrBgEFBQcBAQRXMFUwKgYIKwYBBQUHMAKGHmh0dHA6Ly93d3cuZnJlZXRzYS5vcmcvdHNhLmNydDAnBggrBgEFBQcwAYYbaHR0cDovL3d3dy5mcmVldHNhLm9yZzoyNTYwMDcGA1UdHwQwMC4wLKAqoCiGJmh0dHA6Ly93d3cuZnJlZXRzYS5vcmcvY3JsL3Jvb3RfY2EuY3JsMIHGBgNVHSAEgb4wgbswgbgGAQAwgbIwMwYIKwYBBQUHAgEWJ2h0dHA6Ly93d3cuZnJlZXRzYS5vcmcvZnJlZXRzYV9jcHMuaHRtbDAyBggrBgEFBQcCARYmaHR0cDovL3d3dy5mcmVldHNhLm9yZy9mcmVldHNhX2Nwcy5wZGYwRwYIKwYBBQUHAgIwOxo5RnJlZVRTQSB0cnVzdGVkIHRpbWVzdGFtcGluZyBTb2Z0d2FyZSBhcyBhIFNlcnZpY2UgKFNhYVMpMA0GCSqGSIb3DQEBDQUAA4ICAQClyUTixvrAoU2TCn_QoLFytB_BSDw-lXxoorzZuXZPGpUBYf1yRy1Bpe7Sd3hiA7VCIkD7OibN4XYIe2-xAR30zBniVxqkoFEQlmXpTEb1C9Kt7mrEE34lGyWjnavaRRUV2P-eByCejsILeHT34aDt58AJN_6EozT4syZc7S2O2d9hOWWDZ3_rOCwe47I-bqXwXfMN57n4kAXSUmb2EvOci09tq6bXv7rBljK5Bjcyn1Km8GahDkPqqB-Emmxf4_6LXqIydfaH8gUuUC6mwwdipmjM4Hhx3Y6X4xW7qSniVYmXegoxLOlsUQaxQ3x3nys2GxgoiPPuiiNDdPoGPpVhkmJ_fEMQc5ZdEmCSjroAnoA0Ka4yTPlvBCNU83vKWv3cefeTRqs4i_x58B3JhhJU6mzBKZQQdrg9IFVvO-UTJoN_KHb3gzs3Dnw9QQUjgn1PU0AMciGNdSKf8QxviJOpo6HAxCu0yJjBPfQcf2VztPxWUVlxphCnsNKFfIIlqfsgTqzsouiXGqGvh4hqKuPHL-CgquhCmAp3vvFrkhFUWAkNmCtZRmA3ZOdaCtPRFFS5mG9ni5q2r-hJcDOuOr_U60O3vJ3uaIFZSeZIFYKoLnhSd_IoIQfv45AgDgUIrLjqguolBSdvPJ2io9O0rTi7-IQr2jb8JEgpH1WNwC3R4DCCB_8wggXnoAMCAQICCQDB6YYWDajpgDANBgkqhkiG9w0BAQ0FADCBlTERMA8GA1UEChMIRnJlZSBUU0ExEDAOBgNVBAsTB1Jvb3QgQ0ExGDAWBgNVBAMTD3d3dy5mcmVldHNhLm9yZzEiMCAGCSqGSIb3DQEJARYTYnVzaWxlemFzQGdtYWlsLmNvbTESMBAGA1UEBxMJV3VlcnpidXJnMQ8wDQYDVQQIEwZCYXllcm4xCzAJBgNVBAYTAkRFMB4XDTE2MDMxMzAxNTIxM1oXDTQxMDMwNzAxNTIxM1owgZUxETAPBgNVBAoTCEZyZWUgVFNBMRAwDgYDVQQLEwdSb290IENBMRgwFgYDVQQDEw93d3cuZnJlZXRzYS5vcmcxIjAgBgkqhkiG9w0BCQEWE2J1c2lsZXphc0BnbWFpbC5jb20xEjAQBgNVBAcTCVd1ZXJ6YnVyZzEPMA0GA1UECBMGQmF5ZXJuMQswCQYDVQQGEwJERTCCAiIwDQYJKoZIhvcNAQEBBQADggIPADCCAgoCggIBALYCjg4wMvERENlkzalLnQJ44ZQq6ROqpZkHzaaXk5lb2ax-M7rZ_jcE2hwBqY0hr-P1kaWdcGdwUWeZj1AWci4KtGKyH0ORcdLPzEWT83Na95SlqzEfbAEMeJjeM9dcRRDudvS9HRSYzxfTA_BqXdn3lsxsqbZXpW_j6k_vvnzmtqGNPjWjDO5f8XDRzzmjM9P9qJZNIttoWynlYb6JDwqoRYc7LoSrJquDn_6PrenSO7MeYdJzzJuIBkkYX6vs-gU0YAq6kBthTi6FRYLeoiJvwZzX31K-1Q2Hd82ZiMBTo_x9wyh6BopP8StxPNmANmbpVThUVv84-AKYz2uThW6SJHdKZs8c3RHC-O_YUgPXRYslZksT7WOc3tT_gRPWzFNT0nKUc8PDBxV8ciqltd0L-y1sOLG5N0nIgexgAm0IlRs4JL1xusvORzrr1jbwuRi0osj_RpTwdFevLW8c-CVU0XcP15_10xTc0QTN3KvJQTgFbfzwF-frhXL9UvcBRPGI2gX1gj9Y3QYpfnOHvtLXcsE9qCZmAQRf5BLdcJhsDJh7pzRLkDc4dRbSWOeIW1H4lot_JgEhO8TLTIX4_wuEr2qYgzfN-4GGj37PMdymcW1-wt2ALBZyYp5cAFLLNX3Smq_EP2FbOx_51OHOCMccc-H-u33FajNiEynp7WwjAgMBAAGjggJOMIICSjAMBgNVHRMEBTADAQH_MA4GA1UdDwEB_wQEAwIBxjAdBgNVHQ4EFgQU-lUNjDRmUUNM9-ezp2yVr3rmpJcwgcoGA1UdIwSBwjCBv4AU-lUNjDRmUUNM9-ezp2yVr3rmpJehgZukgZgwgZUxETAPBgNVBAoTCEZyZWUgVFNBMRAwDgYDVQQLEwdSb290IENBMRgwFgYDVQQDEw93d3cuZnJlZXRzYS5vcmcxIjAgBgkqhkiG9w0BCQEWE2J1c2lsZXphc0BnbWFpbC5jb20xEjAQBgNVBAcTCVd1ZXJ6YnVyZzEPMA0GA1UECBMGQmF5ZXJuMQswCQYDVQQGEwJERYIJAMHphhYNqOmAMDMGA1UdHwQsMCowKKAmoCSGImh0dHA6Ly93d3cuZnJlZXRzYS5vcmcvcm9vdF9jYS5jcmwwgc8GA1UdIASBxzCBxDCBwQYKKwYBBAGB8iQBATCBsjAzBggrBgEFBQcCARYnaHR0cDovL3d3dy5mcmVldHNhLm9yZy9mcmVldHNhX2Nwcy5odG1sMDIGCCsGAQUFBwIBFiZodHRwOi8vd3d3LmZyZWV0c2Eub3JnL2ZyZWV0c2FfY3BzLnBkZjBHBggrBgEFBQcCAjA7GjlGcmVlVFNBIHRydXN0ZWQgdGltZXN0YW1waW5nIFNvZnR3YXJlIGFzIGEgU2VydmljZSAoU2FhUykwNwYIKwYBBQUHAQEEKzApMCcGCCsGAQUFBzABhhtodHRwOi8vd3d3LmZyZWV0c2Eub3JnOjI1NjAwDQYJKoZIhvcNAQENBQADggIBAGivfr-ThWLvTOs7WAvi-vbMNaJncpYvPZWQH6VjDIfQkZiYTOigajP4qcKC7Z8csRrGwj4XEI7k785vspTelcEzJiJVclUiymGXHUo7f3glDfuNSu7A-xlZsWQQBSC5wQ5kxiZi5K1NCrriKY_JSPxOmejZ5rj9vkQEEh7HwUIurLLJ1zKOBzluYLTzu4A61KVVyA_vtT-F53ZKCp-0r8OZ9M0vX79YcQXGCBzz0FM3trt9GwELdJ9IiMkS82lrobaQLXe338BGwEoMwexPjRheLaVd-3vCogNsYhkkak-Z3btvH4KTmPO4A9wK2Q3LWb70wnx3QEuZBDt4JxhnmRFSw5nxLL_ExiWtwJY1WuRONCEA7FF6UC4vBvlAuNQ1mbvBFU-K52GgsNVV-0oTkdTzQgr42_EvLX3bnXfc4VN4BAdK8XXk8tbVWzS11vfcvdMXMK9WSA1MDP8UP56DvBUYZtC6Dwu9xH_ieGQXa71sGrhd8yXt93eIm8RHG_P6c-VsxZHosWDNp7B4ah7ASsOyT6LijV0Z5eSABNXhZqg8guxv1U-zheuvcTOoW1LeRttSROHDSujTbnEvn84NST19Pt1YbGGY4-w-bpY0b0F6yfIh4K_zOo9qCx70wCNjC3atqo2RQzgl7MQcSaW5ixgcfaMOmXq5VMc8LNgFr9qZMYIDijCCA4YCAQEwgaMwgZUxETAPBgNVBAoTCEZyZWUgVFNBMRAwDgYDVQQLEwdSb290IENBMRgwFgYDVQQDEw93d3cuZnJlZXRzYS5vcmcxIjAgBgkqhkiG9w0BCQEWE2J1c2lsZXphc0BnbWFpbC5jb20xEjAQBgNVBAcTCVd1ZXJ6YnVyZzEPMA0GA1UECBMGQmF5ZXJuMQswCQYDVQQGEwJERQIJAMHphhYNqOmCMA0GCWCGSAFlAwQCAwUAoIG4MBoGCSqGSIb3DQEJAzENBgsqhkiG9w0BCRABBDAcBgkqhkiG9w0BCQUxDxcNMjIwNTI3MjMwNzU2WjArBgsqhkiG9w0BCRACDDEcMBowGDAWBBSRbaPYYOzKguNLxZ0Xk-fpaIdfFDBPBgkqhkiG9w0BCQQxQgRAO-_WBAh7yMfQPSxDF5avIcZjm32iWF6Z4TYhmVof2CfqUZjRwJpV9L4jbQQuaG0wz8GOr5QimT8N9bso--piwjANBgkqhkiG9w0BAQEFAASCAgBaD6X7uR_SR6jJgzL7rgfw3ESroojyNVy_9EyfFWvPHQLSS22GvwW_6zwNgAMPpDwa5tB9k-aU4Xg92vHcGpABKMlpE0QTEbUVjcv-CU1YNaSaMdUsqY2rspqFaRkVSYaoKIAi8Arza6tgNNqmL0EasoomYGmIG3zqKhztrzvptiuwVsIssOTUQBEDVxsUshIOn50Z3HfZvaa9St-ndFl7Bj63O_bBNkxTJGYHDjbhWfeB1MK3oKt_j6UoZGVjgwhRs6iOLgn_P0_WuId3UjymYIfmTqfI6HhaevZff-PCiFR-QMidJzltjoiEVSit93PIDcRSZrgVoB_kHxZ82vVP13bKkKeJzos9j0Ly7rCNCIDG7UCrQ-6nVqPqYWpIlFfDdV5x-gYsETb_OkAEs9UzMmYvQ5rCQZnZl4XMLzSCjx-FEuLnVeoyqNVR56TVo5miGsM1j212lPfTc6is4rG54pwkSKvenSV36qeE8aJEkXAAB8-hb65xCNIOatjJc1tzo8h_ksC0YNQ2LodCcREGel0e2GaY4RDtZsVgUqZkLLIEiUzzhm8iJ8DiNIsUrmAjDxyYqzWn6uQJdmoUWr8-S3abHGtm97AVr3xWuAcEoK34ZZ3_cvINhaFxoumJnE94njtpQ_F-l0Sf5uj_TO0_HEPRtGhdLgm4BVD1me9g-w).

So the current content `"io.cncf.notary.timestamp": "<Base64(TimeStampToken)>"` is accurate, we don't specifically require DER, it's whichever format TSA servers respond with. I'm going to defer changes to timestamp related details, we can refine when we do the implementation. From the signature format's perspective it's an opaque token that is separately validated.

roywill (approved):

shizhMSFT (approved):
LGTM

gokarnm:
Resolved merge conflicts.

signature-envelope-jws.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Storage
+
+A JWS signature envelope will be stored in an OCI registry as a blob, and referenced in the signature manifest as a blob with `mediaType` of `"application/jose+json"`.
+
+Signature Manifest Example
+
+```jsonc
+{
+    "artifactType": "application/vnd.cncf.notary.signature",
```
SteveLasker, line 13: need version here: `application/vnd.cncf.notary.signature.v2`

PR #164, opened 2022-07-05
Support multiple trust stores and custom verification level

gokarnm:
- Updated trust policy to support multiple trust stores (which was supported in initial version of policy). Without this support, users are forced to combine multiple trust stores themselves if they wish to reference two or more logical trust stores (e.g. an organizational CA and a public CA trust store).
- Added support for custom signature verification level. This allows users to have fine grained control of all the validations performed during signature verification.
- Updated terminology
  - Validation check -> Validation
  - Trusted timestamp -> Authentic timestamp


gokarnm:
cc: @shizhMSFT @roywill @priteshbandi @rgnote for review

trust-store-trust-policy-specification.md
```diff
@@ -146,37 +145,70 @@ Trust policy for the scenario where ACME Rockets uses artifacts signed by themse
...
-  - **`trustStore`**(*string*): This REQUIRED property specifies a named trust store. Uses the format `{trust-store-type}:{named-store}`. Currently supported values for `trust-store-type` are `ca` and `tsa`. **NOTE**: When support for publicly trusted TSA is available, `tsa:publicly-trusted-tsa` is the default value, and implied without explictly specifying it. If a custom TSA is used the format `ca:acme-rockets,tsa:acme-tsa` is supported to specify it.
+  - **`verificationLevel`**(*string*): This OPTIONAL property dictates how signature verification is performed. Either `verificationLevel` or `verification` property MUST be specified. Supported values are `strict`, `permissive`, `audit` and `skip`. Detailed explaination of each value is present [here](#signatureverification-details).
+  - **`verification`**(*object*): This OPTIONAL property dictates how signature verification is performed using a [custom verification level](#custom-verification-level). Either `verificationLevel` or `verification` property MUST be specified.
```
shizhMSFT: What if `verificationLevel` and `verification` both exist?

trust-store-trust-policy-specification.md
```diff
@@ -146,37 +145,70 @@ Trust policy for the scenario where ACME Rockets uses artifacts signed by themse
...
+  - **`trustStores`**(*array of string*): This REQUIRED property specifies a set of one or more named trust stores, each of which contain the trusted roots against which signatures are verified. Each named trust store uses the format `{trust-store-type}:{named-store}`. Currently supported values for `trust-store-type` are `ca` and `tsa`.
+    - **NOTE**: When support for publicly trusted TSA is available, `tsa:publicly-trusted-tsa` is the default value, and implied without explictly specifying it. If a custom TSA is used the format `ca:acme-rockets,tsa:acme-tsa` is supported to specify it.
```
shizhMSFT: Should ``ca:acme-rockets,tsa:acme-tsa`` be `["ca:acme-rockets","tsa:acme-tsa"]`?

trust-store-trust-policy-specification.md
gokarnm: The intent was only one would be supported so that would be a trust policy validation error. Now it's simplified to a single `signatureVerification` property which is multi typed.

trust-store-trust-policy-specification.md
gokarnm: The `ca:acme-rockets,tsa:acme-tsa` format allow you to group related trust stores.

e.g
[
 "ca:acme-rockets,tsa:acme-tsa",
 "ca:wabbit-networks" // Publicly trusted TSA implied
]

trust-store-trust-policy-specification.md
shizhMSFT: Multi-typed property is hard to be implemented correctly. Can we avoid multi-typed properties?

gokarnm: Agree with the additional complexity, but other options make the user experience less desirable. You either require two separate properties or set value to an object even for simple configurations. We are using the multi-typed property in this case as we see a benefit in user experience at the cost of implementation complexity.

Alternate 1 - Use one of the following to set a level, or to override a level. Uses two different top level properties which may be bit confusing for users.
```
     "signatureVerificationLevel" : "strict"

      OR

    "signatureVerification": {
      "level" : "strict",
      "override" : {
        "expiry" : "log"
      }
    }
```

Alternate 2 - use the same object definition, but makes the simple configuration which most users will use, more complicated.
```
"signatureVerification": {
     "level" : "strict"
}

OR

  "signatureVerification": {
    "level" : "strict",
    "override" : { 
      "expiry" : "log"
    }
  }

```

trust-store-trust-policy-specification.md
priteshbandi: Should we mention that white spaces before and after of `,` and `"` and  are ignored?

rgnote (approved):

signature-envelope-jws.md
```diff
@@ -88,7 +88,7 @@ Notary v2 supports following unprotected headers: `timestamp`, `x5c` and `io.cnc
...
-- **`io.cncf.notary.timestampSignature`** (*string*): This OPTIONAL header is used to store countersignature that provides trusted signing time. Only [RFC3161]([rfc3161](https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2)) compliant `TimeStampToken` are supported.
+- **`io.cncf.notary.timestampSignature`** (*string*): This OPTIONAL header is used to store countersignature that provides authentic signing time. Only [RFC3161]([rfc3161](https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2)) compliant `TimeStampToken` are supported.
   - **TODO** Define the opaque datum (hash of envelope) that is sent to TSA, and how TSA response (time stamp token) is represented in this header.
 - **`io.cncf.notary.signingAgent`**(*string*): This OPTIONAL header provides the identifier of a client (e.g. Notation) that produced the signature. E.g. “notation/1.0.0”. Refer [*Signing Agent* unsigned attribute](signature-specification.md#unsigned-attributes) for more details.
```
SteveLasker, line 119: What's the purpose of this? Does it really matter what was used to sign?

signature-specification.md
```diff
@@ -135,7 +135,7 @@ Notary v2 requires the signature envelope to support the following signed attrib
...
-- **Timestamp signature** : An OPTIONAL counter signature which provides [trusted timestamp](#signing-time)e.g. Time Stamp Authority (TSA) generated timestamp signature. Only [RFC3161](ietf-rfc3161) compliant TimeStampToken are currently supported.
+- **Timestamp signature** : An OPTIONAL counter signature which provides [authentic timestamp](#signing-time)e.g. Time Stamp Authority (TSA) generated timestamp signature. Only [RFC3161](ietf-rfc3161) compliant TimeStampToken are currently supported.
 - **Signing Agent**: An OPTIONAL claim that provides the identifier of the software (e.g. Notation) that produced the signature on behalf of the user. It is an opaque string set by the software that produces the signature. It's intended primarily for diagnostic and troubleshooting purposes, this attribute is unsigned, the verifier MUST NOT validate formatting, or fail validation based on the content of this claim. The suggested format is one or more tokens of the form `{id}/{version}` containing identifier and version of the software, seperated by spaces. E.g. “notation/1.0.0”, “notation/1.0.0 com.example.nv2plugin/0.8”.
```
SteveLasker, line 142: Same as https://github.com/notaryproject/notaryproject/pull/164/files#r918406206
Why is it duplicated?
This description at least provides some details around troubleshooting. I really like troubleshooting data. But, I imagine many internal tools may not wish to publish their toolset. 

Just curious if we really need this.

trust-store-trust-policy-specification.md
```diff
@@ -71,7 +71,7 @@ Trust policy for a simple scenario where ACME Rockets uses only artifacts signed
             "name": "wabbit-networks-images",
             "registryScopes": [ "*" ],
             "signatureVerification" : "audit",
-            "trustStore": "ca:acme-rockets",
+            "trustStores": ["ca:acme-rockets"],
```
SteveLasker: Good to see it's an array, but should we provide an example of two `["ca:acme-rockets-staging", "ca:acme-rockets-prod"]`

trust-store-trust-policy-specification.md
```diff
@@ -95,7 +95,7 @@ Trust policy for the scenario where ACME Rockets uses artifacts signed by themse
               "registry.acme-rockets.io/software/net-logger"
             ],
             "signatureVerification" : "strict",
-            "trustStore": "ca:acme-rockets",
+            "trustStores": ["ca:acme-rockets"],
```
SteveLasker: Good to see it's an array, but should we provide an example of two `["ca:acme-rockets-staging", "ca:acme-rockets-prod"]`

trust-store-trust-policy-specification.md
```diff
@@ -111,15 +111,14 @@ Trust policy for the scenario where ACME Rockets uses artifacts signed by themse
...
-              "override" :
-              {
-                "expiry" : "skip"
+              "override" : {
+                "expiry" : "log",
+                "revocation" : "skip"
               }
             },
-            "trustStore": "ca:acme-rockets",
+            "trustStores": ["ca:acme-rockets"],
```
SteveLasker, line 121: Should we provide an example of two `["ca:acme-rockets-staging", "ca:acme-rockets-prod"]`

trust-store-trust-policy-specification.md
```diff
@@ -128,7 +127,7 @@ Trust policy for the scenario where ACME Rockets uses artifacts signed by themse
             "name": "global-policy-for-all-other-images",
             "registryScopes": [ "*" ],
             "signatureVerification" : "audit",
-            "trustStore": "ca:acme-rockets",
+            "trustStores": ["ca:acme-rockets"],
```
SteveLasker: Should we provide an example of two `["ca:acme-rockets-staging", "ca:acme-rockets-prod"]`

trust-store-trust-policy-specification.md
```diff
@@ -146,37 +145,70 @@ Trust policy for the scenario where ACME Rockets uses artifacts signed by themse
...
+ Signature verification is a multi step process performs the following validations - integrity (artifact is unaltered, signature is not corrupted), authenticity (the signature is really from the identity that claims to have signed it), trusted timestamping (the signature was generated when the key/certificate were unexpired), expiry (an optional check if the artifact specifies an expiry time), revocation check (is the signing identity still trusted at the present time). Based on the signature verification level, each of these validations is *enforced* or *logged*. If a validation is *enforced*, a failure is treated as critical failure, and causes the overall signature verification to fail.  If a validation is *logged*, a failure causes details to be logged, and the next validation is evaluated till all validations succeed or a critical failure is encountered. NOTE: Implementations may change the ordering of these validations based on efficiency, but all validation MUST to be performed till the first critical failure is encountered, or all validation succeed, for the overall signature verification process to be considered complete.
```
SteveLasker: Nit: break each sentence into a separate line for reviewing.

SteveLasker: > but all validation _MUST to be performed_ till the first critical failure is encountered,
Clarification of wording.

SteveLasker: Suggest breaking these up into bullets.

trust-store-trust-policy-specification.md
```diff
@@ -146,37 +145,70 @@ Trust policy for the scenario where ACME Rockets uses artifacts signed by themse
...
-- `permissive` : The `permissive` level enforces most signature verification checks, but will only logs failures for revocation and expiry. The `permissive` level is recommended to be used if signature verification is done at deploy time or runtime, and the user only needs integrity and authenticity guarantees.
-- `audit` : The `audit` level only enforces integrity check if a signature is present. Failure of all other checks are only logged.
+- `strict` : Signature verification is performed at `strict` level, which enforces all validations. If any of these validations fail, the signature verification fails. This is the recommended level in environments where a signature verification failure does not have high impact to other concerns (like application availability). It is recommended that build and development environments where images are initially injested, or for high assurance at deploy time use `strict` level.
```
SteveLasker, line 182: A bit redundant, and I'd suggest moving recommended to the front. But, I worry about conflating MUST, MAY, etc.
Suggestion:

`strict` : RECOMMENDED: Signature verification enforces all validations. If any of the validations fail, ...

SteveLasker: For the details of recommended, we should think about teasing out the deployment verification check, which should be strict, but runtime may be log.
For instance, you shouldn't deploy something you know is bad. However, once it's deployed, you don't want to kill a process if the signature is later determined to be "bad".
I think that's what you're getting at.

trust-store-trust-policy-specification.md
```diff
@@ -146,37 +145,70 @@ Trust policy for the scenario where ACME Rockets uses artifacts signed by themse
...
 
-**Trusted timestamp** : Guarantees that the signature was generated when the certificate was valid. It also allows a verifier to determine if a signature be treated as valid when a certificate is revoked, if the certificate was revoked after the signature was generated. In the absence of a trusted timestamp, signatures are considered invalid after certificate expires, and all signatures are considered revoked when a certificate is revoked. **NOTE**: `notation` RC1 will generate trusted timestamp using a TSA when the signature is generated, but will not support verification of TSA countersignatures.
+**Authentic timestamp** : Guarantees that the signature was generated when the certificate was valid. It also allows a verifier to determine if a signature be treated as valid when a certificate is revoked, if the certificate was revoked after the signature was generated. In the absence of a authentic timestamp, signatures are considered invalid after certificate expires, and all signatures are considered revoked when a certificate is revoked.
```
SteveLasker: > It also allows a verifier to determine if a signature be treated as valid when a certificate is revoked

It also allows a verifier to determine if a signature must be treated as valid when a certificate is revoked

trust-store-trust-policy-specification.md
```diff
@@ -146,37 +145,70 @@ Trust policy for the scenario where ACME Rockets uses artifacts signed by themse
...
+**Revocation check** : Guarantees that the signing identity is still trusted at signature verification time. Events such as key or system compromise can make a signing identity that was previously trusted, to be subsequently untrusted. This guarantee typically requires a verification-time call to an external system, which may not be consistently reliable. The `permissive` verification level only logs failures of revocation check and does not enforce it. If a particular revocation mechanism is reliable, use `strict` verification level instead.
+
+- **NOTE** : `notation` RC1 will not have in built support for Revocation Check feature. When this feature is subsequently implemented, the `strict` verification level will **enforce** this validation, and will fail signatures which would previously pass signature verification when revocation checks were not supported. This is considered a **breaking change**, but is the appropriate behavior for a `strict` verification level.
```
SteveLasker: > - **NOTE** : `notation` RC1 will not have in built support for Revocation Check feature

- **NOTE** : `notation` RC1 will not provide support for revocation checking. See #72 for details

SteveLasker (commented):
There's a lot of great content here, and lots of good design options.
That said, I'm a little concerned about the number of dials that could make notary v2 complex, where v2 was supposed to solve the complexity and usability of nv1.
Just a concerned for discussion and wondering if we can start simpler.

signature-envelope-jws.md
gokarnm: The details are in the linked section which points to generic signature specification.

signature-specification.md
gokarnm: This is the Notary v2 signature spec, the other doc is the JWS "implementation" of it. This doc provides context. 

>But, I imagine many internal tools may not wish to publish their toolset.

By internal tools do you mean the plugin?

trust-store-trust-policy-specification.md
gokarnm: We only make recommendation here for deploy time, the next section `permissive` calls out runtime. These are pretty broad suggestions, which is why we don't to preface them in bold. Users need to determine which level is appropriate for their use case, but we provide all the information here for them to decide.

trust-store-trust-policy-specification.md
gokarnm: I added one example in policy, we need not show multiple trusted stores for all examples as it would be less common.

trust-store-trust-policy-specification.md
gokarnm: Good call, updated.

gokarnm:
@SteveLasker addressed your feedback.

gokarnm:
> There's a lot of great content here, and lots of good design options. That said, I'm a little concerned about the number of dials that could make notary v2 complex, where v2 was supposed to solve the complexity and usability of nv1. Just a concerned for discussion and wondering if we can start simpler.

@SteveLasker can you call out specific elements that seem complex and if there are alternates to simplify.

trust-store-trust-policy-specification.md
gokarnm: Addressed.

trust-store-trust-policy-specification.md
gokarnm: Addressed.

gokarnm:
@SteveLasker Regarding complexity of trust policy, we believe this is the right set of controls to provide users. At a minimum they specify registryScope, trust store and trusted identity, and the verification level. We simplified the verification level quite a lot by having Notary defined presets.

Having CLI commands may make the experience simpler, but users still need to understand the concepts and implication of setting different values.

signature-specification.md
SteveLasker: Yeah, I'm referring to internal companies that may not want to publish their internal info. It is optional, so deferring.

SteveLasker (approved):
Approving, to move forward. I do think we'll need implementation, experience, user validation to get more confidence this is the final experience.

trust-store-trust-policy-specification.md
sajayantony: My recommendation would be for uniform objects even for usability it would avoid issues when one person changes it from audit to strict 

```json
"signatureVerification" : {
      "level" : "audit" 
}
```

if we change it to strict we won't get mutually exclusive properties.

gokarnm: Updated `signatureVerification` to an object type and not a multi-typed attribute.

gokarnm:
@sajayantony @SteveLasker updated `signatureVerification` to an object type and not a multi-typed attribute, as per last discussion. Also made examples clearer. Can you approve latest changes?

sajayantony (approved):
This covers the core scenarios and if at all we have any issues it most likely would only be uncovered once we start implementing this. 
I'm ok merging this and waiting for an implementation to iron out any edge cases.

gokarnm:
Resolved merge conflicts.

PR #165, opened 2022-07-06
Verification Plugin spec

gokarnm:
- Added verification plugin spec with new `verify-signature` plugin command.
- Updated Notary v2 Signature spec and JWS signature spec for new attributes to support Signing Scheme and Verification plugin


gokarnm:
cc: @shizhMSFT @roywill @priteshbandi @rgnote for review

signing-scheme.md
```diff
@@ -0,0 +1,73 @@
+# Signing Scheme
+
+Signatures are primarily used to provide a consumer of an artifact the following guarantees - integrity (the signed artifact was not tampered after signing it), and authenticity (the artifact was indeed signed by the entity who claims to have signed it). X.509 PKI based identities are commonly used to for signing artifacts. It has well established standards for issuing and representing identities (CAs and certificates) which sign artifacts, mechanism to establish authenticity using trust stores, and support for managing key lifetimes and rotation, and revocation of identities. There is interest in the community to support other ways to establish integrity and authenticity, based on other systems and techniques (Notary TUF, ledger based). These approaches also can provide different feature sets than ones traditionally provided by standard X.509 PKI based signing. E.g. TUF is geared for software update systems and provides a freshness guarantee.
```
letmaik: Assuming that the idea is to also support non-X.509 based identities, I think there are quite a few more places in the spec that would have to be changed. E.g., the certificate chain in the unprotected header is currently unconditionally marked as required, but some signing scheme wouldn't have that. Or the various places in `trust-store-trust-policy-specification.md` that talk about X.509. Should those be changed in this PR as well?

gokarnm: You are right, the PR does not update other areas like cert chain header or trust policy. The intent of Signing Scheme is to build the provision which will allow the signature schema and trust policy/trust store to be segregated per signing scheme where required. 
Current signing schemes only support X.509 identities, so certificate chain can specified as `required`. When other signing schemes that support non X.509 identities are introduced, we can define signature schema specific to them.
Similarly the trust policy references X.509 indirectly in two places, trust store and trusted identities. It's difficult to anticipate what type of identity representation and trust store mechanism a future signing scheme may use. So the policy is optimized for current known signing schemes and has a policy `version` attribute to handle breaking and non breaking changes when we introduce new policy attributes.

specs/plugin-extensibility.md
```diff
@@ -394,7 +411,155 @@ All response attributes are required.
...
+            "reason" : ""
+        }
+  },
+  // Array of strings containing critical attributes processed by the plugin.
+  "processedAttributes" : [ 
+  ]
+}
+```
+
+*verificationResults* : Array of verification results performed by the plugin. Each array element has following properties.
```
qmuntal: Here `verificationResults` is specified to be an array, but in the previous example it is a dictionary. Is it required?

qmuntal (changes requested):

signing-scheme.md
```diff
@@ -0,0 +1,73 @@
...
+
+*Q:* Why trust store for Signing Authority (`x509/signingAuthority`) is distinct from trust store for Certificate Authority (`x509/ca`), why can’t they share the same trust store?
+*A:* Signing Authority is a different type of trusted entity as compared to Certificate Authority (CA) or Timestamping Authority (TSA) which are much widely used. A CA is trusted for verifying the identity of a signing entity (end user) and issuing it a certificate, whereas a TSA is trusted to generate authentic timestamp. In contrast, an SA is trusted for generating signatures on behalf of a signing entity (end user) and also to generate authentic timestamp. If we use a shared trust store for CA and SA roots, an end user can self generate a signature with signing scheme `notary.signingAuthority.x509` and misrepresent themselves as a signing authority that would have ideally generated the signature. A verifying entity places certain trust in a signing authority - like the ability to generate authentic timestamp, and have uniform policies related to signature generation process, which aren't available when an end user generates a signature themselves.
```
priteshbandi: As discussed we should add some blurb around `separating trust store gives verifier control to dictate what they trust as CA and signing authority`.

specs/plugin-extensibility.md
```diff
@@ -394,7 +411,155 @@ All response attributes are required.
...
+#### Verification workflow using plugin
+
+1. If signature envelope contains *Verification Plugin* attribute, check if the plugin with the given name exists, else fail signature verification.
+    1. For the resolved plugin, call the `get-plugin-metadata` plugin command to get plugin version and capabilities.
+    2. If signature envelope contains *Verification plugin minimum version* attribute.
+        1. Validate that *Verification plugin minimum version* and plugin version are in SemVer format
+        2. Validate that plugin version is greater than or equal to *Verification plugin minimum version*
+        3. Fail signature verification if these validations fail
+    3. Validate if plugin capabilities contains any `SIGNATURE_VERIFIER` capabilities
+        1. Fail signature verification if a matching plugin with `SIGNATURE_VERIFIER` capability cannot be found. If signature envelope contains the *Verification Plugin Name* attribute, include it as a hint in error response.
```
priteshbandi: > If signature envelope contains the *Verification Plugin Name* attribute, include it as a hint in error response.

If plugin is invoked, doesnt that means signature envelope contains the *Verification Plugin Name* attribute?

specs/plugin-extensibility.md
gokarnm: Good catch, updated the description to map to match the response structure.

specs/plugin-extensibility.md
gokarnm: Good point, updated the language Verification Plugin Name attribute will be always present.

signing-scheme.md
gokarnm: Made some changes in the wording, let me know if you have further feedback.

qmuntal (approved):

specs/plugin-extensibility.md
```diff
@@ -394,7 +411,154 @@ All response attributes are required.
...
+    2. For each element in `response.verificationResults` map, process each result as follows
+        1. If `verificationResult.success` is true, proceed to next element.
+        2. Else if `verificationResult.success` is `false`, check if the failure should be enforced or logged based on value of `signatureVerification` level in Trust Policy. Proceed to next element if failure is logged, else fail signature verification with `verificationResult.reason`.
+    3. Validate values in `response.processedAttributes` match the set of values in `request.signature.unprocessedAttributes`. If they don't, fail signature verification, as the plugin did not process a critical attribute that is unknown to the caller.
+7. Perform any remaining steps in [signature verification workflow](../trust-store-trust-policy-specification.md#signature-verification) which are not covered by plugin's verification capabilities. These steps MUST process any remaining critical attributes. Fail signature verification if any critical attributes are unprocessed at the end of this step.
+
+#### *verify-signature* command
+
+*Response*
```
shizhMSFT: Request?

specs/plugin-extensibility.md
```diff
@@ -146,9 +154,18 @@ All response attributes are required.
...
   
-  // Currently one of 
-  // SIGNATURE_GENERATOR or
-  // SIGNATURE_ENVELOPE_GENERATOR
+  // List of one or more capabilities supported by plugin.
+  // Valid values are 
+  //    SIGNATURE_GENERATOR.RAW
+  //    SIGNATURE_GENERATOR.ENVELOPE
+  //    SIGNATURE_VERIFIER.TRUSTED_IDENTITY
+  //    SIGNATURE_VERIFIER.REVOCATION_CHECK
```
shizhMSFT, line 162: Will we support more `SIGNATURE_VERIFIER` capabilities? like `SIGNATURE_VERIFIER.RAW` and `SIGNATURE_VERIFIER.ENVELOPE`?

specs/plugin-extensibility.md
```diff
@@ -394,7 +411,154 @@ All response attributes are required.
...
+*processedAttributes* (required): Array of strings containing critical attributes processed by the plugin. Values must be one or more of attribute names in `verify-signature` request's `signature.unprocessedAttributes`.
+
+#### Error codes for *verify-signature*
+
+* VALIDATION_ERROR - Any of the required request fields was empty, or a value was malformed/invalid.
+* UNSUPPORTED_CONTRACT_VERSION - The contract version used in the request is unsupported.
+* ACCESS_DENIED - Authentication/authorization during signature verification, this may be due to external calls made by the plugin.
+* TIMEOUT - The verification process timed out and can be retried by Notation.
+* THROTTLED - The verification process was throttled and can be retried by Notation.
+* ERROR - Any general error that does not fall into previous error categories.
```
shizhMSFT, line 567: Which field should we put the error code?

specs/plugin-extensibility.md
gokarnm: The intent is to only allow specific verification steps in the overall workflow to be offloaded to plugin.  We can extend this interface to offload step 4 onwards in [verification workflow](https://github.com/notaryproject/notaryproject/blob/main/trust-store-trust-policy-specification.md#signature-verification) as required. We can also add steps like Trust Store resolver, Key resolver etc. as needed. 
I'm not in favor of adding a SIGNATURE_VERIFIER.ENVELOPE which will offload full verification unless we have a justified use case, we want notation to perform basic validation wherever possible.

specs/plugin-extensibility.md
gokarnm: Errors are returned as a separate error response JSON with non zero exit code - https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#plugin-contract

specs/plugin-extensibility.md
gokarnm: Fixed.

priteshbandi (approved):

signature-envelope-jws.md
```diff
@@ -70,10 +70,18 @@ Example
...
+- **`io.cncf.notary.expiry`**(*string*)(critical): This OPTIONAL header provides a “best by use” time for the artifact, as defined by the signer. Its value is a [RFC 3339][rfc3339] formatted date time, the optional fractional second ([time-secfrac][rfc3339][[1](https://datatracker.ietf.org/doc/html/rfc3339#section-5.3)]) SHOULD NOT be used.
+- **[`crit`](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.11)**(*array of strings*): This OPTIONAL header lists the headers that implementation MUST understand and process. It MUST only contain headers apart from registered headers (e.g. `alg`, `cty`) in JWS specification. This header MUST contain `io.cncf.notary.signingScheme` which is a required critical header, and optionally contain `io.cncf.notary.authenticSigningTime` and `io.cncf.notary.expiry` if these critical headers are present in the signature.
+
+## Extended Protected Headers for *Notation* Plugins
+
+See [Extended attributes for *Notation* Plugins](./signature-specification.md#extended-attributes-for-notation-plugins) for detailed description og these headers.
```
roywill: typo:

og => of

signing-scheme.md
```diff
@@ -0,0 +1,74 @@
...
+
+Notary v2 will initially support X.509 PKI identity based signing, but provide the flexibility for additional systems through an abstraction called Signing Scheme. The Signing Scheme covers aspects of signature generation and verification, and formalizes the feature set (guarantees) provided by the signature produced using a signing scheme. Generally it covers the following aspects, but can be extended to other aspects as required by newer systems.
+
+*Signature creation*
+
+1. Supported entities that can generate signatures - typically an end user generates the signature, other models can be supported using signature scheme.
+1. Representation of identities - e.g. X.509 certificates are used to represent end users and entities which verify the end user’s authenticity (CAs).
```
roywill: Do you need to bump these numbers?

signature-specification.md
```diff
@@ -225,6 +240,16 @@ The signing time denotes the time at which the signature was generated. A X509 c
 
 This is an optional feature that provides a “best by use” time for the artifact, as defined by the signer. Notary v2 allows users to include an optional expiry time when they generate a signature. The expiry time is not set by default and requires explicit configuration by users at the time of signature generation. The artifact is considered expired when the current time is greater than or equal to expiry time, users performing verification can either configure their trust policies to fail the verification or even accept the artifact with expiry date in the past using policy. This is an advanced feature that allows implementing controls for user defined semantics like deprecation for older artifacts, or block older artifacts in a production environment. Users should only include an expiry time in the signed artifact after considering the behavior they expect for consumers of the artifact after it expires. Users can choose to consume an artifact even after the expiry time based on their specific needs.
 
+### Guidelines for Notary v2 implementors
```
SteveLasker: nit, and fav converter: [Title Case conversion](https://titlecapitalize.com/)
Guidelines for Notary v2 Implementors

roywill (approved):

signature-envelope-jws.md
```diff
@@ -70,10 +70,18 @@ Example
...
-- **`io.cncf.notary.expiry`**(*string*): This OPTIONAL header provides a “best by use” time for the artifact, as defined by the signer. Its value is a [RFC 3339][rfc3339] formatted date time, the optional fractional second ([time-secfrac][rfc3339][[1](https://datatracker.ietf.org/doc/html/rfc3339#section-5.3)]) SHOULD NOT be used.
-- **[`crit`](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.11)**(*array of strings*): This OPTIONAL header lists the headers that implementation MUST understand and process. It MUST only contain headers apart from registered headers (e.g. `alg`, `cty`) in JWS specification, therefore this header is only present when the optional `io.cncf.notary.expiry` header is present in the protected headers collection.
-  If present, the value MUST be `["io.cncf.notary.expiry"]`.
+- **`io.cncf.notary.signingScheme`**(*string*)(critical): This REQUIRED header specifies the [Notary v2 Signing Scheme](./signing-scheme.md) used by the signature. Supported values are `notary.default.x509` and `notary.signingAuthority.x509`.
```
SteveLasker: Can you elaborate on why we have `notary.default.x509` and `notary.signingAuthority.x509`
what's the purpose of `default`?

signature-envelope-jws.md
```diff
@@ -70,10 +70,18 @@ Example
...
+- **[`crit`](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.11)**(*array of strings*): This OPTIONAL header lists the headers that implementation MUST understand and process. It MUST only contain headers apart from registered headers (e.g. `alg`, `cty`) in JWS specification. This header MUST contain `io.cncf.notary.signingScheme` which is a required critical header, and optionally contain `io.cncf.notary.authenticSigningTime` and `io.cncf.notary.expiry` if these critical headers are present in the signature.
+
+## Extended Protected Headers for *Notation* Plugins
+
+See [Extended attributes for *Notation* Plugins](./signature-specification.md#extended-attributes-for-notation-plugins) for detailed description og these headers.
+
+- **`io.cncf.notary.verificationPlugin`**(*string*)(critical): An OPTIONAL header that specifies the name of the verification plugin that MAY be used to verify the signature.
```
SteveLasker, line 108: Hmm, why do we think we need this?
The notary client should be able to verify any signature format supported. I haven't seen where we need plugins for verification.

signature-specification.md
```diff
@@ -126,9 +126,24 @@ Notary v2 requires the signature envelope to support the following signed attrib
...
+- **Authentic Signing Time** (critical): The authenticated time at which the signature was generated. This claim is REQUIRED and only valid when signing scheme is `notary.signingAuthority.x509` . More details [here](#signing-time).
 - **Expiry** (critical): An OPTIONAL claim that provides a “best by use” time for the artifact, as defined by the signer. More details [here](#expiry).
-- **Content type** (critical): A REQUIRED claim that indicates the content type of the [payload](#payload). The supported value is `application/vnd.cncf.notary.payload.v1+json`. Other payload types MAY be supported in future.
+- **Content Type** (critical): A REQUIRED claim that indicates the content type of the [payload](#payload). The supported value is `application/vnd.cncf.notary.payload.v1+json`. Other payload types MAY be supported in future.
+
+#### Extended attributes
+
+Implementations of Notary v2 signature spec MAY include additional signed attributes in the signature envelope. These attributes MAY be marked critical, i.e. the attribute MUST be understood and processed by a verifier, unknown critical attributes MUST cause signature verification to fail.
```
SteveLasker: This sounds like it would impose instability and inconsistency. 
A primary goal of Notary v2 is cloud/location independence. If a user copies an image from [MAR](mcr.microsoft.com) to their [ecr](https://aws.amazon.com/ecr/), they shouldn't need any special Microsoft stuff to validate it. 
I'm guessing you have some ideas around new headers. Can we discuss those and find a platform neutral way?

signature-specification.md
```diff
@@ -126,9 +126,24 @@ Notary v2 requires the signature envelope to support the following signed attrib
...
+
+#### Extended attributes for *Notation* Plugins
+
+This section documents extended attributes used by Notary v2 reference implementation *Notation* to support plugins. Plugins is a *Notation* concept that allows parts of signing and verification logic to be performed by an external provider. *Signing plugins* allow *Notation* to be extended for integration with remote keys remote key management services and signing services, where as *verification plugins* allow for customization of verification logic. Detailed specification for plugins can be found [here](https://github.com/notaryproject/notaryproject/blob/main/specs/plugin-extensibility.md#notation-extensibility-for-signing-and-verification). These extended attributes are documented in this spec, as other Notary V2 implementations may encounter these attributes if they verify a signature that indicated it required a verification plugin for complete signature verification.
```
SteveLasker: We should discuss this, due to the concerns above for location independence.

signature-specification.md
```diff
@@ -225,6 +240,16 @@ The signing time denotes the time at which the signature was generated. A X509 c
...
 
+### Guidelines for Notary v2 implementors
+
+Implementations of Notary v2, can choose to be [Notation plugin protocol](./specs/plugin-extensibility.md#plugin-contract) aware or not. If an implementation chooses to be plugin protocol aware, and it encounters the Verification Plugin and Verification Plugin minimum version attributes during signature verification, it MUST process these attributes. This involves finding the appropriate plugin and the version to use, and executing `verify-signature` plugin command with correct inputs and processing the plugin response, as per the [Verification Plugin interface](../specs/../notaryproject-specs/specs/plugin-extensibility.md#verification-extensibility).
```
SteveLasker, line 295: Red flag, noted above about verification plugins.

SteveLasker (changes requested):
I'm all for extensibility and flexibility, enabling various scenarios that don't require notary maintainers to be involved. The remote signing extensibility has proven to work really well, enabling various cloud providers to meet their product goals for how they support remote signing/key management solutions. This was a key for Notary v2 [scenarios](https://github.com/notaryproject/notaryproject/blob/main/scenarios.md) and [requirements](https://github.com/notaryproject/notaryproject/blob/main/requirements.md)

I have general concerns about a verification plugin that could limit the portability of signatures, as this could be counter to our usability and platform/vendor neutrality goals.

signing-scheme.md
gokarnm: This is correct, it's auto numbered in markdown. I fixed other occurrence.

signature-specification.md
sajayantony: Are these the critical headers we are discussing here - 
 - JSW - https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.11
 - COSE - https://datatracker.ietf.org/doc/html/rfc8152#appendix-C.1.4 

Given the statement - basically it means that a signature with a critical header will put a requirement on the client environment to have the plugin available.

gokarnm: Yes, the links point to how *criticality* of signed attributes is implemented in JWS and COSE. If an extended attribute is marked critical (not all extended attributes need to be critical) the verifier must process it either using a plugin or equivalent verification logic.

signature-envelope-jws.md
SteveLasker: @gokarnm r/ `.default`

signature-envelope-jws.md
```diff
@@ -70,10 +70,18 @@ Example
...
-  If present, the value MUST be `["io.cncf.notary.expiry"]`.
+- **`io.cncf.notary.signingScheme`**(*string*)(critical): This REQUIRED header specifies the [Notary v2 Signing Scheme](./signing-scheme.md) used by the signature. Supported values are `notary.default.x509` and `notary.signingAuthority.x509`.
+- **`io.cncf.notary.signingTime`**(*string*): This header specifies the time at which the signature was generated. This is an untrusted timestamp, and therefore not used in trust decisions. Its value is a [RFC 3339][rfc3339] formatted date time, the optional fractional second ([time-secfrac][rfc3339][[1](https://datatracker.ietf.org/doc/html/rfc3339#section-5.3)]) SHOULD NOT be used. This claim is REQUIRED and only valid when signing scheme is `notary.default.x509`.
```
SteveLasker: r/ `.default`

signature-specification.md
```diff
@@ -126,9 +126,24 @@ Notary v2 requires the signature envelope to support the following signed attrib
 
 #### Standard attributes
 
-- **Signing time**: A REQUIRED claim that indicates the time at which the signature was generated. Though this claim is signed by the signing key, it’s considered unauthenticated as a signer can modify local time and manipulate this claim. More details [here](#signing-time).
+- **Signing Scheme** (critical): A REQUIRED claim that defines the [Notary v2 Signing Scheme](./signing-scheme.md) used by the signature. This attribute dictates the rest of signature schema - the set of signed and unsigned attributes to be included in the signature. Supported values are `notary.default.x509` and `notary.signingAuthority.x509`.
+- **Signing Time**: A claim that indicates the time at which the signature was generated. Though this claim is signed by the signing key, it’s considered unauthenticated as a signer can modify local time and manipulate this claim. More details [here](#signing-time). This claim is REQUIRED and only valid when signing scheme is `notary.default.x509` .
```
SteveLasker: r/ `.default`

SteveLasker (approved):

gokarnm:
Removed Signing scheme related spec updates in this PR which is now a separate PR (#175).

signature-specification.md
```diff
@@ -248,6 +249,47 @@ The signing time denotes the time at which the signature was generated. A X509 c
...
+E.g. Images for containerized applications and services used within an organization, or shared with limited authorized parties. 
+These applications and services may be either intended to run in any infrastructure or may run on specific infrastructure e.g. integration with a cloud service provider. 
+Therefore, the signature’s portability should at least as much as the artifact’s portability.
+
+*Notary v2 signature portability* is based on the following
+
+**Signature discovery**
+
+Notary v2 addressed signature discovery by storing signatures in the same registry (location) where an artifact is present.
+This is currently supported through [ORAS artifact spec](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md) which allows reference artifacts such as signatures, SBOMs to be associated with existing artifacts like Images.
```
SteveLasker: suggest removing currently, as we can't predict future state. As things change, we can address at that time

signature-specification.md
```diff
@@ -248,6 +249,47 @@ The signing time denotes the time at which the signature was generated. A X509 c
...
+
+Portability of signatures is associated with the portability of associated artifacts which are being signed.
+OCI artifacts are inherently location agnostic, artifacts can be pulled from and pushed to any OCI compliant registry to which a user has access. 
+The artifacts themselves can be classified as follow.
+
+1. *Public Artifacts* -  Artifacts that are distributed publicly for broad consumption. 
+Artifacts distributed via public registries fall in this category. 
+E.g. Public images for software distributed by software vendors, and open source projects. 
+Signatures associated with these artifacts require broad portability.
+1. *Private Artifacts* - Artifacts that are private to a user or organization, and may be shared with limited parties.
```
SteveLasker: What is the intent when users/customers are multi-cloud/hybrid with on-prem?

signature-specification.md
```diff
@@ -248,6 +249,47 @@ The signing time denotes the time at which the signature was generated. A X509 c
...
+
+Notary v2 also supports signatures generated using compliant signing plugins, which allow vendors to optionally provide additional features on top of Notary v2 standard features.
+Verification of these signatures may require additional dependencies like Notary v2 compliant verification plugin, making these signatures more appropriate to use where broad portability may not be required for the associated signed artifact.
+This allows users to implement security controls required for their organizations, that are not broadly applicable and may take time to standardize.
+E.g. Integration with a signature transparency log as part of signature verification.
+
+Based on user’s requirements, a user can select appropriate signing mechanism that produces signatures with desired portability.
+Notation generated signatures using local keys and signing plugins of type *Raw Signature Generator* always produce signatures which can be verified without any additional dependencies.
```
SteveLasker: Does "local" here refer to local keys on the node doing the signing? Or, local to a cloud provider/isv where the actual private key is stored securely in an ISV keyvault?

signature-specification.md
gokarnm: Removed "currently".

signature-specification.md
gokarnm: I've simplified the language used in this section without referring all reasons an artifact may not be portable.

signature-specification.md
gokarnm: I've simplified the language used here without referring local keys or signing plugin type.

gokarnm:
Resolved merge conflicts.

SteveLasker:
Thanks @gokarnm for the latest updates.
My LGTM still applies, and believe we should merge with the understanding that we should have time before rc.1 to test any implementation and verify we're happy with the results. This is a baseline to iterate and learn.

NiazFK (approved):
LGTM

PR #167, opened 2022-07-07
Relax EKU requirements on signing certificates

### LINKED ISSUE #130: id-kp-codeSigning EKU restriction for signing certificate.
Discuss if we need to enforce id-kp-codeSigning EKU restriction for signing certificate.

----

Recommend removing this requirement, as not all signed artifacts will be "code".

_Originally posted by @briankr-ms in https://github.com/notaryproject/notaryproject/pull/122#discussion_r798210495_

ianjmcm:
Softening the requirement on the "signing" certificate (not "codesigning") for the signing of content in the registry to "MAY" contain the codeSign EKU ('id-kp-codeSigning') is the best way to allow for a wider variety of existing certificates and new EKUs to be defined for the signing of content types that are not "software" as defined in RFC 4949. Using code signing certificates (with the codeSign EKU) for signing data objects that are not software or directly related to the execution of software can lead to unintended trust by existing code integrity implementations that rightly require the codeSign EKU. As such, Private PKI owners may want to NOT issue code signing certificates to entities that are signing content in the registry that is not code (e.g. Container descriptors), not requiring the codeSigning EKU allows them to create the right controls on entities allowed to use codesigning certificates vs. entities that should/must not use codesigning certificates.

### DIFF
--- a/signature-specification.md
+++ b/signature-specification.md
@@ -214,7 +214,7 @@ The `basicConstraints` extension is OPTIONAL and can OPTIONALLY be marked as cri
 1. **[Key Usage:](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.3)**
 The `keyUsage` extension MUST be present and MUST be marked critical. Bit positions for `digitalSignature` MUST be set. The Bit positions for `keyCertSign` and `cRLSign` MUST NOT be set.
 1. **[Extended Key Usage:](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12)**  The `extendedKeyUsage` extension is OPTIONAL and can OPTIONALLY be marked as critical.
-    - **For codesigning certificate:** If present, the value MUST contain `id-kp-codeSigning` and MUST NOT contain `anyExtendedKeyUsage`, `serverAuth`, `emailProtection` and `timeStamping`.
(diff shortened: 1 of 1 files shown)


SteveLasker:
Thanks @ianjmcm 
The text above seems super valuable. Is there a means to put that into the PR, so readers of the specs will understand the reasoning behind the subtle change?

dtzar:
Thank you @ianjmcm.  Need to sign your commit so it passes [DCO check steps to resolve there](https://github.com/notaryproject/notaryproject/pull/167/checks?check_run_id=7239005886).

dtzar (approved):

dtzar:
@priteshbandi @iamsamirzon @gokarnm - this is ready for merge now passing DCO checks upon one of your approvals.

priteshbandi (approved):
LGTM.

NiazFK (approved):
LGTM

gokarnm:
@ianjmcm the change you are proposing allows Notation to support different types of signed artifacts in a registry, which may have certificates with different EKUs (for code, document, or supply chain metadata like SBOMs). The proposed solution relaxes the code signing EKU check to "MAY" which means it won't be enforced. This can still allow misusing EKU intended for one set of artifacts to be used for other. A longer term solution which will require some standardization may be to derive the enforced set of EKUs based on media type of the artifact being signed.

gokarnm (approved):
LGTM!

PR #185, opened 2022-08-15
feat(plugin)!: pass payload digest when `generate-signature`

shizhMSFT:
Changes to `SIGNATURE_GENERATOR.RAW` capability:

* `payloadDigest` instead of `payload` is passed in `generate-signature.request`.
  * Since only the digest of the payload is required, the plugin can fit [crypto.Signer](https://pkg.go.dev/crypto#Signer), and it simplifies the implementation of `notation-core-go`.
  * For the concern of the FIPS-compliance of hash implementation, it is not sufficient to delegate the hash computation to the plugin as `notation-go` still uses golang built-in hash implementation to do signature and certificate validation. For advanced scenarios, it should be considered compiling `notation` under FIPS-compliant golang or implementing a plugin with `SIGNATURE_GENERATOR.ENVELOPE` capability.

Signed-off-by: Shiwei Zhang <shizh@microsoft.com>

### DIFF
--- a/specs/plugin-extensibility.md
+++ b/specs/plugin-extensibility.md
@@ -250,7 +250,7 @@ This command is used to get metadata for a given key.
 }
 ```
 
-*keySpec* : One of following [supported key types](../signature-specification.md#algorithm-selection) - `RSA_2048`, `RSA_3072`, `RSA_4096`, `EC_256`, `EC_384`, `EC_512`.
+*keySpec* : One of following [supported key types](../signature-specification.md#algorithm-selection) - `RSA-2048`, `RSA-3072`, `RSA-4096`, `EC-256`, `EC-384`, `EC-521`.
 
 *certificateChain* : Ordered list of certificates starting with leaf certificate and ending with root certificate.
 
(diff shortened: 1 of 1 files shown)


shizhMSFT:
This PR is also for COSE support since the [Signer](https://pkg.go.dev/github.com/veraison/go-cose#Signer) of `go-cose` accepts digests only.

priteshbandi:
>     * Moved `response.certificateChain` from `generate-signature` to `describe-key`.
>       
>       * It is redundant to return the same certificate chain.
>       * Simplifies the implementation of `notation-core-go` where certificate chain can be obtained without generating any signature.

Moving certificateChain from `generate-signature` to `describe-key` opens up an edgecase during certification rotation, where `describe-key` returns old certificateChain and `generate-signature` signs with new certificateChain

shizhMSFT:
> Moving certificateChain from `generate-signature` to `describe-key` opens up an edgecase during certification rotation, where `describe-key` returns old certificateChain and `generate-signature` signs with new certificateChain

As discussed in the community call, I will remove this part out to another PR #187 for discussions.

shizhMSFT:
We can close this PR if https://github.com/veraison/go-cose/pull/101 is merged.

PR #187, opened 2022-08-16
feat(plugin)!: return cert chain in `describe-key`

shizhMSFT:
Changes to `SIGNATURE_GENERATOR.RAW` capability:

* Moved `response.certificateChain` from `generate-signature` to `describe-key`.
  * It is redundant to return the same certificate chain.
  * Simplifies the implementation of `notation-core-go` where certificate chain can be obtained without generating any signature.

Signed-off-by: Shiwei Zhang <shizh@microsoft.com>

### DIFF
--- a/specs/plugin-extensibility.md
+++ b/specs/plugin-extensibility.md
@@ -199,19 +199,19 @@ This interface targets plugins that integrate with providers of basic cryptograp
 5. Execute the plugin with `get-plugin-metadata` command
     1. If plugin supports capability `SIGNATURE_GENERATOR.RAW`
         1. Execute the plugin with `describe-key` command, set `request.keyId` and the optional `request.pluginConfig` to corresponding values associated with signing key `keyName` in `config.json`.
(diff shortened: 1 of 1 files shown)


shizhMSFT:
@priteshbandi The `keyId` defined in the spec is the key identifier, which uniquely identifies the signing key. Rotating the signing key should not re-use the key id. Otherwise, we will have 1-to-many relationship for unique key IDs and signing keys. It is worth noting that **key IDs** should points to unique keys but **key names** are not.

priteshbandi:
Yes, `keyId` uniquely identifies a signing key at any given point in time. When keys are rotated, usually the old key is archived/disabled and a new key is actived so at any given time there is only one active key.

The `keyId` from the uber perspective is an alias to a local/remote key which needs to be unique in notation config only, it's out of the scope of notation to dictate or restrict how local/remote keys pointed by keyIds are updated/rotated.

shizhMSFT:
Closing as some cloud providers may silently rotate the key without changing the key id.

PR #199, opened 2022-11-04
spec: update data type of signingTime to date/time

yizha1:
**What**

In order to make `signingTime` rfc compliant, update data type of `signingTime` to `date/time`, which is [RFC 8949](https://datatracker.ietf.org/doc/html/rfc8949#section-3.4.2) formatted date time.


signature-envelope-cose.md
```diff
@@ -90,9 +90,9 @@ Note: The above examples are represented using the [extended CBOR diagnostic not
...
-- **`io.cncf.notary.authenticSigningTime`** (*uint*, critical): This header specifies the authenticated time at which the signature was generated. Its value is the number of seconds from `1970-01-01T00:00Z` in UTC time, commonly known as UNIX timestamp. This claim is REQUIRED and only valid when signing scheme is `notary.x509.signingAuthority` .
-- **`io.cncf.notary.expiry`** (*uint*, critical): This OPTIONAL header provides a "best by use" time for the artifact, as defined by the signer. Its value is the number of seconds from `1970-01-01T00:00Z` in UTC time, commonly known as UNIX timestamp.
+- **`io.cncf.notary.signingTime`** (*date/time*): This header specifies the time at which the signature was generated. This is an untrusted date/time, and therefore not used in trust decisions. Its value is a [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6) formatted date time. This claim is REQUIRED and only valid when signing scheme is `notary.x509`.
```
shizhMSFT: JSON does not have a native datetime data type, and thus the JWS spec adopts the RFC 3339 date time format.

In contrast, COSE is based on CBOR where CBOR supports date time type natively ([RFC 8949 3.4.2](https://datatracker.ietf.org/doc/html/rfc8949#section-3.4.2)) with tag `1`. Although text string (tag `0`) is also available, it is better to use tag `1` for concise signature.

shizhMSFT (changes requested):

signature-envelope-cose.md
```diff
@@ -90,9 +90,9 @@ Note: The above examples are represented using the [extended CBOR diagnostic not
...
-- **`io.cncf.notary.authenticSigningTime`** (*uint*, critical): This header specifies the authenticated time at which the signature was generated. Its value is the number of seconds from `1970-01-01T00:00Z` in UTC time, commonly known as UNIX timestamp. This claim is REQUIRED and only valid when signing scheme is `notary.x509.signingAuthority` .
-- **`io.cncf.notary.expiry`** (*uint*, critical): This OPTIONAL header provides a "best by use" time for the artifact, as defined by the signer. Its value is the number of seconds from `1970-01-01T00:00Z` in UTC time, commonly known as UNIX timestamp.
+- **`io.cncf.notary.signingTime`** (*date/time*): This header specifies the time at which the signature was generated. This is an untrusted date/time, and therefore not used in trust decisions. Its value is an Epoch-Based Date/Time defined in [RFC 8949](https://datatracker.ietf.org/doc/html/rfc8949#section-3.4.2). This claim is REQUIRED and only valid when signing scheme is `notary.x509`.
```
letmaik, line 93: Is it tagged? Untagged?

shizhMSFT, line 93: Should we restrict the seconds to be non-fractional.

Reference: [RFC 8949 3.4.2](https://www.rfc-editor.org/rfc/rfc8949.html#section-3.4.2)
> An application that requires tag number 1 support may restrict the tag content to be an integer (or a floating-point value) only.

shizhMSFT: It should always be tagged with tag `1`.

shizhMSFT: I am not suggesting the restriction but open the discussion.

shizhMSFT (approved):
LGTM

signature-envelope-cose.md
toddysm: Can you elaborate on what is the benefit/drawback of fractional seconds?

dtzar (approved):

signature-envelope-cose.md
shizhMSFT: > Can you elaborate on what is the benefit/drawback of fractional seconds?

I'm not sure about this but I think most systems only recognize unix epoch in integers (i.e. non-fractional seconds). In the JWS based envelope spec, fractional seconds are marked as `SHOULD NOT`.

ianjmcm:
This looks good from my perspective and agree that fractional seconds must not be used.

SteveLasker (approved):
LGTM

ianjmcm:
LGTM

PR #211, opened 2022-11-23
Pass expiry to generate-envelope plugin

priteshbandi:
Pass expiry to envelope generator plugin so that it can be used for generating signature.
Issue: https://github.com/notaryproject/notation/issues/443

### DIFF
--- a/specs/plugin-extensibility.md
+++ b/specs/plugin-extensibility.md
@@ -375,6 +375,9 @@ All request attributes are required.
   // Optional plugin configuration, map of string-string
   "pluginConfig" : { },
 
+  // Optional signature expiry duration in seconds, number
+  "expiryDurationInSeconds" : 1234567,
+
   "payload" : "<Base64 encoded payload to be signed>",
   
   // The type of payload - currently a descriptor
@@ -388,6 +391,8 @@ All request attributes are required.
 *keyId* : Required field that has the key identifier (`keyId`) associated with signing key `keyName` in `config.json`.
 
 *pluginConfig* : Optional field for plugin configuration. For details, see [Plugin Configuration section](#plugin-configuration).
+
+*expiryDurationInSeconds* : Optional field which contains signature expiry duration(in seconds) as integer number.
 
 *signatureEnvelopeType* - defines the type of signature envelope expected from the plugin. As  Notation clients need to be updated in order to parse and verify new signature formats, the default signature format can only be changed with new major version releases of Notation. Users however can opt into using an updated signature envelope format supported by Notation, by passing an optional parameter.
 e.g. `notation sign $IMAGE --key {key-name} --envelope-type {some-new-type}`


jondonas (approved):

gokarnm (approved):

gokarnm (approved):
LGTM!

specs/plugin-extensibility.md
```diff
@@ -375,6 +375,9 @@ All request attributes are required.
   // Optional plugin configuration, map of string-string
   "pluginConfig" : { },
 
+  // Optional signature expiry duration in seconds, number
+  "expiryDurationInSeconds" : 1234567,
```
shizhMSFT, line 379: Taking a look at the current [notation.SignOptions](https://pkg.go.dev/github.com/notaryproject/notation-go@main#SignOptions), the `expriy` is a `time.Time` and not a `time.Duration`. Therefore, this field should also be a time field.

shizhMSFT, line 379: In addition, the `ArtifactReference` is not passed to the plugin. How would the trust policy in the plugin know which policy to apply?

shizhMSFT (changes requested):

specs/plugin-extensibility.md
priteshbandi: In notation cli user pass expiry as duration `24h5m`, so `notaiton.SignOptions`  should use `time.Duration` instead of `time.Time` because actual expiry should be calculated as close as possible to sign operation.

Scenario: user called `notation sign $Image -e 1m` with plugin and if the plugin execute took 30 secs, the user got the expiry only of 30 sec not 1 minute.

priteshbandi: This is for sign operation not verify.

shizhMSFT: Oh. I mixed the paths.

shizhMSFT: This is exactly I am worried about. The expiry should be deterministic and is not based on the signature generation time and influenced by any possible time skew.

For example, I want to sign an artifact for deployment and let it expire immediately after Thanksgiving 12:00AM sharp time. Therefore, I calculate the duration based on my local time and run `notation sign`. The signing process takes 30s and due a time skew between my machine and the server, the signature will actually be expired on the Thanksgiving Day 11:55 PM.

Therefore, expiry should be an exact date time for all internal component communication. The duration-typed expiry at the notation CLI side is for user experience, which can be reviewed later.

yizha1: Maybe we should remove `expiry` from CLI for now,  if we didn't think through the behavior. Once it is released, it's a breaking change if we want to change the behaviors.

shizhMSFT (approved):
LGTM as discussed in the community meeting that expiry in the signing options should be duration all of the places.

PR #217, opened 2023-01-06
doc: notation supports OCI Image manifest

yizha1:
Update:
- Add a section "Backwards Compatibility" on notation support of using OCI image manifest to store signatures
- Add an example of signature manifest per OCI image manifest
- Add manifest properties requirements accordingly

### DIFF
--- a/specs/signature-specification.md
+++ b/specs/signature-specification.md
@@ -90,8 +90,8 @@ Besides the [image manifest property requirements][image-manifest-property-descr
 - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.image.manifest.v1+json`.
 - **`config`** (*descriptor*): This property is REQUIRED to be compatible with [OCI image specification][oci-image-manifest]. Notary v2 doesn't require any configurations for a signature, and the configuration contents are not consumed by Notary v2.
   - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.cncf.notary.signature`.
-  - **`digest`** (*string*): This REQUIRED property is the digest of the targeted content.
-  - **`size`** (*int64*): This REQUIRED property specifies the size, in bytes, of the raw content.
+  - **`digest`** (*string*): This REQUIRED property is the digest of the config content.
+  - **`size`** (*int64*): This REQUIRED property specifies the size, in bytes, of the raw config content.
 - **`layers`** (*array of objects*): This REQUIRED property contains collection of only one [OCI descriptor][oci-descriptor] referencing signature envelope.
   - **`mediaType`** (*string*): This REQUIRED property contains media type of signature envelope blob. Following values are supported
     - `application/jose+json`
--- a/specs/signature-specification.md
+++ b/specs/signature-specification.md
@@ -14,7 +14,7 @@ The signature manifest has an artifact type that specifies it's a Notary V2 sign
 
 ![Signature storage inside registry](../media/signature-specification.svg)
 
-Signature Manifest Example per OCI artifact manifest
+Signature manifest example per OCI artifact manifest
 
 ```jsonc
 {
@@ -55,14 +55,14 @@ Signature Manifest Example per OCI artifact manifest
 
 Notary v2 MAY support using [OCI Image manifest][oci-image-manifest] to store the signature in the registries that implement partial or older versions of the OCI Image specification.
 
-Signature Manifest example per OCI Image manifest:
+Signature manifest example per OCI image manifest:
 
 ```jsonc
 {
     "schemaVersion": 2,
     "mediaType": "application/vnd.oci.image.manifest.v1+json",
     "config": {
-        "mediaType": "application/vnd.oci.image.config.v1+json",
+        "mediaType": "application/vnd.cncf.notary.signature",
         "size": 2,
         "digest": "sha256:ca3d163bab055381827226140568f3bef7eaac187cebd76878e0b63e9e442356"
     },
@@ -88,7 +88,10 @@ Signature Manifest example per OCI Image manifest:
 Besides the [image manifest property requirements][image-manifest-property-descriptions], the properties has the following additional restrictions:
 
 - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.image.manifest.v1+json`.
-- **`config`** (*descriptor*): This REQUIRED property references a configuration object for a signature by digest.
+- **`config`** (*descriptor*): This REQUIRED property references a configuration object for a signature by digest. Notary v2 doesn't require specific configurations for a signature. Implementations SHOULD decide the configuration content based on their specific needs.
+  - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.cncf.notary.signature`.
+  - **`digest`** (*string*): This REQUIRED property is the digest of the targeted content.
+  - **`size`** (*int64*): This REQUIRED property specifies the size, in bytes, of the raw content.
 - **`layers`** (*array of objects*): This REQUIRED property contains collection of only one [OCI descriptor][oci-descriptor] referencing signature envelope.
   - **`mediaType`** (*string*): This REQUIRED property contains media type of signature envelope blob. Following values are supported
     - `application/jose+json`
@@ -357,7 +360,7 @@ Alternatively, an implementation of Notary v2 can choose not to implement plugin
 [annotation-rules]: https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules
 [oci-descriptor]: https://github.com/opencontainers/image-spec/blob/main/descriptor.md
 [ietf-rfc3161]: https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2
-[oci-artifact-manifest]: https://github.com/opencontainers/image-spec/blob/main/artifact.md
-[oci-distribution-referrers]: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers
-[oci-image-manifest]: https://github.com/opencontainers/image-spec/blob/main/manifest.md
-[image-manifest-property-descriptions]: https://github.com/opencontainers/image-spec/blob/main/manifest.md#image-manifest-property-descriptions
+[oci-artifact-manifest]: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/artifact.md
+[oci-distribution-referrers]: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers
+[oci-image-manifest]: https://github.com/opencontainers/image-spec/blob/v1.0.2/manifest.md
+[image-manifest-property-descriptions]: https://github.com/opencontainers/image-spec/blob/v1.0.2/manifest.md#image-manifest-property-descriptions
--- a/specs/signature-specification.md
+++ b/specs/signature-specification.md
@@ -8,7 +8,7 @@ This document provides the following details for Notary v2 signatures:
 ## Storage
 
 This section describes how Notary v2 signatures are stored in the OCI Distribution conformant registry.
-Notary v2 uses [OCI artifact manifest][oci-artifact-manifest] to store the signature in the registry.
+Notary v2 uses [OCI artifact manifest][oci-artifact-manifest] to store signatures in the registry.
 The media type of the signature manifest is `application/vnd.oci.artifact.manifest.v1+json`.
 The signature manifest has an artifact type that specifies it's a Notary V2 signature, a reference to the manifest of the artifact being signed, a blob referencing the signature, and a collection of annotations.
 
@@ -53,7 +53,7 @@ Signature manifest example per OCI artifact manifest
 
 ### Backward Compatibility
 
-Notary v2 MAY support using `OCI image manifest` to store the signature in the registries that implement partial of the [OCI image specification v1.1][oci-image-manifest].
+Notary v2 MAY support using `OCI image manifest` to store signatures in the registries that implement partial of the [OCI image specification v1.1][oci-image-manifest].
 
 Signature manifest example per OCI image manifest:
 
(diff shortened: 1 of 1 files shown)


specs/signature-specification.md
```diff
@@ -51,6 +51,55 @@ Signature Manifest Example per OCI artifact manifest
   - **`io.cncf.notary.x509chain.thumbprint#S256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprint of signing certificate and certificate chain (including root) used for signature generation. The annotation name contains the hash algorithm as a suffix (`#S256`) and can be extended to support other hashing algorithms in future.
     The list of fingerprints is present as a JSON array string, corresponding to ordered certificates in [*Certificate Chain* unsigned attribute](#unsigned-attributes) in the signature envelope.
 
+### Backwards Compatibility
+
+Notary v2 MAY support using [OCI Image manifest][oci-image-manifest] to store the signature in the registries that implement partial or older versions of the OCI Image specification.
+
+Signature Manifest example per OCI Image manifest:
```
shizhMSFT: nit:
```suggestion
Signature manifest example per OCI image manifest:
```

specs/signature-specification.md
```diff
@@ -51,6 +51,55 @@ Signature Manifest Example per OCI artifact manifest
...
+
+```jsonc
+{
+    "schemaVersion": 2,
+    "mediaType": "application/vnd.oci.image.manifest.v1+json",
+    "config": {
+        "mediaType": "application/vnd.oci.image.config.v1+json",
```
shizhMSFT: The media type of the config layer is used as the artifact type. Therefore,
```suggestion
        "mediaType": "application/vnd.cncf.notary.signature",
```

specs/signature-specification.md
```diff
@@ -310,3 +359,5 @@ Alternatively, an implementation of Notary v2 can choose not to implement plugin
 [ietf-rfc3161]: https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2
 [oci-artifact-manifest]: https://github.com/opencontainers/image-spec/blob/main/artifact.md
 [oci-distribution-referrers]: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers
+[oci-image-manifest]: https://github.com/opencontainers/image-spec/blob/main/manifest.md
+[image-manifest-property-descriptions]: https://github.com/opencontainers/image-spec/blob/main/manifest.md#image-manifest-property-descriptions
```
shizhMSFT: Specific version of the spec should be referenced instead of `main`.

specs/signature-specification.md
```diff
@@ -51,6 +51,55 @@ Signature Manifest Example per OCI artifact manifest
...
+        "io.cncf.notary.x509chain.thumbprint#S256": 
+        "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\",\"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
+    }
+}
+```
+
+Besides the [image manifest property requirements][image-manifest-property-descriptions], the properties has the following additional restrictions:
+
+- **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.image.manifest.v1+json`.
+- **`config`** (*descriptor*): This REQUIRED property references a configuration object for a signature by digest.
```
shizhMSFT: The content of the `config` is not specified.

specs/signature-specification.md
```diff
@@ -51,6 +51,55 @@ Signature Manifest Example per OCI artifact manifest
   - **`io.cncf.notary.x509chain.thumbprint#S256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprint of signing certificate and certificate chain (including root) used for signature generation. The annotation name contains the hash algorithm as a suffix (`#S256`) and can be extended to support other hashing algorithms in future.
     The list of fingerprints is present as a JSON array string, corresponding to ordered certificates in [*Certificate Chain* unsigned attribute](#unsigned-attributes) in the signature envelope.
 
+### Backwards Compatibility
+
+Notary v2 MAY support using [OCI Image manifest][oci-image-manifest] to store the signature in the registries that implement partial or older versions of the OCI Image specification.
```
shizhMSFT: This statement is not quite accurate. The `subject` field for the OCI image manifest is introduced in `image-spec v1.1`. Therefore, registries supporting OCI image-spec v1.0 may still not support OCI image manifest with the `subject` field.

vaninrao10: > This statement is not quite accurate. The `subject` field for the OCI image manifest is introduced in `image-spec v1.1`. Therefore, registries supporting OCI image-spec v1.0 may still not support OCI image manifest with the `subject` field.

when pushing the fallback manifest, oras library handles the complexities of creating the relationship between artifact and subject via fallback mechanism.

sajayantony: @shizhMSFT - I think if registries don't support additional properties like the subject, I think its fair to say that the it doesn't support extended fields for even Image-spec 1.0 and maybe we should not handle that. For e.g. example the registry might not even support index. From the tool perspective do you think having the subject property in v1.0 is ok?

shizhMSFT: From the tool perspective, having the `subject` property in `v1.0` is definitely ok since it is indeed partial `v1.1`.

My focus is the accuracy of the wording in the spec documentation. That is, implementing `image-spec v1.0` does not imply supporting the `subject` field.

How about
```suggestion
Notary v2 MAY support using [OCI image manifest][oci-image-manifest] to store the signature in the registries that implement partial OCI Image specification v1.1.
```

yizha1: @shizhMSFT How about I update the link of `oci-image-manifest` to `OCI image manifest v1.1.0-rc.2` to indicate that notation using OCI image manifest with `subject` field, while keep the wording `implement partial or older versions of the OCI Image specification. ` Because with your suggestion, it seems registries implementing v1.0 are excluded. I think some registries implementing v1.0 support `subject` field, since it is an optional field.

shizhMSFT: > some registries implementing v1.0 support `subject` field

Yes, you may consider those registries partially v1.1 compatible.

specs/signature-specification.md
```diff
@@ -51,6 +51,58 @@ Signature Manifest Example per OCI artifact manifest
...
+        "io.cncf.notary.x509chain.thumbprint#S256": 
+        "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\",\"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
+    }
+}
+```
+
+Besides the [image manifest property requirements][image-manifest-property-descriptions], the properties has the following additional restrictions:
+
+- **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.image.manifest.v1+json`.
+- **`config`** (*descriptor*): This REQUIRED property references a configuration object for a signature by digest. Notary v2 doesn't require specific configurations for a signature. Implementations SHOULD decide the configuration content based on their specific needs.
```
shizhMSFT: Should we call out that the content of `config` is never used by notary v2 as it exists just because of compatibility?

shizhMSFT: Note: There is no config in the original OCI Artifact Manifest.

shizhMSFT: If we store meaningful config in the OCI image manifest, then we are not able to convert OCI image manifest to OCI artifact manifest and vice versa in the future.

shizhMSFT (approved):
LGTM

specs/signature-specification.md
```diff
@@ -308,5 +360,7 @@ Alternatively, an implementation of Notary v2 can choose not to implement plugin
 [annotation-rules]: https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules
 [oci-descriptor]: https://github.com/opencontainers/image-spec/blob/main/descriptor.md
 [ietf-rfc3161]: https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2
```
patrickzheng200, line 362: Somehow, I cannot see these links in the webpage. I think they are not successfully rendered in the markdown? hence:
```suggestion
## References
annotation-rules: https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules</br>
oci-descriptor: https://github.com/opencontainers/image-spec/blob/main/descriptor.md</br>
ietf-rfc3161: https://datatracker.ietf.org/doc/html/rfc3161#section-2.4.2</br>
```

specs/signature-specification.md
```diff
@@ -308,5 +360,7 @@ Alternatively, an implementation of Notary v2 can choose not to implement plugin
...
-[oci-artifact-manifest]: https://github.com/opencontainers/image-spec/blob/main/artifact.md
-[oci-distribution-referrers]: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers
+[oci-artifact-manifest]: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/artifact.md
+[oci-distribution-referrers]: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers
+[oci-image-manifest]: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md
+[image-manifest-property-descriptions]: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md#image-manifest-property-descriptions
```
patrickzheng200, line 366: Same here:
```suggestion
oci-artifact-manifest: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/artifact.md</br>
oci-distribution-referrers: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers</br>
oci-image-manifest: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md</br>
image-manifest-property-descriptions: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md#image-manifest-property-descriptions</br>
```

patrickzheng200 (changes requested):

specs/signature-specification.md
yizha1: This is the correct usage for replacing the reference with the links for a markdown file. So I will keep it unchanged.

specs/signature-specification.md
yizha1: This is the correct usage for replacing the reference with the links for a markdown file. So I will keep it  unchanged.

patrickzheng200 (approved):
LGTM

specs/signature-specification.md
```diff
@@ -51,6 +51,58 @@ Signature Manifest Example per OCI artifact manifest
...
+        "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\",\"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
+    }
+}
+```
+
+Besides the [image manifest property requirements][image-manifest-property-descriptions], the properties has the following additional restrictions:
+
+- **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.image.manifest.v1+json`.
+- **`config`** (*descriptor*): This property is REQUIRED to be compatible with [OCI image specification][oci-image-manifest]. Notary v2 doesn't require any configurations for a signature, and the configuration contents are not consumed by Notary v2.
+  - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.cncf.notary.signature`.
```
priteshbandi, line 92: If the config is not consumed by notary then are we enforcing mediaType to be  `application/vnd.cncf.notary.signature` just for sanity check? Should notary implementation check the value?

specs/signature-specification.md
```diff
@@ -51,6 +51,58 @@ Signature Manifest Example per OCI artifact manifest
...
+  - **`digest`** (*string*): This REQUIRED property is the digest of the config content.
+  - **`size`** (*int64*): This REQUIRED property specifies the size, in bytes, of the raw config content.
```
priteshbandi, line 94: What should be values of these two fields?

yizha1: This is the spec. The values leave to the decision of different implementations.

specs/signature-specification.md
yizha1: This is defined by the spec here 
https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pushing-manifests-with-subject

and referrer API also requires https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers

specs/signature-specification.md
```diff
@@ -8,13 +8,13 @@ This document provides the following details for Notary v2 signatures:
 ## Storage
 
 This section describes how Notary v2 signatures are stored in the OCI Distribution conformant registry.
-Notary v2 uses [OCI artifact manifest][oci-artifact-manifest] to store the signature in the repository.
+Notary v2 uses [OCI artifact manifest][oci-artifact-manifest] to store the signature in the registry.
```
patrickzheng200:
```suggestion
Notary v2 uses [OCI artifact manifest][oci-artifact-manifest] to store signatures in the registry.
```

specs/signature-specification.md
```diff
@@ -51,6 +51,58 @@ Signature Manifest Example per OCI artifact manifest
   - **`io.cncf.notary.x509chain.thumbprint#S256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprint of signing certificate and certificate chain (including root) used for signature generation. The annotation name contains the hash algorithm as a suffix (`#S256`) and can be extended to support other hashing algorithms in future.
     The list of fingerprints is present as a JSON array string, corresponding to ordered certificates in [*Certificate Chain* unsigned attribute](#unsigned-attributes) in the signature envelope.
 
+### Backward Compatibility
+
+Notary v2 MAY support using `OCI image manifest` to store the signature in the registries that implement partial of the [OCI image specification v1.1][oci-image-manifest].
```
patrickzheng200:
```suggestion
Notary v2 MAY support using `OCI image manifest` to store signatures in registries that implement partial of the [OCI image specification v1.1][oci-image-manifest].
```

specs/signature-specification.md
```diff
@@ -51,6 +51,58 @@ Signature Manifest Example per OCI artifact manifest
...
+        "size": 16724
+    },
+    "annotations": {
+        "io.cncf.notary.x509chain.thumbprint#S256": 
+        "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\",\"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
+    }
+}
+```
+
+Besides the [image manifest property requirements][image-manifest-property-descriptions], the properties has the following additional restrictions:
```
patrickzheng200:
```suggestion
Besides the [image manifest property requirements][image-manifest-property-descriptions], the property has the following additional restrictions:
```

specs/signature-specification.md
```diff
@@ -51,6 +51,58 @@ Signature Manifest Example per OCI artifact manifest
...
+
+- **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.image.manifest.v1+json`.
+- **`config`** (*descriptor*): This property is REQUIRED to be compatible with [OCI image specification][oci-image-manifest]. Notary v2 doesn't require any configurations for a signature, and the configuration contents are not consumed by Notary v2.
```
patrickzheng200:
```suggestion
- **`config`** (*descriptor*): This property is REQUIRED to be compatible with [OCI image specification][oci-image-manifest]. Notary v2 doesn't require any configuration for a signature, and the configuration content is not consumed by Notary v2.
```

specs/signature-specification.md
```diff
@@ -51,6 +51,58 @@ Signature Manifest Example per OCI artifact manifest
...
+  - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.cncf.notary.signature`.
+  - **`digest`** (*string*): This REQUIRED property is the digest of the config content.
+  - **`size`** (*int64*): This REQUIRED property specifies the size, in bytes, of the raw config content.
+- **`layers`** (*array of objects*): This REQUIRED property contains collection of only one [OCI descriptor][oci-descriptor] referencing signature envelope.
```
patrickzheng200:
```suggestion
- **`layers`** (*array of objects*): This REQUIRED property contains collection of only one [OCI descriptor][oci-descriptor] referencing the signature envelope.
```

priteshbandi (approved):
LGTM

specs/signature-specification.md
yizha1: I would like to keep the plural and changed the word `has` to `have`.

shizhMSFT (approved):
LGTM

PR #249, opened 2023-03-27
fix: Update revocation spec

priteshbandi:
Some updates to revocation specs. Also, removed `remotes/upstream/main/ADA-fuzzing-audit-22-23.pdf` file.

*This PR only updates revocation specs and doesn't addresses renaming issue. Renaming is covered by @yizha1 in separate PR https://github.com/notaryproject/notaryproject/pull/262*

References:
1. https://bugzilla.mozilla.org/show_bug.cgi?id=745747
2. https://www.ibm.com/docs/en/ibm-mq/8.0?topic=certificates-revoked-ocsp
3. https://groups.google.com/g/mozilla.dev.security.policy/c/CLg_MOZoG-M
4. https://docs.oracle.com/middleware/12213/wls/SECMG/ssl_cert_revoc_check.htm#SECMG579
5. https://www.cisco.com/c/en/us/td/docs/security/ise/2-6/pic_admin_guide/PIC_admin26/PIC_admin26_chapter_0101.html
6. https://www.sysadmins.lv/retired-msft-blogs/xdot509/pki-design-considerations-certificate-revocation-and-crl-publishing-strategies.aspx
7. https://www.securew2.com/blog/certificate-revocation-crl-explained
8. https://blog.cloudflare.com/high-reliability-ocsp-stapling/
9. https://uptime.netcraft.com/perf/reports/performance/CRL?reverse=1&orderby=avg_total 
10. https://uptime.netcraft.com/perf/reports/performance/OCSP?reverse=1&orderby=avg_total


specs/trust-store-trust-policy.md
```diff
@@ -285,7 +285,6 @@ The RDN consists of an attribute type name followed by an equal sign and the str
 
 Notary v2 allows user to execute custom validations during verificaton using plugins. Please refer [plugin-extensibility.md](plugin-extensibility.md#verification-extensibility) for more information.
```
yizha1, line 286: Notary v2 ==> Notation

specs/trust-store-trust-policy.md
```diff
@@ -379,8 +376,8 @@ Implementations MAY add support for caching CRLs and OCSP response to improve av
 CRL download location (URL) can be obtained from the certificate's CRL Distribution Point (CDP) extension.
 If the certificate contains multiple CDP locations then each location download is attempted in sequential order.
 For each CDP location, Notary V2 will try to download the CRL for the default threshold of 10 seconds.
```
yizha1: Notary v2 => Notation

specs/trust-store-trust-policy.md
```diff
@@ -415,9 +412,10 @@ When delta CRLs are implemented, the following results can occur during revocati
 ##### OCSP Download
 
 OCSP URLs can be obtained from the certificate's authority information access (AIA) extension as defined in [RFC-2560](https://datatracker.ietf.org/doc/html/rfc2560).
+If the certificate contains multiple OCSP URLs, then each URL is invoked in sequential order.
 Notary V2 will wait for a default threshold of 5 seconds to receive an OCSP response.
```
yizha1: Notary v2 ==> Notation

specs/trust-store-trust-policy.md
```diff
@@ -427,11 +425,12 @@ To check the revocation status of a certificate using OCSP, the following steps
...
-   A CRL is considered expired if the current date is after the `NextUpdate` field in the CRL.
+   Response is considered expired if the current date is after the `NextUpdate` field in the response.
 1. Verify that the OCSP response indicates that the certificate is not revoked i.e `CertStatus` is `good`.
     1. If the certificate is revoked i.e `CertStatus` is `revoked`, look for `InvalidityDate`.
        If the invalidity date is present and timestamp signature is also present then if the invalidity date is before the timestamping date, the certificate is considered revoked.
        If the invalidity date is not present in OCSP response, the certificate is considered revoked.
+    1. If the certificate status is `unknown`, the revocation status is considered `revocation unavailable`.
 1. If `id-pkix-ocsp-nocheck`(1.3.6.1.5.5.7.48.1.5) extension is not present on the OCSP signing certificate then revocation checking must be performed using CRLs for the OCSP signing certificate.
 
 ## FAQ
```
yizha1, line 436: Please update `Notary v2` to `Notation` for other places in the document.

yizha1, line 436: Maybe we can remove the following "TODO"?

**Q: Does Notary v2 support overriding of revocation endpoints to support signature verification in disconnected environments?**
**A:** TODO: Update after verification extensibility spec is ready.

yizha1 (commented):
Suggest removing Notary v2 references in this document.

shizhMSFT (dismissed):
LGTM

specs/trust-store-trust-policy.md
toddysm: Second all commends from @yizha1 about changing the name. When we talk about the signature, we should use "Notary signature", when we talk about the CLI, we should use "Notation".

toddysm (changes requested):
I will need to go over the spec in more detail but one question that immediately pops up is how this revocation will work in air-gapped environments where the OCSP and CDP may not be accessible.

specs/trust-store-trust-policy.md
priteshbandi: Yes we should move away from `notary v2` but we shouldn't include unrelated changes as part of updating revocation spec PR. There should be a separate pr to remove `notary v2` usage.

specs/trust-store-trust-policy.md
priteshbandi: Yes will do it in separate PR

vaninrao10 (dismissed):
LGTM

specs/trust-store-trust-policy.md
yizha1: @priteshbandi I am OK to create another PR to address the `notary v2` wording.

priteshbandi:
> I will need to go over the spec in more detail but one question that immediately pops up is how this revocation will work in air-gapped environments where the OCSP and CDP may not be accessible.

Not inherently but user can use plugin verification extensibility.

yizha1:
@toddysm Maybe we can discuss further on air-gapped environment requirement. My understanding is that in the air-gapped environment, there could still be PKI infrastructure established, right?

specs/trust-store-trust-policy.md
```diff
@@ -415,9 +412,10 @@ When delta CRLs are implemented, the following results can occur during revocati
 ##### OCSP Download
 
 OCSP URLs can be obtained from the certificate's authority information access (AIA) extension as defined in [RFC-2560](https://datatracker.ietf.org/doc/html/rfc2560).
```
patrickzheng200: Should we update the RFC to 6960? It's stated that rfc-6960 obsoletes rfc-2560.
```suggestion
OCSP URLs can be obtained from the certificate's authority information access (AIA) extension as defined in [RFC-6960](https://datatracker.ietf.org/doc/html/rfc6960).
```

yizha1:
@priteshbandi have you considered this https://en.wikipedia.org/wiki/OCSP_stapling

priteshbandi:
> @priteshbandi have you considered this https://en.wikipedia.org/wiki/OCSP_stapling

Yes but its not directly applicable to our usecase because stapling is done by content provider in our case it would be registry.

yizha1:
@priteshbandi overall is good, could elaborate a bit about the last comment, how user can use plugin verification extensibility?

> I will need to go over the spec in more detail but one question that immediately pops up is how this revocation will work in air-gapped environments where the OCSP and CDP may not be accessible.

>Not inherently but user can use plugin verification extensibility.

shizhMSFT (dismissed):
LGTM

specs/trust-store-trust-policy.md
```diff
@@ -414,10 +411,11 @@ When delta CRLs are implemented, the following results can occur during revocati
 
 ##### OCSP Download
 
-OCSP URLs can be obtained from the certificate's authority information access (AIA) extension as defined in [RFC-2560](https://datatracker.ietf.org/doc/html/rfc2560).
+OCSP URLs can be obtained from the certificate's authority information access (AIA) extension as defined in [RFC-6960](https://datatracker.ietf.org/doc/html/rfc6960).
```
shizhMSFT: nit:
```suggestion
OCSP URLs can be obtained from the certificate's authority information access (AIA) extension as defined in [RFC 6960](https://www.rfc-editor.org/rfc/rfc6960).
```

specs/trust-store-trust-policy.md
```diff
@@ -345,11 +344,9 @@ Notary v2 allows user to execute custom validations during verificaton using plu
 
 ### Certificate Revocation Evaluation
 
-This section should be considered as **DRAFT** and will be updated after RC1 release.
-
 If the certificate revocation validation is set to `skip` using policy, skip the below steps, otherwise check for revocation status for certificate and certificate chain.
 
-- If the revocation status of any of the certificates is revoked or cannot be determined (revocation status unavailable), fail this step. This will fail the overall signature verification or log the error, depending on if revocations status check is `enforced` or `logged`, as per the signature verification level.
+If the revocation status of any of the certificates is revoked or cannot be determined (revocation unavailable), fail this step. This will fail the overall signature verification or log the error, depending on if revocations status check is `enforced` or `logged`, as per the signature verification level.
```
toddysm, line 349: The first sentence is a bit confusing. Can we provide some more clarity? At the moment is reads:
> If the revocation status of any of the certificates is revoked or cannot be determined (revocation unavailable)
If the _revocation status is revoked_ doesn't this mean that the certificate is valid? I would suggest to reword to make sure we are clear what the behavior is.

specs/trust-store-trust-policy.md
```diff
@@ -395,7 +392,7 @@ To check the revocation status of a certificate against CRL, the following steps
        If the invalidity date is before the timestamping date, the certificate is considered revoked.
```
toddysm: Should those two be numbered?

specs/trust-store-trust-policy.md
```diff
@@ -414,10 +411,11 @@ When delta CRLs are implemented, the following results can occur during revocati
 
 ##### OCSP Download
 
-OCSP URLs can be obtained from the certificate's authority information access (AIA) extension as defined in [RFC-2560](https://datatracker.ietf.org/doc/html/rfc2560).
+OCSP URLs can be obtained from the certificate's authority information access (AIA) extension as defined in [RFC-6960](https://datatracker.ietf.org/doc/html/rfc6960).
+If the certificate contains multiple OCSP URLs, then each URL is invoked in sequential order.
 Notary V2 will wait for a default threshold of 5 seconds to receive an OCSP response.
-If OCSP response is not available within the timeout threshold the revocation result will be "revocation unavailable".
 The user may be able to configure this threshold.
+If OCSP response is not available within the timeout threshold the revocation result will be "revocation unavailable".
```
toddysm, line 418: Is it 5 secs for all OCSP URLs or for each individual URL?

toddysm (changes requested):

specs/trust-store-trust-policy.md
yizha1, line 418: I am updating the threat model for OCSP. If OCSP responder is not accessible, could it be a DoS? Since the revocation status is always unavailable, so verify will always fail. If we disable the revocation check, we are at the risk of using an invalid certificate for verification.

priteshbandi: Yes, it is indeed possible for a Denial of Service (DoS) attack on the OCSP endpoint to render the OCSP server inaccessible. In the scenario where the revocation verification level is set to "enforce", this will result in a signature verification failure. However, for other cases (log or skip), the signature verification will succeed.

specs/trust-store-trust-policy.md
priteshbandi:
```suggestion
If any certificate in the certificate chain is revoked or their revocation status cannot be determined (revocation unavailable), fail this step. Depending upon the revocation validation enforcement level specified in the trust policy, this will either lead to signature verification failure or an error being logged
```

Does this helps?

> If the revocation status is revoked doesn't this mean that the certificate is valid? I would suggest to reword to make sure we are clear what the behavior is.

If certificate is revoked than certificate is not valid from verification perspective. IMO valid is an overloaded term, syntactically the certificate is valid because its a valid x509 cert(follows rcf) but semantically its invalid as its revoked.

specs/trust-store-trust-policy.md
priteshbandi: For individual URL, updated the spec.

specs/trust-store-trust-policy.md
priteshbandi: yes, updated

specs/trust-store-trust-policy.md
```diff
@@ -414,10 +411,11 @@ When delta CRLs are implemented, the following results can occur during revocati
 
 ##### OCSP Download
 
-OCSP URLs can be obtained from the certificate's authority information access (AIA) extension as defined in [RFC-2560](https://datatracker.ietf.org/doc/html/rfc2560).
-Notary V2 will wait for a default threshold of 5 seconds to receive an OCSP response.
-If OCSP response is not available within the timeout threshold the revocation result will be "revocation unavailable".
+OCSP URLs can be obtained from the certificate's authority information access (AIA) extension as defined in [RFC 6960](https://www.rfc-editor.org/rfc/rfc6960).
+If the certificate contains multiple OCSP URLs, then each URL is invoked in sequential order.
+For each OCSP URL, Notary V2 will wait for a default threshold of 5 seconds to receive an OCSP response.
```
gokarnm: Can you provide some details on how we arrived at 5 seconds? Why can't it be lower?

s/Notary v2/Notation

priteshbandi: We can have it lower; just for consistency between crl and ocsp I picked 5 sec. Do you have any recommendation?

priteshbandi: As per discussion with Milind will update ocsp timeout to 2 seconds

gokarnm (dismissed):
LGTM!

gokarnm (approved):
LGTM!

shizhMSFT (commented):
LGTM except the term `Notary V2`.

yizha1:
@priteshbandi @shizhMSFT @gokarnm I had [another PR](https://github.com/notaryproject/notaryproject/pull/262) covering the naming issue, do you agree that we only fix the revocation related issue in this PR? If yes, maybe @priteshbandi you can update the PR description or add a comment to make it clear about the scope of this PR and point to the PR for naming changes.

priteshbandi:
> @priteshbandi @shizhMSFT @gokarnm I had [another PR](https://github.com/notaryproject/notaryproject/pull/262) covering the naming issue, do you agree that we only fix the revocation related issue in this PR? If yes, maybe @priteshbandi you can update the PR description or add a comment to make it clear about the scope of this PR and point to the PR for naming changes.

Updated PR description to reflect that, this PR address only revocation spec changes and renaming is done is separate PR.

shizhMSFT (approved):
LGTM

toddysm (approved):
I am OK with the changes in this PR but we need to have another one with the proper terminology.

Issue #266, opened 2023-06-27
Fix broken links and pictures


FeynmanZhou:
There are many broken links and pictures in the notaryproject repository, here are just a couple of examples that I find:

- https://github.com/notaryproject/notaryproject/blob/main/requirements/definitions-terms.md
- https://github.com/notaryproject/notaryproject/blob/main/requirements/verification-by-reference.md#reverse-lookup
- https://github.com/notaryproject/notaryproject/tree/main#project-status

![image](https://github.com/notaryproject/notaryproject/assets/40452856/043e5f13-6bb4-4991-821b-49faa5b9062d)

We need to check those links and pictures and then fix them. @priteshbandi suggested a [GitHub Actions](https://shreeshail.medium.com/how-to-checks-all-markdown-files-in-your-repository-for-broken-links-474c106eec87) scan and detect broken links in CI. We can investigate this solution.

PR #267, opened 2023-06-27
Fix broken media links

sajayantony:
Fixes broken media links that might have happened when moving it to sub dirs.


yizha1 (approved):
LGTM
Thanks @sajayantony

priteshbandi (approved):
LGTM

FeynmanZhou (approved):
Thanks @sajayantony . It fixed a part of the broken media links mentioned in #266 . We will need to scan all documents.

toddysm (approved):

Issue #298, opened 2024-03-04
Relax minimum subject RDN attributes and add parsing for UID attribute.


r-nolasco:
Can we reduce the requirements for the minimum relative distinguished names to remove ST?
Here are a couple examples of others having issues with it as well.

-  https://github.com/notaryproject/specifications/issues/293
-  https://github.com/notaryproject/specifications/issues/287

Wouldn't setting requirements in the trust policy be sufficient?

Would it be possible to add parsing of the UID attribute as well. 
The parser it is using does not expect UID (0.9.2342.19200300.100.1.1) so it is dumping it out in raw form which notation can't handle and errors out with notation does not support x509.subject identities containing \"=#\".

Issue #293, opened 2024-01-29
Relax minimum subject DN field values for trustedIdentities to not include state/province (S/ST)


ianjmcm:
Currently in the [Trust Store and Trust Policy Specification](https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md) in the [Trusted Identities Constraints](https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#trusted-identities-constraints) section there is a minimum field requirement on x.509 cert subject DN values stated as:

"Each identity in identities list MUST contain country (C), state or province (ST), and organization (O) RDNs. All other RDNs are optional. The minimal possible value is `x509.subject: C=${country}, ST=${state}, O={organization}`,"

Not all identities will have a state/province value unless the identity is in the US or Canada, so the `ST` or `S` value need to NOT be required. The minimum subject DN fields should be `CN=`, `O=`, `L=`, `C=`. Signing certs commonly use these values as the minimum for subject DN.

yizha1:
@gokarnm @priteshbandi @shizhMSFT @Two-Hearts Would you mind taking a look at this issue?

priteshbandi:
Hi Ian - 
As per [BR of cabforum-  Section 7.1.4.2](https://cabforum.org/uploads/Baseline-Requirements-for-the-Issuance-and-Management-of-Code-Signing.v3.2.pdf), should it be  either `C=${country}, ST=${state}, O={organization}` Or `C=${country},  L=${localityName}, O={organization}` ? Why do we need CN?

ianjmcm:
CN and O field values are commonly the same values, but there are many cases where a legal tradename or "dba" (doing business as) name can be placed in the O field while the CN value remains to the be legal organization or individual name. That said, we could allow for the minimum to exclude CN as @priteshbandi recommends.

yizha1:
@priteshbandi  I checked the section `7.1.4.2.2` in [specification](https://cabforum.org/uploads/Baseline-Requirements-for-the-Issuance-and-Management-of-Code-Signing.v3.2.pdf), it seems `commonName` is a required field for both EV and non-EV Code Signing Certificates. Would you mind checking it again?

priteshbandi:
[...]

PR #307, opened 2024-07-30
doc: support short name S for state/province

### LINKED ISSUE #287: Support x509 subject with RDN state/province using short name "S"
The current [trust policy and store specification](https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#trusted-identities-constraints) has constraints on x509 subject DN. The state or province RDN need to be `ST=xxx`, for example `ST=Washington`, however `S=` is also an acceptable short name. If users set x509 subject DN with `S=xxx`, the trust policy will fail with the following error:

```
Error: trust policy statement "mypolicy" has trusted identity "x509.subject: xxxxxx" with invalid identity value: distinguished name (DN) "xxxxxx" has no mandatory RDN attribute for "ST", it must contain 'C', 'ST', and 'O' RDN attributes at a minimum
```

This issue is to ask for the support of state/province RDN with short name `S`.

yizha1:
Resolve #287 


yizha1:
@priteshbandi @shizhMSFT @Two-Hearts @JeyJeyGao Would you mind taking a look at this PR?

ghost (approved):
LGTM

shizhMSFT (commented):
Will this PR cover #293 and #298?

yizha1:
@shizhMSFT no, only fixing 287 to support additional `S`.

shizhMSFT (approved):
LGTM. It is better to have a reference where `S` and `ST` come from.

specs/trust-store-trust-policy.md
```diff
@@ -442,9 +442,9 @@ The RDN consists of an attribute type name followed by an equal sign and the str
   - The value of each `trustedIdentities` list item, if it begins with `x509.subject:`, MUST be followed by comma-separated one or more RDNs.
     Other types of trusted identities may be supported, by using an alternate prefix, or a different format.
     For example, `x509.subject: C=${country}, ST=${state}, L=${locallity}, O={organization}, OU=${organization-unit}, CN=${common-name}`.
-  - Each identity in `identities` list MUST contain country (C), state or province (ST), and organization (O) RDNs.
+  - Each identity in `identities` list MUST contain country (C), state or province (ST or S), and organization (O) RDNs.
     All other RDNs are optional.
-    The minimal possible value is `x509.subject: C=${country}, ST=${state}, O={organization}`,
+    The minimal possible value is `x509.subject: C=${country}, ST=${state}, O={organization}` or `x509.subject: C=${country}, S=${state}, O={organization}`.
```
priteshbandi, line 447: Shouldn't it be` C=${country}, ST=${state}, O={organization}` Or `C=${country},  L=${localityName}, O={organization}` ?

yizha1: This PR is only for the short name `S` that can be used for province/state as well, currently only `ST` is allowed, but `S` is also correct. Using `L` is a different issue, which is tracked here https://github.com/notaryproject/specifications/issues/293, but there is no decision yet.

toddysm (approved):

specs/trust-store-trust-policy.md
priteshbandi: Can you share RFC which defines `S` and is oid for `S` different from [`ST`](http://www.oid-info.com/get/2.5.4.8) ?

yizha1: @priteshbandi the oid is the same. In the section "6.3.3" of [ITU-X.520](https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-X.520-201210-S!!PDF-E&type=items), `S` is used in the example, although `ST` is used for the LDAP name. Another doc also suggests `S` not `ST`, see https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certnametostra?redirectedfrom=MSDN#remarks. So, there are some cases that `S` is used. Allowing the use of either 'S' or 'ST' as the field label for State/Province will be beneficial.

priteshbandi (approved):
LGTM

priteshbandi:
Can you please open an issue in notation-go to implement this?

PR #312, opened 2024-08-09
fix: correct typo for the registry scopes field name

ghost:
This is a quick fix as we don't have the term `scopes` in the OCI trust policy, the name should be `registryScopes` instead.


JeyJeyGao (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #319, opened 2024-10-29
fix(doc): reponse status code should be `200`

### LINKED ISSUE #313: CRL response status should be 200
https://github.com/notaryproject/specifications/blob/00abcea0ad35bb80c74cb82890834440bf8f218d/specs/trust-store-trust-policy.md?plain=1#L567

As mentioned in the spec, the response status code could be `2xx`. However, it doesn't make sense if the status code is `204 No Content`. Therefore, we need to ensure the status code is `200`.

JeyJeyGao:
As the security audit team suggested, the response status code should be `200` instead of `2xx`.

Fix:
- updated the response status code to be `200` for both CRL and OCSP spec

Resolves #313

### DIFF
--- a/specs/trust-store-trust-policy.md
+++ b/specs/trust-store-trust-policy.md
@@ -564,7 +564,7 @@ Implementations MAY add support for caching CRLs and OCSP response to improve av
 ##### CRL Download
 
 CRL download location (URL) can be obtained from the certificate's CRL Distribution Point (CDP) extension.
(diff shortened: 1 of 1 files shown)


shizhMSFT (approved):
LGTM

FeynmanZhou (approved):
LGTM

ghost (approved):
LGTM

yizha1 (approved):
LGTM

priteshbandi (approved):
LGTM

PR #325, opened 2025-03-31
docs: add `artifactType` to signature spec

### LINKED ISSUE #295: Update specifications to accommodate image-spec/v1.1.0
https://github.com/opencontainers/image-spec/releases/tag/v1.1.0

FeynmanZhou:
This is the spec change for https://github.com/notaryproject/notation/issues/975
This PR adds an "artifactType" property to signature spec, which aligns with the [OCI image spec v1.1](https://github.com/opencontainers/image-spec/blob/v1.1.0/manifest.md#image-manifest-property-descriptions). The OCI registry can leverage this property "artifactType" to filter out Notary Project signatures from various OCI artifacts.


specs/signature-specification.md
```diff
@@ -207,6 +209,11 @@ Besides the [image manifest property requirements][image-manifest-property-descr
   Keys using the `io.cncf.notary` namespace are reserved for use in the Notary Project signature specification and MUST NOT be used by other specifications.
   - **`io.cncf.notary.x509chain.thumbprint#S256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprints of signing certificate and certificate chain (including root) used for signature generation. The list of fingerprints is present as a JSON array string, corresponding to ordered certificates in [*Certificate Chain* unsigned attribute](#unsigned-attributes) in the signature envelope. The annotation name contains the hash algorithm as a suffix (`#S256`) and can be extended to support other hashing algorithms in future.
 
+#### OCI Signature Verification Backward Compatibility  
+
```
FeynmanZhou, line 250: Forward compatibility support should be considerred as well.

JeyJeyGao:
What happens if the registry is not OCI 1.1 compliant and does not accept the `artifactType` field?

baarde:
> What happens if the registry is not OCI 1.1 compliant and does not accept the `artifactType` field?

According to [OCI 1.0 Image specification](https://specs.opencontainers.org/image-spec/considerations/?v=v1.0.0#extensibility), nothing should happen.
> Implementations that are reading/processing [manifests](https://specs.opencontainers.org/image-spec/manifest/?v=v1.0.0) or [image indexes](https://specs.opencontainers.org/image-spec/image-index/?v=v1.0.0) MUST NOT generate an error if they encounter an unknown property. Instead they MUST ignore unknown properties.

For OCI 1.0 registries, signature and verification should continue to work as before. The `subject` and `artifactType` fields in the signature manifest are ignored. The `artifactType` is retrieved from the manifest descriptor in the referrers tag's image index.

specs/signature-specification.md
baarde: According to my tests, new signatures remain compatible with older versions of Notation. I was able to list* the signature with version 1.0.0-rc.1.

\*`list` only. I did not try to `verify` the signature because my OCI 1.1 registry has a bug in Referrers API (see notaryproject/notation#1222)

FeynmanZhou: Thanks for testing `notation` with GAR!

specs/signature-specification.md
```diff
@@ -193,8 +194,9 @@ Signature manifest example per OCI image manifest:
 Besides the [image manifest property requirements][image-manifest-property-descriptions], the properties have the following additional restrictions:
 
 - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.image.manifest.v1+json`.
+- **`artifactType`** (*string*): This REQUIRED property MUST be `application/vnd.cncf.notary.signature`.
 - **`config`** (*descriptor*): This property is REQUIRED to be compatible with [OCI image specification][oci-image-manifest]. The Notary Project signature specification doesn't require any configuration for a signature, and the configuration content is not consumed by implementations of the Notary Project signature specification.
-  - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.cncf.notary.signature`.
+  - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.empty.v1+json`.
```
ghost: On the specifications side:
This is a breaking change to the `specifications v1`.

On the implementation side of Notation:
Notation v1 filters signatures by `Config.MediaType` with value `application/vnd.cncf.notary.signature`. Therefore, if this PR gets merged and released in `specifications v1`, Notation v1 will no longer implements the Notary project spec. 
In other words, if Notation v1 does make the change accordingly, all current signatures signed by Notation v1 will no longer be listed or verified, which is a regression.

Potential solution:
If we only want one version of Notary project spec, i.e. `specifications v1`, then we should have a fall back mechanism on `Config.MediaType`:
1. Instead of requiring `artifactType`, make it optional. This aligns with OCI 1.1 spec: https://github.com/opencontainers/image-spec/blob/v1.1.0/manifest.md, in which, `artifactType` is an optional field.
2. When `artifactType` exists and the value is `application/vnd.cncf.notary.signature`, use it. `Config.MediaType` must be set to `application/vnd.oci.empty.v1+json`.
3. When `artifactType` does not exist OR its value is not `application/vnd.cncf.notary.signature`, fall back to use `Config.MediaType`, which must be set to `application/vnd.cncf.notary.signature`.

specs/signature-specification.md
```diff
@@ -155,8 +155,8 @@ These attributes are considered unsigned with respect to the signing key that ge
 ## OCI Signatures
 
 This section describes how a Notary Project signature is stored in an OCI Distribution conformant registry.
-OCI image manifest is used to store signatures in the registry, see [OCI image spec v1.1.0-rc3][oci-image-manifest] for details.
-The signature manifest has a configuration media type that specifies it's a Notary Project signature, a subject referencing the manifest of the artifact being signed, a layer referencing the signature, and a collection of annotations.
+OCI image manifest is used to store signatures in the registry, see [OCI image spec v1.1.0][oci-image-manifest] for details. Notary Project signature is compliant with OCI image spec v1.1.0.
+Starting from `notation` v2.0.0, the signature manifest uses the `artifactType` property that specifies it's a Notary Project signature, a subject referencing the manifest of the artifact being signed, a layer referencing the signature, and a collection of annotations.
```
ghost: Please be aware, this is the Notary Project specifications repo. 
We shouldn't mention `notation v2.0.0` here, as the spec is designed to be implemented by others as well.

specs/signature-specification.md
```diff
@@ -207,6 +209,16 @@ Besides the [image manifest property requirements][image-manifest-property-descr
...
 
+For OCI v1.0 compliant registries, the `subject` and `artifactType` fileds in the signature manifest are ignored by the registry. Notation v2.x SHOULD fall back to use OCI referrers tag schema to push and store the signature in the registry.
+
+#### OCI Signature Verification Forward and Backward Compatibility  
+
+The Notary Project signature uses the `artifactType` property to indicate its type starting from `notation` v2.0.0, whereas older versions rely on `config.mediatype`.  
+
+To minimize the impact of breaking changes in the signature manifest, `notation` v2.x SHOULD be able to list, inspect, and verify the signature produced by `notation` v1.x. It ensures backward compatibility with older OCI signatures.
+
+Forward compatibility SHOULD also be considered in `notation` v1.x since it reduces friction in updrades and won't make older workflows suddenly fail when verifying a newer signature. `notation` v1.x SHOULD be able to list, inspect, and verify the signature produced by `notation` v2.0.
```
ghost: Same as above, this is the Notary Project spec repo. Discussions on Notation should happen in the Notation repo.

ghost (changes requested):

specs/signature-specification.md
JeyJeyGao: I agree that `artifactType` is an optional field; we should keep the support for `Config.MediaType`.

JeyJeyGao: Another possibility: signing migrate to `artifactType`, but verification supports both `Config.MediaType` and `artifactType` for backward compatibility.

ghost: > Another possibility: signing migrate to `artifactType`, but verification supports both `Config.MediaType` and `artifactType` for backward compatibility.

That's what I mean. For `signing`, all new signatures should use the `artifactType` field.

FeynmanZhou: @Two-Hearts Good point. The proposal of making `artifactType` optional and having fallback mechanism is less breaking to `notation` v1.x and also the Notary Project Specifications v1.x.

FeynmanZhou: I've incorporated the suggestions from @Two-Hearts per the discussion above.
 
cc @notaryproject/notaryproject-notaryproject-maintainers @yizha1 @toddysm PTAL. Thanks.

specs/signature-specification.md
```diff
@@ -155,8 +155,8 @@ These attributes are considered unsigned with respect to the signing key that ge
 ## OCI Signatures
 
 This section describes how a Notary Project signature is stored in an OCI Distribution conformant registry.
-OCI image manifest is used to store signatures in the registry, see [OCI image spec v1.1.0-rc3][oci-image-manifest] for details.
-The signature manifest has a configuration media type that specifies it's a Notary Project signature, a subject referencing the manifest of the artifact being signed, a layer referencing the signature, and a collection of annotations.
+OCI image manifest is used to store signatures in the registry, see [OCI image spec v1.1.0][oci-image-manifest] for details. Notary Project signature is compliant with OCI image spec v1.1.0.
```
shizhMSFT: Should we update it to `v1.1.1`?

specs/signature-specification.md
shizhMSFT: Can `notation` 2.0.0 opt to produce the old-style manifest with a toggled flag?

specs/signature-specification.md
```diff
@@ -193,8 +194,9 @@ Signature manifest example per OCI image manifest:
 Besides the [image manifest property requirements][image-manifest-property-descriptions], the properties have the following additional restrictions:
 
 - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.image.manifest.v1+json`.
+- **`artifactType`** (*string*): This OPTIONAL property MUST be `application/vnd.cncf.notary.signature`.
```
shizhMSFT: When `config.mediaType` is empty, this field is required, isn't it?

OCI image-spec `v1.1.1` states for the `artifactType` field:
> This OPTIONAL property contains the type of an artifact when the manifest is used for an artifact. This MUST be set when config.mediaType is set to the [empty value](https://github.com/opencontainers/image-spec/blob/v1.1.1/manifest.md#guidance-for-an-empty-descriptor).

specs/signature-specification.md
```diff
@@ -207,6 +209,14 @@ Besides the [image manifest property requirements][image-manifest-property-descr
   Keys using the `io.cncf.notary` namespace are reserved for use in the Notary Project signature specification and MUST NOT be used by other specifications.
   - **`io.cncf.notary.x509chain.thumbprint#S256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprints of signing certificate and certificate chain (including root) used for signature generation. The list of fingerprints is present as a JSON array string, corresponding to ordered certificates in [*Certificate Chain* unsigned attribute](#unsigned-attributes) in the signature envelope. The annotation name contains the hash algorithm as a suffix (`#S256`) and can be extended to support other hashing algorithms in future.
 
+For OCI v1.0 compliant registries, the `subject` and `artifactType` fileds in the signature manifest are ignored by the registry. Notation v2.x SHOULD fall back to use OCI referrers tag schema to push and store the signature in the registry.
```
shizhMSFT: `OCI v1.0` is not accurate, please specify whether it is image-spec or distribution-spec or both.

specs/signature-specification.md
ghost: Again, this is the Notary Project spec for all implementations. Details regarding `Notation v2.0.0` should be put into Notation CLI's docs/spec.

specs/signature-specification.md
ghost: @FeynmanZhou Your most recent update doesn't seem to align with my original proposal. Below is incorrect:
```
**`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.empty.v1+json`
```
Please read through the proposed solution again.

More specifically, the `artifactType` field is REQUIRED when `config.MediaType` is empty. When artifactType does not exist OR its value is not application/vnd.cncf.notary.signature, fall back to use Config.MediaType, which must be set to application/vnd.cncf.notary.signature.

specs/signature-specification.md
ghost: @FeynmanZhou your change doesn't align with my original proposed solution. I left a detailed comment in the original discussion: https://github.com/notaryproject/specifications/pull/325#discussion_r2043612510. PTAL.

specs/signature-specification.md
yizha1: This is the specification, and Notation is one of the reference implementations. So, it is not recommended to mention `Notation` in the spec. The b.ehavior of `Notation` should be defined in Notation CLI spec.

The spec shoud include the following:

- Describe there are two formats per the OCI 1.1 spec
  - What is the new format and the example. The new format has  `artifactType` and `config.mediaType` set to empty
  - What is the old format, which has no `artifactType` and `config.mediaType` set to the notary project signature type. We already have the description and example.

- For producing signature, 
   - The new format is recommended as the default format, but reference implementation can decide based on their own implementation and provide options for user to specify the format. for example, in Notation CLI spec, it can use the new format as the default starting from Notation 2.0, and provide an option for users to switch to the old format for some reasons)

-  For consuming signature
   - Both new format and old format should be supported
   - Provide streamlined experience and backward compatibility , if one format is not matched, then check another format. Fail if both formats are not matched. 

/cc @shizhMSFT @Two-Hearts @JeyJeyGao

specs/signature-specification.md
FeynmanZhou: @Two-Hearts I've updated it per your suggestions. Would you mind reviewing it again? I listed two cases for two different siganture formats (default and legacy)

specs/signature-specification.md
FeynmanZhou: @shizhMSFT Updated.

specs/signature-specification.md
FeynmanZhou: @yizha1 @shizhMSFT Updated per your suggestions. Thanks.
I will raise another PR to update `notation` CLI to include the new flag.

specs/signature-specification.md
FeynmanZhou: Updated

specs/signature-specification.md
FeynmanZhou: Updated and removed `notation`. I use "signing agent" to represent the implementation reference which produces siganture.

specs/signature-specification.md
FeynmanZhou: Updated per your suggestions. PTAL again. Thanks!

.DS_Store
ghost, line 1: What does this file do?

media/new-oci-signature-manifest.svg
ghost, line 1: Do we want to add the `Config` field under `The Notary Project Signature Manifest`?

specs/signature-specification.md
```diff
@@ -155,19 +155,21 @@ These attributes are considered unsigned with respect to the signing key that ge
...
+OCI image manifest is used to store signatures in the registry, see [OCI image spec v1.1.1][oci-image-manifest] for details. 
 
-![Signature storage inside registry](../media/oci-signature-specification.svg)
+The signature manifest uses either the `artifactType` property or `config.mediaType` to indicate it's a Notary Project signature, a subject referencing the manifest of the artifact being signed, a layer referencing the signature, and a collection of annotations.
+
+![Signature storage inside registry](../media/new-oci-signature-specification.svg)
 
-Signature manifest example per OCI image manifest:
+- For registries compliant with OCI Distribution Spec v1.1, the Notary Project signature manifest by default uses the required `artifactType` property to indicate it's a Notary Project signature when `config.mediaType` is empty. The signing agent is recommended to produce the signature with `artifactType` property by default. The sample signature manifest is as follows:
```
ghost: We should mention the value `application/vnd.cncf.notary.signature` here.

specs/signature-specification.md
```diff
@@ -193,8 +195,9 @@ Signature manifest example per OCI image manifest:
 Besides the [image manifest property requirements][image-manifest-property-descriptions], the properties have the following additional restrictions:
 
 - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.image.manifest.v1+json`.
+- **`artifactType`** (*string*): This REQUIRED property MUST be `application/vnd.cncf.notary.signature`.
```
ghost: `artifactType` shouldn't be REQUIRED, otherwise, it's a breaking change to Notary Project spec v1.
It should be OPTIONAL, when it's missing, `config.MediaType` is REQUIRED and the value MUST be `application/vnd.cncf.notary.signature`.

specs/signature-specification.md
```diff
@@ -193,8 +195,9 @@ Signature manifest example per OCI image manifest:
...
 - **`config`** (*descriptor*): This property is REQUIRED to be compatible with [OCI image specification][oci-image-manifest]. The Notary Project signature specification doesn't require any configuration for a signature, and the configuration content is not consumed by implementations of the Notary Project signature specification.
-  - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.cncf.notary.signature`.
+  - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.empty.v1+json`.
```
ghost: This is a breaking change to Notary Project spec v1.

specs/signature-specification.md
```diff
@@ -207,6 +210,46 @@ Besides the [image manifest property requirements][image-manifest-property-descr
...
 
+- For registries compliant with OCI Distribution Spec v1.0, the `subject` and `artifactType` fileds in the signature manifest are ignored by the registry. The signing agent SHOULD fall back to use OCI referrers tag schema to produce a Notary Project signature using a legacy format. 
+
+![Signature storage inside registry](../media/oci-signature-specification.svg)
+
+In former versions of Notary Project signature speocification, the legacy Notary Project signature manifest uses the required property `config.mediaType` to indicate it's a Notary Project signature when `artifactType` does not exist OR its value is not `application/vnd.cncf.notary.signature`. The sample legacy signature manifest is as follows:
```
ghost:
```suggestion
In former versions of Notary Project signature specification, the legacy Notary Project signature manifest uses the required property `config.mediaType` to indicate it's a Notary Project signature when `artifactType` does not exist OR its value is not `application/vnd.cncf.notary.signature`. The sample legacy signature manifest is as follows:
```

specs/signature-specification.md
```diff
@@ -207,6 +210,46 @@ Besides the [image manifest property requirements][image-manifest-property-descr
...
+        "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\",\"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
+    }
+}
+```
+
+#### OCI Signature Verification Forward and Backward Compatibility  
+
+When signing an artifact, the Notary Project signature uses the `artifactType` property to indicate its type, whereas older versions rely on `config.mediatype`.  
+
+To minimize the impact of breaking changes in the signature manifest, verifier SHOULD be able to list, inspect, and verify either the new or legacy format of Notary Project signature. It ensures forward and backward compatibility with OCI signatures. It reduces friction in updrades and won't make older workflows suddenly fail when verifying a newer signature.
```
ghost:
```suggestion
To minimize the impact of breaking changes in the signature manifest, verifier SHOULD be able to list, inspect, and verify either the new or legacy format of Notary Project signature. It ensures forward and backward compatibility with OCI signatures. It reduces friction in upgrades and won't make older workflows suddenly fail when verifying a newer signature. 
```

specs/signature-specification.md
FeynmanZhou: @Two-Hearts I changed to
> **`mediaType`** (*string*): This property is REQUIRED. Its value MUST be `application/vnd.cncf.notary.signature` when `artifactType` does not exist. If `artifactType` is set to `application/vnd.cncf.notary.signature`, uses `application/vnd.oci.empty.v1+json` as its value.

.DS_Store
FeynmanZhou: Removed

media/new-oci-signature-manifest.svg
FeynmanZhou: Updated. Is the `Config` supposed to be inside the siganture manifest, just like this?
image

specs/signature-specification.md
```diff
@@ -193,8 +195,9 @@ Signature manifest example per OCI image manifest:
 Besides the [image manifest property requirements][image-manifest-property-descriptions], the properties have the following additional restrictions:
 
 - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.image.manifest.v1+json`.
+- **`artifactType`** (*string*): This OPTIONAL property MUST be `application/vnd.cncf.notary.signature`.
 - **`config`** (*descriptor*): This property is REQUIRED to be compatible with [OCI image specification][oci-image-manifest]. The Notary Project signature specification doesn't require any configuration for a signature, and the configuration content is not consumed by implementations of the Notary Project signature specification.
-  - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.cncf.notary.signature`.
+  - **`mediaType`** (*string*): This property is REQUIRED. Its value MUST be `application/vnd.cncf.notary.signature` when `artifactType` does not exist. If `artifactType` is set to `application/vnd.cncf.notary.signature`, uses `application/vnd.oci.empty.v1+json` as its value.
```
ghost:
```suggestion
  - **`mediaType`** (*string*): This property is REQUIRED. Its value MUST be `application/vnd.cncf.notary.signature` when `artifactType` does not exist. When `artifactType` exists, `mediaType` MUST be `application/vnd.oci.empty.v1+json`.
```

specs/signature-specification.md
```diff
@@ -207,6 +210,46 @@ Besides the [image manifest property requirements][image-manifest-property-descr
   Keys using the `io.cncf.notary` namespace are reserved for use in the Notary Project signature specification and MUST NOT be used by other specifications.
   - **`io.cncf.notary.x509chain.thumbprint#S256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprints of signing certificate and certificate chain (including root) used for signature generation. The list of fingerprints is present as a JSON array string, corresponding to ordered certificates in [*Certificate Chain* unsigned attribute](#unsigned-attributes) in the signature envelope. The annotation name contains the hash algorithm as a suffix (`#S256`) and can be extended to support other hashing algorithms in future.
 
+- For registries compliant with OCI Distribution Spec v1.0, the `subject` and `artifactType` fileds in the signature manifest are ignored by the registry. The signing agent SHOULD fall back to use OCI referrers tag schema to produce a Notary Project signature using a legacy format which does not contain `artifactType` property.
```
ghost:
```suggestion
- For registries compliant with OCI Distribution Spec v1.0, the `subject` and `artifactType` fields in the signature manifest are ignored by the registry. The signing agent SHOULD fall back to use OCI referrers tag schema to produce a Notary Project signature using a legacy format which does not contain `artifactType` property. 
```

specs/signature-specification.md
```diff
@@ -155,19 +155,21 @@ These attributes are considered unsigned with respect to the signing key that ge
 ## OCI Signatures
 
 This section describes how a Notary Project signature is stored in an OCI Distribution conformant registry.
-OCI image manifest is used to store signatures in the registry, see [OCI image spec v1.1.0-rc3][oci-image-manifest] for details.
-The signature manifest has a configuration media type that specifies it's a Notary Project signature, a subject referencing the manifest of the artifact being signed, a layer referencing the signature, and a collection of annotations.
+OCI image manifest is used to store signatures in the registry, see [OCI image spec v1.1.1][oci-image-manifest] for details. 
 
-![Signature storage inside registry](../media/oci-signature-specification.svg)
+The signature manifest uses either the `artifactType` property or `config.mediaType` to indicate it's a Notary Project signature, a subject referencing the manifest of the artifact being signed, a layer referencing the signature, and a collection of annotations.
```
ghost:
```suggestion
The signature manifest uses either the `artifactType` property or `config.mediaType` to indicate it's a Notary Project signature. It also contains a subject to reference the manifest of the artifact being signed, a layer to reference the signature, and a collection of annotations.
```

specs/signature-specification.md
```diff
@@ -155,19 +155,21 @@ These attributes are considered unsigned with respect to the signing key that ge
...
+
+![Signature storage inside registry](../media/new-oci-signature-specification.svg)
 
-Signature manifest example per OCI image manifest:
+- For registries compliant with OCI Distribution Spec v1.1, the Notary Project signature manifest by default uses the required `artifactType: application/vnd.cncf.notary.signature` property to indicate it's a Notary Project signature when `config.mediaType` is `application/vnd.oci.empty.v1+json`. The signing agent is recommended to produce the signature with `artifactType` property by default. The sample signature manifest is as follows:
```
ghost:
```suggestion
- For registries compliant with OCI Image Spec v1.1, the Notary Project signature manifest, by default, uses `artifactType: application/vnd.cncf.notary.signature` property to indicate it's a Notary Project signature. In this case, `config.mediaType` MUST be set to `application/vnd.oci.empty.v1+json`. The signing agent is recommended to produce the signature with `artifactType` property by default. When `artifactType` is not used, `config.mediaType` MUST be set to `application/vnd.cncf.notary.signature`. The sample signature manifest is as follows:
```

specs/signature-specification.md
```diff
@@ -193,8 +195,9 @@ Signature manifest example per OCI image manifest:
 Besides the [image manifest property requirements][image-manifest-property-descriptions], the properties have the following additional restrictions:
 
 - **`mediaType`** (*string*): This REQUIRED property MUST be `application/vnd.oci.image.manifest.v1+json`.
+- **`artifactType`** (*string*): This OPTIONAL property MUST be `application/vnd.cncf.notary.signature`.
```
ghost:
```suggestion
- **`artifactType`** (*string*): This OPTIONAL property, if present, MUST be `application/vnd.cncf.notary.signature`.
```

specs/signature-specification.md
```diff
@@ -207,6 +210,46 @@ Besides the [image manifest property requirements][image-manifest-property-descr
...
+    "annotations": {
+        "io.cncf.notary.x509chain.thumbprint#S256": 
+        "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\",\"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
+    }
+}
+```
+
+#### OCI Signature Verification Forward and Backward Compatibility  
+
+When signing an artifact, the Notary Project signature uses the `artifactType` property to indicate its type, whereas older versions rely on `config.mediatype`.
```
ghost:
```suggestion
When signing an artifact, the Notary Project signature uses the `artifactType` property to indicate its type, by default, whereas older versions rely on `config.mediatype`.  
```

specs/signature-specification.md
FeynmanZhou: Updated

specs/signature-specification.md
FeynmanZhou: Updated

specs/signature-specification.md
```diff
@@ -207,6 +210,46 @@ Besides the [image manifest property requirements][image-manifest-property-descr
   Keys using the `io.cncf.notary` namespace are reserved for use in the Notary Project signature specification and MUST NOT be used by other specifications.
   - **`io.cncf.notary.x509chain.thumbprint#S256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprints of signing certificate and certificate chain (including root) used for signature generation. The list of fingerprints is present as a JSON array string, corresponding to ordered certificates in [*Certificate Chain* unsigned attribute](#unsigned-attributes) in the signature envelope. The annotation name contains the hash algorithm as a suffix (`#S256`) and can be extended to support other hashing algorithms in future.
 
+- For registries compliant with OCI Distribution Spec v1.0, the `subject` and `artifactType` fileds in the signature manifest are ignored by the registry. The signing agent SHOULD fall back to use OCI referrers tag schema to produce a Notary Project signature using a legacy format which does not contain `artifactType` property. 
+
```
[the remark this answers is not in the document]
FeynmanZhou: Updated

specs/signature-specification.md
```diff
@@ -207,6 +210,46 @@ Besides the [image manifest property requirements][image-manifest-property-descr
...
+        "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\",\"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
+    }
+}
+```
+
+#### OCI Signature Verification Forward and Backward Compatibility  
+
+When signing an artifact, the Notary Project signature uses the `artifactType` property to indicate its type, whereas older versions rely on `config.mediatype`.  
+
+To minimize the impact of breaking changes in the signature manifest, verifier SHOULD be able to list, inspect, and verify either the new or legacy format of Notary Project signature. It ensures forward and backward compatibility with OCI signatures. It reduces friction in upgrades and won't make older workflows suddenly fail when verifying a newer signature.
```
ghost, line 251: Just an open discussion:
I think this design is still not backward compatible though. If a new signature manifest is using the `artifactType` way, then an old Notary Project implementation cannot take it as a notary signature? Because any old implementation of this spec is expecting the `config.MediaType`.
But I guess the above is not really an issue? (the signer has the option to create the signature with the old way.)

specs/signature-specification.md
FeynmanZhou: I revised it based on GPT's suggeston https://chatgpt.com/share/6807022d-a95c-8001-a88b-6660c1ce2647

specs/signature-specification.md
```diff
@@ -155,12 +155,46 @@ These attributes are considered unsigned with respect to the signing key that ge
...
+        "size": 16724
+    },
+    "annotations": {
+        "io.cncf.notary.x509chain.thumbprint#S256": 
+        "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\",\"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
+    }
+}
+```
+
+- For registries support OCI Image Spec v1.0, the `subject` and `artifactType` fields in the signature manifest are ignored by the registry. The signing agent SHOULD fall back to use OCI referrers tag schema to produce a Notary Project signature using a legacy format which does not contain `artifactType` property as follows::
```
ghost:
```suggestion
- For registries support OCI Image Spec v1.0 only, the `artifactType` field in the signature manifest is ignored by the registry. The required property `config.mediaType: "application/vnd.cncf.notary.signature"` is used to indicate Notary Project signature.

A sample signature manifest is as follows:
```

specs/signature-specification.md
```diff
@@ -190,11 +224,14 @@ Signature manifest example per OCI image manifest:
 }
 ```
 
+In former versions of Notary Project signature specification, the legacy Notary Project signature manifest uses the required property `config.mediaType: "application/vnd.cncf.notary.signature"` to indicate it's a Notary Project signature when `artifactType` does not exist, OR its value is not `application/vnd.cncf.notary.signature`.
```
ghost: This is duplicated as above. We can remove it.
```suggestion 
```

specs/signature-specification.md
ghost: Based on our offline discussion, verifier will need to support verifying both new and legacy format signatures. 
Please feel free to resolve this comment.

ghost (approved):
LGTM

specs/signature-specification.md
FeynmanZhou: Per our discussion, the signing agent (e.g. `notation`) can choose to support both new and old signature specification to achieve forward and backward compability.

media/new-oci-signature-manifest.svg
shizhMSFT, line 1: The image is not compact. There are empty spaces on the top.

specs/signature-specification.md
```diff
@@ -155,12 +155,46 @@ These attributes are considered unsigned with respect to the signing key that ge
...
 
 This section describes how a Notary Project signature is stored in an OCI Distribution conformant registry.
-OCI image manifest is used to store signatures in the registry, see [OCI image spec v1.1.0-rc3][oci-image-manifest] for details.
-The signature manifest has a configuration media type that specifies it's a Notary Project signature, a subject referencing the manifest of the artifact being signed, a layer referencing the signature, and a collection of annotations.
+OCI image manifest is used to store signatures in the registry, see [OCI image spec v1.1.1][oci-image-manifest] for details. 
 
-![Signature storage inside registry](../media/oci-signature-specification.svg)
+The signature manifest uses either the `artifactType` property or `config.mediaType` to indicate it's a Notary Project signature. It also contains a subject to reference the manifest of the artifact being signed, a layer to reference the signature, and a collection of annotations.
+
+![Signature storage inside registry](../media/new-oci-signature-specification.svg)
```
shizhMSFT: The pic link is broken.

specs/signature-specification.md
```diff
@@ -155,12 +155,46 @@ These attributes are considered unsigned with respect to the signing key that ge
...
+    },
+    "annotations": {
+        "io.cncf.notary.x509chain.thumbprint#S256": 
+        "[\"B7A69A70992AE4F9FF103EBE04A2C3BA6C777E439253CE36562E6E98375068C3\",\"932EB6F5598435D4EF23F97B0B5ACB515FAE2B8D8FAC046AB813DDC419DD5E89\"]"
+    }
+}
+```
 
-Signature manifest example per OCI image manifest:
+- For registries support OCI Image Spec v1.0 only, the `artifactType` field in the signature manifest is ignored by the registry. The required property `config.mediaType: "application/vnd.cncf.notary.signature"` is used to indicate Notary Project signature:
```
shizhMSFT: Not just "ignored'... some registries reject manifest with unknown fields.

specs/signature-specification.md
```diff
@@ -207,6 +242,14 @@ Besides the [image manifest property requirements][image-manifest-property-descr
   Keys using the `io.cncf.notary` namespace are reserved for use in the Notary Project signature specification and MUST NOT be used by other specifications.
   - **`io.cncf.notary.x509chain.thumbprint#S256`**: A REQUIRED annotation whose value contains the list of SHA-256 fingerprints of signing certificate and certificate chain (including root) used for signature generation. The list of fingerprints is present as a JSON array string, corresponding to ordered certificates in [*Certificate Chain* unsigned attribute](#unsigned-attributes) in the signature envelope. The annotation name contains the hash algorithm as a suffix (`#S256`) and can be extended to support other hashing algorithms in future.
 
+![Signature storage inside registry](../media/oci-signature-specification.svg)
+
+#### OCI Signature Verification Forward and Backward Compatibility  
+
+When signing an artifact, the Notary Project signature uses the `artifactType` property by default to indicate the artifact's type, whereas older versions rely on `config.mediatype`.
```
shizhMSFT:
```suggestion
When signing an artifact, the Notary Project signature uses the `artifactType` property by default to indicate the artifact's type, whereas older versions rely on `config.mediaType`.  
```

shizhMSFT: Define `older versions`

specs/signature-specification.md
FeynmanZhou: I updated to  

> For registries support OCI Image Spec v1.0 only, the `artifactType` field in the signature manifest might be ignored or rejected by the registry.

Do you have an example of the registry that may reject `artifactType` field? This behavior has not been defined in https://github.com/opencontainers/distribution-spec/blob/v1.0.1/spec.md.

specs/signature-specification.md
FeynmanZhou: Updated

specs/signature-specification.md
FeynmanZhou: Fixed.

media/new-oci-signature-manifest.svg
FeynmanZhou: Fixed the image problem

media/.DS_Store
JeyJeyGao, line 1: Please remove this file.

specs/signature-specification.md
shizhMSFT: It's implementation specific and the v1.0.x spec does not restrict the implementation to reject unknown fields like `artifactType`. Just be aware of the wording.

specs/signature-specification.md
```diff
@@ -155,12 +155,46 @@ These attributes are considered unsigned with respect to the signing key that ge
...
+OCI image manifest is used to store signatures in the registry, see [OCI image spec v1.1.1][oci-image-manifest] for details. 
 
-![Signature storage inside registry](../media/oci-signature-specification.svg)
+The signature manifest uses either the `artifactType` property or `config.mediaType` to indicate it's a Notary Project signature. It also contains a subject to reference the manifest of the artifact being signed, a layer to reference the signature, and a collection of annotations.
+
+![Signature storage inside registry](../media/new-oci-signature-manifest-diagram.svg)
+
+- For registries support the OCI Image Spec v1.1, the Notary Project signature manifest uses the `artifactType: "application/vnd.cncf.notary.signature"` property by default to indicate that it is a Notary Project signature. In this case, the `config.mediaType` MUST be set to `application/vnd.oci.empty.v1+json`. The signing agent is recommended to use the `artifactType` property by default when producing signatures. If the `artifactType property` is not used, then `config.mediaType` MUST be set to `application/vnd.cncf.notary.signature`.
```
yizha1:
```suggestion
- For registries support the OCI Image Spec v1.1, the Notary Project signature manifest uses the `artifactType: "application/vnd.cncf.notary.signature"` property by default to indicate that it is a Notary Project signature. In this case, the `config.mediaType` MUST be set to `application/vnd.oci.empty.v1+json`. The signing agent is recommended to use the `artifactType` property by default when producing signatures. If the `artifactType` property is not used, then `config.mediaType` MUST be set to `application/vnd.cncf.notary.signature`. 
```

yizha1 (dismissed):
LGTM with a nit

FeynmanZhou:
@Two-Hearts @yizha1 Would you mind approving it again? Thanks

ghost (approved):
LGTM

shizhMSFT (approved):
LGTM

yizha1 (approved):
LGTM

FeynmanZhou:
Need one more approval to get it merged. @vaninrao10 @toddysm @priteshbandi Can anyone help review or approve it? Thanks

toddysm (approved):
LGTM

PR #328, opened 2025-05-15
docs: add support of COSE Hash Envelope

ghost:
This PR adds COSE Hash Envelope to Notary Project spec.


specs/signature-envelope-cose.md
```diff
@@ -239,6 +239,52 @@ The final signature envelope is a `COSE_Sign1_Tagged` object, consisting of Payl
 )
 ```
 
+## COSE Hash Envelope
+Notary Project also supports [COSE Hash Envelope](https://datatracker.ietf.org/doc/draft-ietf-cose-hash-envelope/05/). COSE Hash Envelope is essentially another format of [COSE Sign1 signature structure](https://datatracker.ietf.org/doc/html/rfc8152#section-4.2). 
+The differences between a `COSE Hash Envelope` and the signature envelope defined under [Signature Envelope](./signature-envelope-cose.md/#signature-envelope) are as follows:
```
FeynmanZhou, line 244: As a developer who wants to implement the Notary Projectsignature spec for their signing tools, how could the developer know when to choose COSE Hash Envelope and when to choose the signature envelope? 
We only declared the difference between them but did not briefly introduce their typical use case. This might not be a common knowledge for tool developers. Are we expecting the developers to read through each RFC?

specs/signature-envelope-cose.md
```diff
@@ -239,6 +239,52 @@ The final signature envelope is a `COSE_Sign1_Tagged` object, consisting of Payl
...
+1. Instead of `targetArtifact` as shown under [COSE Payload](./signature-envelope-cose.md/#cose-payload), the payload of a `COSE Hash Envelope` is the hash of the original content bytes that is being signed. The hash algorithm MUST be deduced from the signing certificate's public key (see [Algorithm Selection](./signature-specification.md#algorithm-selection)).
+2. New protected header parameters under `COSE Hash Envelope` only:
+    - **[`payload hash alg`](https://datatracker.ietf.org/doc/draft-ietf-cose-hash-envelope/05/)** (*int*): REQUIRED protected header payload hash alg (label `258`) specifies the hash algorithm used to produce the payload.
+    - **[`preimage content type`](https://datatracker.ietf.org/doc/draft-ietf-cose-hash-envelope/05/)** (*uint/tstr*): OPTIONAL protected header preimage content type (label `259`) specifies the content type of the bytes that were hashed (preimage) to produce the payload, given as a content-format number (Section 12.3 of [RFC7252]) or as a media-type name optionally with parameters (Section 8.3 of [RFC9110]).
```
FeynmanZhou: Do we want to add links to RFC7252 and RFC9110?

specs/signature-envelope-cose.md
```diff
@@ -239,6 +239,52 @@ The final signature envelope is a `COSE_Sign1_Tagged` object, consisting of Payl
 )
 ```
 
+## COSE Hash Envelope
+Notary Project also supports [COSE Hash Envelope](https://datatracker.ietf.org/doc/draft-ietf-cose-hash-envelope/05/). COSE Hash Envelope is essentially another format of [COSE Sign1 signature structure](https://datatracker.ietf.org/doc/html/rfc8152#section-4.2).
```
yizha1: COSE Hash Envelope will be the only format for Blob COSE envelope. I believe the main difference comparing with previous is the payload. For OCI COSE envelope, there are not changes required for now. Since blob signing is not officially released yet, we can make the change without breaking anything. /cc @shizhMSFT

specs/signature-envelope-cose.md
shizhMSFT: We do need them and they are missing.

specs/signature-envelope-cose.md
```diff
@@ -239,6 +239,52 @@ The final signature envelope is a `COSE_Sign1_Tagged` object, consisting of Payl
...
+    - **[`payload hash alg`](https://datatracker.ietf.org/doc/draft-ietf-cose-hash-envelope/05/)** (*int*): REQUIRED protected header payload hash alg (label `258`) specifies the hash algorithm used to produce the payload.
+    - **[`preimage content type`](https://datatracker.ietf.org/doc/draft-ietf-cose-hash-envelope/05/)** (*uint/tstr*): OPTIONAL protected header preimage content type (label `259`) specifies the content type of the bytes that were hashed (preimage) to produce the payload, given as a content-format number (Section 12.3 of [RFC7252]) or as a media-type name optionally with parameters (Section 8.3 of [RFC9110]).
+    - **[`payload location`](https://datatracker.ietf.org/doc/draft-ietf-cose-hash-envelope/05/)** (*tstr*): OPTIONAL protected header payload location (label `260`) is an identifier enabling retrieval of the original resource (preimage) identified by the payload.
+3. Header parameters that MUST NOT be present in `COSE Hash Envelope`:
+    - `content type` (label `3`) MUST NOT be present in the protected or unprotected headers. (Section 4 of [draft-ietf-cose-hash-envelope/05])
+
+## COSE Hash Envelope example
```
shizhMSFT:
```suggestion
### COSE Hash Envelope example
```

specs/signature-envelope-cose.md
```diff
@@ -239,6 +239,52 @@ The final signature envelope is a `COSE_Sign1_Tagged` object, consisting of Payl
 )
 ```
 
+## COSE Hash Envelope
+Notary Project also supports [COSE Hash Envelope](https://datatracker.ietf.org/doc/draft-ietf-cose-hash-envelope/05/). COSE Hash Envelope is essentially another format of [COSE Sign1 signature structure](https://datatracker.ietf.org/doc/html/rfc8152#section-4.2). 
+The differences between a `COSE Hash Envelope` and the signature envelope defined under [Signature Envelope](./signature-envelope-cose.md/#signature-envelope) are as follows:
+1. Instead of `targetArtifact` as shown under [COSE Payload](./signature-envelope-cose.md/#cose-payload), the payload of a `COSE Hash Envelope` is the hash of the original content bytes that is being signed. The hash algorithm MUST be deduced from the signing certificate's public key (see [Algorithm Selection](./signature-specification.md#algorithm-selection)).
```
shizhMSFT: COSE Hash Envelope does not use `targetArtifact` defined in `COSE Payload` as the payload or pre-image payload. Instead, it decomposes the information specified in `argetArtifact` to COSE Hash Envelope specific headers. Basically, there is a mapping:
- `targetArtifact.mediaType` --> `cose_sign1.protected_header.preimage_content_type`
- `targetArtifact.digest` --> `cose_sign1.payload`

Note that `targetArtifact.size`, `targetArtifact,annotations`, and `targetArtifact.artifactType` are lost. We may think about other ways to support them.

shizhMSFT (changes requested):

specs/signature-envelope-cose.md
ghost: > COSE Hash Envelope will be the only format for Blob COSE envelope. I believe the main difference comparing with previous is the payload.

@yizha1 The differences are payload and protected headers. You can find the differences [here](https://datatracker.ietf.org/doc/draft-ietf-cose-hash-envelope/05/).

specs/signature-envelope-cose.md
ghost: > Note that `targetArtifact.size`, `targetArtifact,annotations`, and `targetArtifact.artifactType` are lost. We may think about other ways to support them.

Updated.

specs/signature-envelope-cose.md
ghost: @FeynmanZhou This is a good question, the answer is YES. As a developer who wants to implement this spec, they MUST read these RFCs. In other words, they MUST know exactly what they are implementing and offering as a security tool. This is a requirement that's so fundamental that I believe any developer/maintainer/implementer of this spec should always keep in their mind.

specs/signature-envelope-cose.md
ghost: Updated.

specs/signature-envelope-cose.md
ghost: Updated.

specs/signature-envelope-cose.md
ghost: Added NOTE to clarify `COSE Hash Envelope` is only supported in blob signing for now.

specs/signature-envelope-cose.md
ghost: Added NOTE to clarify `COSE Hash Envelope` is only supported in blob signing for now.

specs/signing-and-verification-workflow.md
```diff
@@ -82,11 +82,27 @@ The user wants to sign an arbitrary blob with a detached signature.
 ### Signing Steps
 
 1. **Generate signature:** Using notation CLI or any other compliant signing tool, sign the blob. The signing tool should follow the following guideline.
-    1. Construct the blob payload as defined in [`signature specification`](./signature-specification.md#payload)
+    1. Construct the blob payload:
+       - For [`COSE Hash Envelope`](./signature-envelope-cose.md/#cose-hash-envelope): 
+
+         Calculate the hash of the blob using the hash algorithm deduced from signing certificate's public key (see [Algorithm Selection](./signature-specification.md#algorithm-selection)).
+
+       - Signature envelope other than `COSE Hash Envelope`:
```
yizha1: I would suggest removing other signature envelope for blob signing, as we only support one signature format for blob signing.

specs/signing-and-verification-workflow.md
```diff
@@ -82,11 +82,27 @@ The user wants to sign an arbitrary blob with a detached signature.
...
-        1. Generate signature using a signature format specified in [supported signature envelopes](./signature-specification.md#supported-signature-envelopes). Also, as part of this step, the user-defined/supplied custom attributes should be added to the annotations of the signature payload.
+    1. Generate signature:
+        1. Generate signature using a signature format specified in [supported signature envelopes](./signature-specification.md#supported-signature-envelopes):
+           - For [`COSE Hash Envelope`](./signature-envelope-cose.md/#cose-hash-envelope):
+
+               If provided by the user, the blob's media type MUST be placed at `protectedHeader.preimage-content-type (label 259)`.
+
+               Any user-defined/supplied custom attributes should be added to the [`Extended attributes`](./signature-specification.md/#extended-attributes).
+           
+           - Signature envelope other than `COSE Hash Envelope`:
```
yizha1: Simliar to the previous comment, we only support one signature format.

specs/signing-and-verification-workflow.md
```diff
@@ -110,7 +126,14 @@ The user wants to consume an arbitrary blob only if it was signed by a trusted p
...
\ No newline at end of file
+1. **Verify the payload:**
+    1. For [`COSE Hash Envelope`](./signature-envelope-cose.md/#cose-hash-envelope):
+       1. Deduce payload hash algorithm from signing certificate's public key (see [Algorithm Selection](./signature-specification.md#algorithm-selection)) and match it with the value specified in `protectedHeader.payload-hash-alg (label 258)`. Fail signature verification if there is a mismatch. 
+       1. Calculate the hash of the blob using the hash algorithm from step above, and match it with the value specified at signature envelope `payload`. Fail signature verification if there is a mismatch.
+       1. If provided by the user, verify blob's media type to the one present in `protectedHeader.preimage-content-type (label 259)`. Fail signature verification if there is a mismatch.       
+       1. If there is any user-defined/supplied custom metadata, match them against the ones present in [`Extended attributes`](./signature-specification.md/#extended-attributes). If they match, signature verification is considered successful.
+   1. Signature envelope other than `COSE Hash Envelope`:
```
yizha1: We will only suypport `COSE Hash Envelope` for blob signing

specs/signature-specification.md
```diff
@@ -122,7 +122,7 @@ The Notary Project signature specification requires the signature envelope to su
 
 Implementations of the Notary Project signature specification MAY include additional signed attributes in the signature envelope.
 These attributes MAY be marked critical, i.e. the attribute MUST be understood and processed by a verifier, unknown critical attributes MUST cause signature verification to fail.
-Usage of extended signed attributes which are marked critical in signature will have implications on portability of the signature, these are discussed in [Signature Portability](#signature-portability) section.
+Usage of extended signed attributes which are marked critical in signature will have implications on portability of the signature, these are discussed in [Signature Portability](#oci-signature-portability) section.
```
yizha1, line 125: How about blob signature portability? if it is only for OCI signature, I would suggest to add words like "for OCI signature portability" as this paragraph is under the common section for both OCI and blob signature

specs/signature-envelope-cose.md
```diff
@@ -239,6 +239,58 @@ The final signature envelope is a `COSE_Sign1_Tagged` object, consisting of Payl
 )
 ```
 
+## COSE Hash Envelope for signing arbitrary blobs
+Notary Project also supports [COSE Hash Envelope](https://datatracker.ietf.org/doc/draft-ietf-cose-hash-envelope/05/) when signing an arbitrary blob. COSE Hash Envelope is essentially another format of [COSE Sign1 signature structure](https://datatracker.ietf.org/doc/html/rfc8152#section-4.2).
```
yizha1: We will only support one signature format for blob signing, which is COSE Hash Evenlop. So we only need to describe what is the payload and what does COSE hash evenlope look like, we don't need to do a comparision.

yizha1 (commented):
Thanks @Two-Hearts. I left several comments. The main comment is that we will not support two types of COSE signature format for blob signing. We only support `COSE Hash Evenlope` for COSE signatures.

specs/signature-specification.md
ghost: This change is not related to this PR's major update though. The change is merely correcting the broken link.

specs/signature-envelope-cose.md
ghost: The comparison is necessary, because otherwise, the reader would be confusing on the differences between the two COSE formats. This comparison is for envelope structure and not for purpose.

specs/signing-and-verification-workflow.md
ghost: Updated.

specs/signing-and-verification-workflow.md
ghost: Updated.

specs/signing-and-verification-workflow.md
ghost: Updated.

specs/signature-envelope-cose.md
ghost: @yizha1 I rephrased the wording to make it clear that `We will only support one signature format for blob signing, which is COSE Hash Evenlop.`

specs/signing-and-verification-workflow.md
```diff
@@ -82,11 +82,13 @@ The user wants to sign an arbitrary blob with a detached signature.
 ### Signing Steps
 
 1. **Generate signature:** Using notation CLI or any other compliant signing tool, sign the blob. The signing tool should follow the following guideline.
-    1. Construct the blob payload as defined in [`signature specification`](./signature-specification.md#payload)
+    1. Construct the blob payload:
+       - For COSE format, only [`COSE Hash Envelope`](./signature-envelope-cose.md/#cose-hash-envelope) is supported when signing arbitrary blobs.
```
shizhMSFT, line 86: @yizha1 @FeynmanZhou Is this what we agree with?

shizhMSFT: I see that there are `cose` and `cose-hash`.

shizhMSFT: Basically, we generate payload depending on which envelope type we choose.

specs/signing-and-verification-workflow.md
```diff
@@ -110,7 +112,14 @@ The user wants to consume an arbitrary blob only if it was signed by a trusted p
...
+       1. If there is any user-defined/supplied custom metadata, match them against the ones present in [`Extended attributes`](./signature-specification.md/#extended-attributes). If they match, signature verification is considered successful.
+   1. For JOSE format: 
+      1. Calculate the blob's size and verify that it matches the size present in `targetArtifact.size`. Fail signature verification if there is a mismatch.
+      1. If provided by the user, verify blob's media type to the one present in `targetArtifact.mediaType`. Fail signature verification if there is a mismatch.
+      1. Calculate the digest of the blob using the digest algorithm deduced from signing certificate's public key (see [Algorithm Selection](./signature-specification.md#algorithm-selection)) and match it with the digest specified at `targetArtifact.digest`.  Fail signature verification if there is a mismatch.
+      1. If there is any user-defined/supplied custom annotations, match them against the ones present in `targetArtifact.annotations`. If they match, signature verification is considered successful.
```
shizhMSFT, line 125: We still should have a high-level verification protocol without having a switch case for formats.

specs/signature-envelope-cose.md
```diff
@@ -239,6 +239,55 @@ The final signature envelope is a `COSE_Sign1_Tagged` object, consisting of Payl
...
+    - **[`preimage content type`](https://datatracker.ietf.org/doc/draft-ietf-cose-hash-envelope/05/)** (*uint/tstr*): REQUIRED protected header preimage content type (label `259`) specifies the content type of the bytes that were hashed (preimage) to produce the payload, given as a content-format number (Section 12.3 of [RFC7252](https://datatracker.ietf.org/doc/html/rfc7252#section-12.3)) or as a media-type name optionally with parameters (Section 8.3 of [RFC9110](https://datatracker.ietf.org/doc/html/rfc9110#section-8.3)). This value MUST have the same value as [`targetArtifact.mediaType`](./signature-envelope-cose.md/#cose-payload).
+    - **[`payload location`](https://datatracker.ietf.org/doc/draft-ietf-cose-hash-envelope/05/)** (*tstr*): OPTIONAL protected header payload location (label `260`) is an identifier enabling retrieval of the original resource (preimage) identified by the payload.
+    - **`io.cncf.notary.preimageSize`** (*int*): REQUIRED protected header `io.cncf.notary.preimageSize` specifies the size of the preimage. This value MUST have the same value as [`targetArtifact.size`](./signature-envelope-cose.md/#cose-payload).
```
shizhMSFT, line 251: It should be `uint`. Do we have a better name for that?

specs/signature-envelope-cose.md
```diff
@@ -239,6 +239,55 @@ The final signature envelope is a `COSE_Sign1_Tagged` object, consisting of Payl
...
+    - **`io.cncf.notary.artifactType`** (*tstr*): OPTIONAL protected header `io.cncf.notary.artifactType` specifies the artifactType of the preimage. This value MUST have the same value as [`targetArtifact.artifactType`](./signature-envelope-cose.md/#cose-payload).
```
shizhMSFT, line 252: Do we accept `uint` as defined in https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats?

shizhMSFT (changes requested):

specs/signing-and-verification-workflow.md
yizha1: @shizhMSFT  I believe we only need cose-hash for blob signature.

specs/signing-and-verification-workflow.md
yizha1: @shizhMSFT I don't think we will need JOSE format payload for blob signature. I don't see there is a use case for it.

FeynmanZhou: How about we could keep it simple by supporting `COSE Hash Envelope` for blob signature only in this release. JOSE format payload could be considered in future releases once there is any request.

PR #330, opened 2026-07-02
docs: support registry-level scopes in trust policy

### LINKED ISSUE #289: Support specifying multiple registries in the trust policy
From the [trust policy spec](https://github.com/notaryproject/specifications/blob/main/specs/trust-store-trust-policy.md#registry-scopes-constraints), it requires users to specify a list of one or more fully qualified repository URIs in the `registryScopes`. 

However, this requirement limits the verification scope to repository level and narrows the verification scenario. If users want to specify all repositories under multiple registries in the `registryScopes`, it will be inconvenient. 

I would suggest extending the scope to registry level. It will enable users to specify not only repositories but also registries in the `registryScopes`. 

For example:

```
"registryScopes": [
              "registry.acme-rockets.io",
              "wabbit-networks.io"
            ]
```

rgnote:
Extend `registryScopes` in the OCI trust policy to accept a registry host (a value with no `/` separator) in addition to a fully qualified repository URI. A registry scope applies the policy to every repository in that registry, which addresses the operational pain of enumerating a fully qualified URI per repository when trust is reasoned about at the registry (e.g. account+region) boundary.

To keep the "exactly one applicable policy per artifact" invariant, the scope forms are given an explicit specificity ordering (repository > registry > global) and the most specific match wins, so a repository level exception can coexist with a broader registry or global policy. Matching is defined as exact, case-sensitive string comparison with no reference expansion, prefix matching, or case folding.

Resolves #289


specs/trust-store-trust-policy.md
```diff
@@ -394,13 +427,20 @@ Timestamp countersignature verification is a multi step process performs the fol
 ##### OCI Trust Policy Constraints
 
 - Each trust policy MUST contain scope property `registryScopes` and the scope collection MUST contain at least one value.
-- The scope MUST contain one of the following:
-  - List of one or more fully qualified repository URIs.
+- Each value in the scope collection MUST be one of the following:
+  - A fully qualified repository URI of the form `${registry-name}/${namespace}/${repository-name}`.
```
yizha1, line 431: What is the behavior if a user specify `${registry-name}/${namespace}`, for example `registry.acme-rockets.io/software/`, does it mean all the repositories under software follow the specific policy?

yizha1 (commented):
Thanks @rgnote left one comment.

specs/trust-store-trust-policy.md
rgnote: Thanks for the comment @yizha1 

We accept only 1) a repository path, or 2) a registry path, or 3) a wildcard. 

If a user provides `registry.acme-rockets.io/software/`, it is considered as a bad input since it does not belong to any of the three acceptable inputs (because of the trailing slash)

yizha1: I didn't make my question clear. For example, for these two repos, `registry.acme-rockets.io/software/acme` and `registry.acme-rockets.io/software/acme`, they shared the same policy, can I use `registry.acme-rockets.io/software` to match these two repo, so that I don't have to specify each of them in the policy.

rgnote: Thank you for the clarification. I now understand the question. 

To answer the question, no, we are not supporting partial trust anchors in this revision, mostly to keep the specification simple and easy to reason about. Let me know if it is a MUST requirement that we need to support it in this revision.

yizha1: Hi @rgnote no, it's not a MUST, and we can consider it in the future if there are more customer asks. I would suggest making it clear in the spec this case is not supported.

rgnote: Added some wording to clarify this.

yizha1 (approved):
LGTM. Thanks @rgnote 
We can support more cases in the future.

rgnote:
Thanks, @yizha1. Tagging @gokarnm and @shizhMSFT if they can review this PR.

rgnote:
@yizha1 

Looks like I need 2 more approvals before I can merge this. Can someone from Microsoft review this?

yizha1:
@toddysm would you mind taking a look at this PR?

